PR #9, opened 2018-11-16
Auto populate list of programs

jturney:
The model currently in QCEngine requires the coder to add their new Python file and then modify two other locations to enable it.  This PR allows the coder to simply add their new Python file to the proper location and have a similarly named function inside of it and that's it.

### DIFF
--- a/qcengine/compute.py
+++ b/qcengine/compute.py
@@ -36,13 +36,9 @@ def compute(input_data, program, raise_error=False, capture_output=True):
     # Run the program
     with util.compute_wrapper(capture_output=capture_output) as metadata:
         output_data = input_data
-        if program == "psi4":
-            output_data = programs.run_psi4(input_data)
-        elif program == "rdkit":
-            output_data = programs.run_rdkit(input_data)
-        elif program == "torchani":
-            output_data = programs.run_torchani(input_data)
-        else:
+        try:
+            output_data = programs.get_program(program)(input_data)
+        except RuntimeError as e:
             output_data["success"] = False
             output_data["error_message"] = "QCEngine Call Error:\nProgram {} not understood".format(program)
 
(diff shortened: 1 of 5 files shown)


loriab:
I worry a little that this might catch strays with all the `psi4.py`, `core.so`, `plugin.so`s running around.

dgasmith:
What about a `qcengine.programs.register_program(alias, function)` kind of thing where we have a global dict in the `programs` folder that we can add to via the above. 

Sorry, I didn't get back to this earlier. I had forgotten to watch this repo.

jturney:
There shouldn't be a problem catching strays as it only imports files from that directory. Doing `qcengine.programs.register_program(alias, function)` may be a good idea if we ever need multiple entry points in a file. I'll work on changing over to that.

dgasmith:
Let me know if you need help getting this in.

dgasmith:
I will likely do this in the next few days, let me know if I should not!

jturney:
I'm working on this right now!

qcengine/compute.py
```diff
@@ -36,15 +36,12 @@ def compute(input_data, program, raise_error=False, capture_output=True):
...
-            output_data = programs.run_rdkit(input_data)
-        elif program == "torchani":
-            output_data = programs.run_torchani(input_data)
-        else:
+        try:
+            output_data = programs.get_program(program)(input_data)
+        except Exception as e:
             output_data["success"] = False
-            output_data["error_message"] = "QCEngine Call Error:\nProgram {} not understood".format(program)
+            output_data["error_message"] = "QCEngine Call Error:\nProgram {} not understood.\nError Message: {}".format(
```
dgasmith: As a note the wrapper handles Exception capturing. This try/except should explicitly be for key errors.

jturney: I'll add it and resolve the conflict shortly.

dgasmith:
Merging, thanks for the PR!

PR #14, opened 2019-01-17
Options Overhaul and QCElemental

### LINKED ISSUE #3: Per-computation settings
Settings are currently global are not allowed to be overridden on a per computation basis. Local options passed into a computation should be able to overwrite the global state to allow for more flexibility.

### LINKED ISSUE #12: Pydantic Options
Move to pydantic options rather than raw dict for two reasons:
 - Validation is automatic and error message will look the same across QCA ecosystem,
 - Merging options is straightforward for local options, #3.

dgasmith:
This PR accomplishes the following:
 - Adds QCElemental and pydantic as dependancies (requires Py3.6+).
 - Converts all units over to QCElemental CODATA2014 units.
 - Overhauls the options to have a separate NodeDescriptor area which can effectively match a node name to a preset description. Closes #12.
 - Allows individual configuration to be build from a NodeDescriptor to create a on-the-fly JobConfig.
 - A JobConfig is alterable with a new `local_options` command. Closes #3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: python
+language: generic
 
 # Run jobs on container-based infrastructure, can be overridden per job
 dist: xenial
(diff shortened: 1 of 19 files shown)


dgasmith:
Cool, cleaned up.

PR #16, opened 2019-01-24
Implement the QCElemental Models into QCEngine

Lnaden:
This is an overhaul to the QCEngine program to work with the
Pydantic Models which have been/will be added to the QCElemental package.

Right now, the models are all internal and will always return the
flattened dictionary of near-pure Python objects (numpy arrays being an exception).
The emitted objects should have no Pydantic in them, and right now will not
correctly emit in many casses; errors, procedures; if you try. As we continue
the integration of the QCElemental objects into QCA as a whole, we can
revisit this and get it working correctly.

This will need review since it changes a very large swath of code, all of which
had to be changed to get the tests passing.

Depends on molssi/qcelemental#17

### DIFF
--- a/qcengine/compute.py
+++ b/qcengine/compute.py
@@ -6,7 +6,7 @@ from qcelemental.models import ResultInput, ComputeError, OptimizationInput, Opt
 from pydantic import ValidationError
 
 from .programs import get_program
-from .util import compute_wrapper, handle_output_metadata, get_module_function
+from .util import compute_wrapper, handle_output_metadata, get_module_function, model_wrapper
 from .config import get_config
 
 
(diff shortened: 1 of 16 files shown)


qcengine/compute.py
```diff
@@ -27,42 +30,86 @@ def compute(input_data, program, raise_error=False, capture_output=True, local_o
...
     local_options : dict, optional
         A dictionary of local configuration options
+    return_dict : bool, optional, default True
+        Returns a dict instead of qcelemental.models.ResultInput
 
     Returns
     -------
-    ret : dict
-        A QC Schema output
+    ret : dict or qcelemental.models.ResultInput
```
dgasmith: `Result` not input.

qcengine/config.py
```diff
@@ -222,12 +222,13 @@ def get_config(hostname=None, local_options=None):
...
-        "cpu": GLOBALS["cpu"],
-        "hostname": GLOBALS["hostname"],
-        "username": GLOBALS["username"],
-        "qcengine_version": __version__
-    }
+    ret = Provenance(
+        cpu=GLOBALS["cpu"],
+        hostname=GLOBALS["hostname"],
+        username=GLOBALS["username"],
+        version=__version__,
```
dgasmith: We cannot use `version` here as it will overwrite the programs version. Please revert and remove the creator as well below.

qcengine/programs/rdkit.py
```diff
@@ -18,76 +19,83 @@ def rdkit(ret_data, config):
...
 
     # Build the Molecule
-    jmol = ret_data["molecule"]
+    jmol = input_data.molecule
 
     # Handle errors
-    if ("molecular_charge" in jmol) and (abs(jmol["molecular_charge"]) < 1.e-6):
-        ret_data["error_message"] = "run_rdkit does not currently support charged molecules"
-        return ret_data
+    if abs(jmol.molecular_charge > 1.e-6):
```
dgasmith: `abs(jmol.molecular_charge) > 1.e-6`

qcengine/config.py
Lnaden: I think I fixed this, the behavior should reflect what is in live right now. I think this Provenance's `Creator` still overrides the output program creator, but it does that now. Unsure if thats a bug or on purpose.

On closer look, maybe it does still overwrite the version. Should we instead be updating the Provenance as `QCEngineProvenance.update(SourceProgramProvenance)` instead of the other way around like we do now?

dgasmith: We previously did not set the creator field so this wasn't an issue. This is mostly extra interesting information that had nowhere else to go and should not overlap with the source program provenance.

dgasmith:
Made a few minor tweaks and tests here. This is also rebased.

PR #17, opened 2019-01-25
Reworks Config

dgasmith:
Another Config rework. Still playing around with this in regard to dealing with multiple queue managers. Effectively this allows to have a bit more control over the actual job execution and lessens the general node descriptions. This is interoperating with Parsl/Fireworks/Dask/Slurm/Torque/Cobalt a bit better.

### DIFF
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -16,7 +16,7 @@ import psutil
 import pydantic
 import yaml
 
-__all__ = ["get_config", "get_provenance", "global_repr", "list_compute", "NodeDescriptor"]
+__all__ = ["get_config", "get_provenance", "global_repr", "NodeDescriptor"]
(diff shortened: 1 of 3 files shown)


Lnaden:
I like where this is going. I don't see any conflicts with the #16 at the moment. Do you think there will ever be a need to distinguish between things like `cores` vs. `treads`, or `memory` vs `memory_per_job`?

dgasmith:
`cores` vs `threads` is a long topic, I would like to not worry about that. Ultimately you typically get a node by the core which is why I changed it over.

`memory_per_job` was only for a NodeDescriptor, that is now entirely constructed on the fly based on local data. We can certainly add things back in at some point. Trying to cut down to a core set for now.

PR #18, opened 2019-01-27
add DFTD3 interface

loriab:
- [x] move `intf_dftd3` to qcng, with dependencies on only qcng and qcdl
- [x] all original tests still working when psi4 also detected (for psi4.Mol and qcdb.Mol objects)

TODO
- [x] use qcschema Mol
- [x] use qcschema job
- [x] adjust license headers


loriab:
At this point, CI passes. A few notes so CI things can be relaxed/reverted in future.

* there are now two "conda-env" PSI4 environment.yaml files. this is because latest release psi4 (v1.2) doesn't do py37, so travis No.3 would always fail (after I fixed the travis (in the same way as other qca repos) to actually mind the python version being given to it, not just always run py36).
* the latest-release PSI4 (travis No.2) will always fail in the presence of the qcengine `intf_dftd3` b/c of differences in how qcschema is to/from in the internal psi4 molparse vs the qcel.molparse. no fix, just need to drop v1.2 sooner.
* the nightly-build PSI4DEV (travis No.3) will fail for `Molecule.run_dftd3` with 3-body b/c that's calling `intf_dftd3` in psi4, not in qcengine. have restrained dftd3 to 3.2.0 so these don't run.

devtools/scripts/conda_env.py
```diff
@@ -0,0 +1,31 @@
...
+
+conda_path = shutil.which("conda")
+
+print("CONDA ENV NAME  {}".format(args.name[0]))
+print("PYTHON VERSION  {}".format(args.python[0]))
+print("CONDA FILE NAME {}".format(args.conda_file[0]))
+print("CONDA path      {}".format(conda_path))
+
+sp.call("{} env create -n {} -f {}".format(conda_path, args.name[0], tmp_file), shell=True)
+os.unlink(tmp_file)
```
dgasmith: Can we copy this one from the cookiecutter?

qcengine/config.py
```diff
@@ -14,8 +14,11 @@
 import pydantic
 import yaml
 
+from ._version import get_versions
```
dgasmith: On this I noted that repeated `get_versions()` is fairly slow. Can we add this once to GLOBALS and then pull it from there everywhere else?

qcengine/programs/intf_dftd3/runner.py
```diff
@@ -0,0 +1,506 @@
...
+    jobrec['provenance'] = provenance_stamp(sys._getframe().f_code.co_name + '.' + __name__)
+
+    # strip engine hint
+    if name.startswith('d3-'):
+        name = name[3:]
+
+    jobrec.update(molecule.to_schema(dtype=1))  # --> jobrec['molecule'] =
+    jobrec['model'] = {
+        'method': name,
+        'basis': '(auto)',
```
dgasmith: I usually do `basis: None` for these.

qcengine/tests/addons.py
```diff
@@ -21,12 +37,49 @@ def _plugin_import(plug):
...
 using_geometric = pytest.mark.skipif(
-    _plugin_import("geometric") is False, reason="could not find geomeTRIC. Please install the package to enable tests")
+    _plugin_import("geometric") is False,
+    reason="could not find geomeTRIC. Please install the package to enable tests")
+
 using_torchani = pytest.mark.skipif(
     _plugin_import("torchani") is False, reason="Could not find TorchAni. Please install the package to enable tests")
+
+using_qcdb = pytest.mark.skipif(
+    _plugin_import("qcdb") is False, reason='Not detecting common driver. Install package if necessary and add to envvar PYTHONPATH')
```
dgasmith: Are you planning to do more QCDB testing here?

loriab:
I still have a pydantic step to do here, but at least testing is up-to-date with qcel.

dgasmith:
I would like to pull this in now to give another use case for a new CLI execution framework that would cover dftd3, Molpro, and Terachem. Any objections?

loriab:
go for it. it's at a reasonable stopping point.

actually, I copied your extras.py version handling over from qcel, and lgtm thinks it's returning None. I'd better check that first.

loriab:
it was indeed broken. fixed now, so mergeable after CI.

dgasmith:
LGTM to the rescue!

PR #21, opened 2019-01-28
Cleanup to emulate old style returns

Lnaden:
Ensures the compute dicts are returned as python primatives and let
pydantic serialize the objects

Also ensure output_data exists in all error catching. Told LGTM to
ignore this particular error because its expected

Needed to fix QCFractal

### DIFF
--- a/qcengine/compute.py
+++ b/qcengine/compute.py
@@ -112,7 +112,7 @@ def compute_procedure(input_data,
     # Run the procedure
     with compute_wrapper(capture_output=capture_output) as metadata:
         # Create a base output data in case of errors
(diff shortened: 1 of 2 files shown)


Lnaden:
This will require a patch release of QCElemental to work correctly

dgasmith:
Please do make that patch and I will crank out a new release.

Lnaden:
The QCElemental changes are already in master, a patch version should be able to be spun off there

dgasmith:
Ok, QCEl v0.2.2 is out. Bump setup.py and conda envs versions and roll from there.

Lnaden:
Closing in favor of #22

PR #29, opened 2019-02-16
Molpro: Input generator and output parser (WIP)

sjrl:
This pull request contains the start of a Molpro input generator and output parser.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ install:
       exit 1
     fi
   - source activate test
+  - git+https://github.com/MolSSI/QCEngineRecords.git#egg=qcenginerecords
 
     # Build and install package
   - python setup.py develop --no-deps
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ install:
       exit 1
     fi
   - source activate test
-  - git+https://github.com/MolSSI/QCEngineRecords.git#egg=qcenginerecords
+  - pip install git+https://github.com/MolSSI/QCEngineRecords.git#egg=qcenginerecords
 
     # Build and install package
   - python setup.py develop --no-deps
--- a/qcengine/programs/__init__.py
+++ b/qcengine/programs/__init__.py
@@ -5,6 +5,7 @@ Imports the various compute backends
 from .psi4 import Psi4Executor
 from .rdkit import RDKitExecutor
 from .torchani import TorchANIExecutor
+from .molpro import MolproExecutor
 
 __all__ = ["register_program", "get_program", "get_programs"]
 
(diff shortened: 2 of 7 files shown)


dgasmith:
Thanks for the update! Mind if I hack on this a bit to fit it into the new executor objects?

sjrl:
No problem! And go for it, I was just in the process of parsing more information from the xml file so it shouldn't be a problem merging my changes into what you do.

sjrl:
Do you mind if I merge in the current version of master (from MolSSI/QCEngine) and push the commits I have on `molpro.py`?

dgasmith:
That’s fine! If you could rebase over merge that would be perfect.

sjrl:
Okay so I have rebased on my local machine, but to push it to the remote `molpro` branch it appears I will need to use `git push --force origin/molpro`. Just double checking this is alright before going ahead since I will be effectively rewriting `origin/molpro`.

dgasmith:
Yup. That is expected.

dgasmith:
Updated this branch to test the current parser with QCEngineRecords. This technology is still a bit loose, but should give you a middle ground and potentially ways to improve the parser.

@loriab and I talked about using exact string matching for comparing the generated input file with the ability to regenerate the inputs on the fly.

sjrl:
Sorry for my lack of knowledge, but it looks like pytest is skipping the `test_molpro_output_parser()`. Is there a specific way I'm meant to be calling pytest to actually run the test?

sjrl:
Also I think using exact string matching would be the easiest way to check input file generation.

dgasmith:
Yea, we will need to exact string matching there isn't really another way. We will need a way to re-render the inputs in QCEngineRecord where required in this case.

I messed up the Travis script, but to run these tests you need to install https://github.com/MolSSI/QCEngineRecords. Standard affair- clone/`setup.py install -e .`.

dgasmith:
@sjrl Can you make a cleanup pass here? I say let us merge this sooner rather than later so we can get these changes into the terachem parser as well. We will need to make another PR for the actual execution of Molpro.

sjrl:
Yes that sounds good to me. What are the list of things that you would like me to do for this cleanup? If it's just getting the test job passed for Molpro we just need to resolve if "return_result" is meant to always contain the gradient over the energy if the gradient is available. Otherwise the other TODOs I have listed in `molpro.py` can be done at a future date.

sjrl:
Never mind my previous comment about the 'return_result' conflict. That was an error on my end. I'll get the LGTM issues and we should be good to go.

sjrl:
@dgasmith I've made a quick pass and handled the LGTM issues. Should be good for a merge.

dgasmith:
Thanks! Looks good, hop on Slack sometime to chat about next steps.

PR #33, opened 2019-02-23
DFTD3 as ProgramExecutor

loriab:
- [x] smoke test running in test_programs


qcengine/programs/dftd3/runner.py
```diff
@@ -14,13 +14,51 @@
...
 from .util import provenance_stamp, parse_dertype
 
 
+
+class DFTD3Executor(ProgramExecutor):
+
+    _defaults = {
+        "name": "DFTD3",
+        "scratch": True,
+        "thread_safe": False,
```
dgasmith: This is thread safe.

PR #34, opened 2019-02-27
Model Update

dgasmith:
Moves several items around in preparation of strict keys in the base model which will happen in QCElemental v0.3 slated for later this week. Also:
 - tweaks stdout/stderr so that it is correctly omitted when no data is saved
 - Input errors are now correctly handled and raised.


dgasmith:
This will require a new QCElemental release.

PR #35, opened 2019-02-27
CI: Pull pytorch-cpu to avoid pulling CUDA in CI

dgasmith:
Also nukes the unused `meta.yaml`.


dgasmith:
Still pulls 600MB of depends-- down from 950MB. Better at least.

PR #36, opened 2019-02-28
v0.6 release tweaks

### LINKED ISSUE #15: Python-based codes to add to QCEngine
Several codes to add which may be relatively easy to add are:

Compute:
 - [ ] [PySCF](https://github.com/pyscf/pyscf)

Optimizers:
 - [ ] [PyBerny](https://github.com/jhrmnn/pyberny)
 - [ ] [OptKing](https://github.com/psi-rking/optking)

### LINKED ISSUE #20: Test returning as Pydantic objects and not just dicts
We only return dicts now, we will want to test the return of Pydantic objects as well, including when errors are thrown.

dgasmith:
As a note this will fail until `geomeTRIC` has a new release. This PR accomplishes the following:
 - Adds several docs pages.
 - Compute always return objects by default (breaking change).
 - List all programs available, closes #15.
 - Tests object returns, closes #20.

TODO:
 - Add v0.6 changelog


qcengine/util.py
```diff
@@ -430,3 +429,16 @@ def disk_files(infiles: Dict[str, str], outfiles: Dict[str, None], cwd: Optional
                     LOGGER.info(f'... Writing: {filename}')
             except (OSError, FileNotFoundError) as err:
                 outfiles[fl] = None
+
+
+def which(command, return_bool=False):
```
loriab: some of these `which`, `program_new_enough`, may want to make `qcel/util/`

PR #40, opened 2019-03-04
travis caching, setup.py test, updates psi4 travis

loriab:
* institute caching of conda pkgs for travis
* allow `python setup.py test` to run pytest
* try dftd3 and psi4 pinnings after v1.3 release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ cache:
     - $HOME/miniconda
 
 matrix:
-  allow_failures:
-    - os: linux
-      env:
-        - PYTHON_VER=3.6
-        - PROG=PSI4
+  #allow_failures:
+  #  - os: linux
+  #    env:
+  #      - PYTHON_VER=3.6
+  #      - PROG=PSI4
 
   include:
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ language: generic
 dist: xenial
 
 before_cache:
+  # if I understand internet correctly, this is equiv to "after_success"
+  - |
+      if [[ $TRAVIS_TEST_RESULT = 0 ]]; then
+        codecov
+      fi
   - conda deactivate
   - conda remove --name test --all
 cache:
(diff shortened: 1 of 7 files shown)


loriab:
This is ready for consideration. definitely SQUASH

dgasmith:
Missing `codecov` at the end. Looks like ~3 minutes to pull from the cache so a win, but not massive.

dgasmith:
6GB cache... killed that one off.

loriab:
this cacheing thing is more involved than I expected. have fixed the codecov by moving it to `before_cache` (https://docs.travis-ci.com/user/job-lifecycle). but I suspect the lgtm hang is the same problem, and I'm not seeing a handle to adjust that. finding nothing in their forum either. any suggestions before posting there?

loriab:
hooray, lgtm _wasn't_ hung!

I pushed one more commit to get a with-caching timing run. then this should be ready to go.

w/o cacheing
screen shot 2019-03-04 at 7 16 29 pm

loriab:
hmpf, for all that, no change.
screen shot 2019-03-04 at 8 51 09 pm

dgasmith:
It uploaded new packages at the end which took 2-4 minutes. Conda may be too active of a project.. unsure if it can do delta diffs

loriab:
oh, wait. that `blas-mkl` pkg was a new one I added last commit. didn't even think about it changing the cache. one more try ...

dgasmith:
Rebuilding, worth getting in and trying for a bit still I think. We have had enough Conda pull issues lately that it might at least help stabilize things.

loriab:
yup, I tried a rebuild, too. Apart from the `1.4a1.dev11`, which was to be expected after last night, I got those `.q` and `.json` files (which made the cache reupload). Not sure what they are, and the psi4 travis doesn't have them. perhaps a history feature of a new conda/miniconda version. maybe can get rid of them. but yes, I think that can be later examination, and this is RTG.

screen shot 2019-03-05 at 10 44 01 am

PR #41, opened 2019-03-05
Psi4: Gracefully handle internal failures

dgasmith:
Psi4 internal failures are now handled gracefully. Also does not force `uks` reference when `reference` keyword is present.

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -54,7 +54,7 @@ class Psi4Executor(ProgramExecutor):
 
             mol = psi4.core.Molecule.from_schema(input_model)
             if (mol.multiplicity() != 1) and ("reference" not in input_model["keywords"]):
-                input_model["keywords"]["reference"] = "uks"
+                input_model["keywords"]["reference"] = "uhf"
 
             output_data = psi4.json_wrapper.run_json(input_model)
             if "extras" not in output_data:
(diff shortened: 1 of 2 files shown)


qcengine/programs/psi4.py
```diff
@@ -53,14 +53,16 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
         if psi_version > self.parse_version("1.2"):
 
             mol = psi4.core.Molecule.from_schema(input_model)
-            if mol.multiplicity() != 1:
+            if (mol.multiplicity() != 1) and ("reference" not in input_model["keywords"]):
```
loriab: isn't `uks` a funny choice for generic, as opposed to `uhf`?

dgasmith: Yea, just hit that bug and realized why @ffangliu's example was failing.

dgasmith: See update.

ffangliu: Oh, I see. Thanks!

dgasmith:
Looks like it cannot find Psi4? @loriab

loriab:
failing in rdkit and torchani that don't have psi4 installed. needs something like `@using_psi4` above the test to only run when psi available.

PR #43, opened 2019-03-07
Procedures Base Models

### LINKED ISSUE #42: GeomeTRIC does not add a provenance
Currently geomeTRIC does not provide a provenance. We should also add a check in our standard test suite for provenance info.

dgasmith:
Normalizes procedure compute to be more inline with single program compute. Also fixes #42.


qcengine/procedures/base.py
```diff
@@ -0,0 +1,53 @@
+"""
+Imports the various procedure backends
+"""
+
+from typing import List, Set
+
+from .geometric import GeometricProcedure
```
Lnaden: Didn't we add a second procedure and service in Fractal? Is there a reason that could not be added here? If not, will Fractal's additional procedure derive from the abstract base model?

qcengine/compute.py
```diff
@@ -104,44 +129,39 @@ def compute_procedure(input_data: Dict[str, Any],
         A QC Schema representation of the requested output, type depends on return_dict key.
     """
 
-    # Build the model and validate
-    input_data = model_wrapper(input_data, OptimizationInput, raise_error)
-    if isinstance(input_data, FailedOperation):
-        if return_dict:
-            return input_data.dict()
-        return input_data
+    procedure = procedure.lower()
```
Lnaden: I feel like we could generalize this structure in a call, but its fine for now since it only comes up twice.

qcengine/compute.py
```diff
@@ -63,26 +97,17 @@ def compute(input_data: Union[Dict[str, Any], 'ResultInput'],
...
-                    error_message="QCEngine Call Error:\nProgram {} not understood."
-                    "\nError Message: {}".format(program, str(e))))
+        output_data = input_data.copy()  # lgtm [py/multiple-definition]
+        output_data = executor.compute(input_data, config)
 
     return handle_output_metadata(output_data, metadata, raise_error=raise_error, return_dict=return_dict)
 
 
-def compute_procedure(input_data: Dict[str, Any],
+def compute_procedure(input_data: Union[Dict[str, Any], 'BaseModel'],
```
Lnaden: Are we going to start using Type Hint Annotations everywhere? And because I cant remember, doesn't this require the `from __future__ import annotations` invocation to use in 3.6?

qcengine/procedures/base.py
dgasmith: Procedures will be here in Engine, all Services will stay in Fractal as they need Fractal-level protocols. Engine is also pure input/output QCSchema, Fractal records are Schema + stuff archival items.

qcengine/compute.py
dgasmith: This is only required for forward annotations which have no elected to do, hence the string fields. We are still pretty far from being able to run mypy.

PR #46, opened 2019-03-09
dftd3 renovation

loriab:
- [x] updates mol to dtype=2
- [x] provides full psi4 support for psi4/psi4#1491 (that deletes the interface from psi4)
- [ ] hook up file writer/reader to class
- [x] see about removing extra API interfaces and use purely json interface. calls to dashparam fns may be inevitable for data
- [x] add MP2D energy interface

~I wasn't sure about PR order. But maybe this should go in sooner (since harmless, I believe) so psi can test against.~

depends on MolSSI/QCElemental#47
depended upon by psi4/psi4#1491


loriab:
@dgasmith if you get a chance could you look over the structure of the dftd3/runner.py file and let me know if there's good rearrangements to make. Easier to do before they proliferate.

loriab:
drat, looks like I can't fix CI by pip installing a direct-from-branch qcel b/c psi4 will just grab an old conda qcel anyway to suit its demands. will need a qcel release first. going to be an interesting merge b/c this _still_won't pass w/o psi changes. actually easier to get psi's 1491 in first with a direct-from-branch qcng.

Anyway, I'm done with this until review comments. everything psi/dftd3/mp2d passes locally. others pass or skip.

qcengine/programs/mp2d.py
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    class Config(ProgramExecutor.Config):
+        pass
+
+    def __init__(self, **kwargs):
+        super().__init__(**{**self._defaults, **kwargs})
+
+    @staticmethod
+    def found(raise_error=False) -> bool:
```
dgasmith: Any reason for `raise_error` here. This doesn't associate with other methods it seems.

qcengine/programs/mp2d.py
```diff
@@ -0,0 +1,314 @@
...
+
+        output_data = run_json(input_data)
+
+        if output_data["success"]:
+            return Result(**output_data)
+        return FailedOperation(
+            success=output_data.pop("success", False), error=output_data.pop("error"), input_data=output_data)
+
+
+def run_json(jobrec):
```
dgasmith: The idea here is we eventually deprecate these extra functions correct?

qcengine/programs/psi4.py
```diff
@@ -24,15 +26,32 @@ class Config(ProgramExecutor.Config):
     def __init__(self, **kwargs):
         super().__init__(**{**self._defaults, **kwargs})
 
+    @staticmethod
+    def found(raise_error=False) -> bool:
```
dgasmith: Ah, we are changing all of these. If `raise_error` is necessary we should refactor some of the into the base class and just require `_found()` method.

PR #48, opened 2019-03-13
CI: Tries to prevent cache misses due to index updates

dgasmith:
I have yet to see a case where we have a canonical cache hit without a new upload. This attempts to remove the upload requirements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,8 @@ before_cache:
 
   - conda deactivate
   - conda remove --name test --all
-  - conda clean --index-cache --lock --packages --force-pkg-dirs
+  - conda clean --all
+  #- conda clean --index-cache --lock --packages --force-pkg-dirs
 
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,6 @@ language: generic
 # Run jobs on container-based infrastructure, can be overridden per job
 dist: xenial
 
-before_cache:
-  # if I understand internet correctly, this is equiv to "after_success"
-  - |
-      if [[ $TRAVIS_TEST_RESULT = 0 ]]; then
-        codecov
-      fi
-  - conda deactivate
-  - conda remove --name test --all
 cache:
   timeout: 1000
   directories:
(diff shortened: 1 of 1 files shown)


dgasmith:
Still rebuilding caches. Not sure what is changing, trying a clean all.

loriab:
started commenting on this one this morning and apparently never got back to it. I think you need to keep packages and tarballs, otherwise the act of creating the environment is going to change them. Think the `clean --index-cache` is promising. potentially no-rebuilt-cache only going to happen on second execution, I think.

dgasmith:
Right, the first time will always build. This could potentially be a way forward as well is that a PR has a "frozen" environment.

Lnaden:
The problem with the travis caches is they all operate under the following rule:

"If *any* file in the cache is changed, the *entire* cache is rebuilt"

dgasmith:
Right, which is why we are playing games trying to prevent that.

loriab:
those `pip install`s may be interfering, too

dgasmith:
The current pattern at least gets in base conda through the cache.

loriab:
anyone know if each travis "lane" gets its own cache? miniconda's nice but mkl & libmints are my big concerns.

dgasmith:
Can confirm, each gets its own lane.

loriab:
let me know when you're not working on it for a spell, and I'd like to try it out. disastrous to both be experimenting.

dgasmith:
Feel free!

dgasmith:
In the future, probably the way to go is to pull a Docker container that contains the Dev environment and build ontop of that.

loriab:
docker's great for stable env. but what I really want to see from travis is latest-env subject to known constraints.

dgasmith:
Right, this would just get the base. We would pin MKL/CUDA/Psi/etc to keep the main pulls to a minimum. Docker images are pretty easy mint and continuously roll. Honestly too bad we can use their artifact tech here... would be perfect.

loriab:
Ok, I'm done experimenting. The `--force-pkgs-dirs` was missing plural. But it doesn't do what wanted. My miniconda doesn't even have those extra .q and .json files, so that may be a new thing. What I just pushed is a hack b/c I can't see its consequences b/c I can't see a tidy upgraded/downloaded/installed output b/c it's `conda env create`, not my friend `conda create`. But it doesn't unnecessarily hit the cache: 2nd time through https://travis-ci.org/MolSSI/QCEngine/builds/506036779?utm_source=github_status&utm_medium=notification .

dgasmith:
Same, might consider blowing away the caching?

PR #50, opened 2019-03-14
Psi4: Safer multiprocessing by move Psi temps to scratch

dgasmith:
Seems to be some conflicts when multiprocessing on a single node.

@loriab If I understand this correctly this will build a temporary directory in the supplied folder (if not none)?

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -6,6 +6,8 @@ from qcelemental.models import FailedOperation, Result
 
 from .executor import ProgramExecutor
 
+from ..util import scratch_directory
+
 
 class Psi4Executor(ProgramExecutor):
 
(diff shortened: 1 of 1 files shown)


loriab:
Re-read your question, @dgasmith. `util.scratch_directory` _always_ creates a directory in the `parent` folder (which could be `/tmp` or given). only choices are whether child directory is anonymous (`True`) or known-name (str).
```
parent            child   -->  creates
None              None    --> /tmp/tmpsldkfjksjdls/
/scratch/johndoe  None    --> /scratch/johndoe/tmpsldkfjsdkls/
None              myqcjob --> /tmp/myqcjob/
/scratch/johndoe  myqcjob --> /scratchjohndoe/myqcjob
```

dgasmith:
Roger that, what I derived from your answer. Can we add the above blob to the docs on `scratch_directory`?

PR #51, opened 2019-03-15
add examples to `util.scratch_directory` docstring

loriab:
think default of `child` should move to `None`?


dgasmith:
I would vote on None for consistency.

loriab:
ok, tests healed.

looks like json_out that's been registered in psi https://github.com/psi4/psi4/blob/master/psi4/driver/json_wrapper.py#L223 not too happy to be missing after `scratch_directory(messy=False)` https://travis-ci.org/MolSSI/QCEngine/jobs/506562832#L569

dgasmith:
Always the subprocess option.

PR #52, opened 2019-03-15
Psi4 run from CLI

dgasmith:
Moves Psi4 to run from the CLI to prevent possible multiprocessing issues and fixes a small memory leak after thousands of runs.

This does slow down Psi4 by adding a ~0.5 second startup cost.

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -3,6 +3,8 @@ Calls the Psi4 executable.
 """
 import json
 
+from typing import Dict
+
 from qcelemental.models import FailedOperation, Result
 
 from ..util import scratch_directory, execute, which, popen, safe_version, parse_version
@@ -19,6 +21,7 @@ class Psi4Executor(ProgramExecutor):
         "node_parallel": False,
         "managed_memory": True,
     }
+    version_cache: Dict[str, str] ={}
 
     class Config(ProgramExecutor.Config):
         pass
@@ -38,10 +41,14 @@ class Psi4Executor(ProgramExecutor):
     def get_version(self) -> str:
         self.found(raise_error=True)
 
-        with popen([which('psi4'), '--version']) as exc:
-            exc["proc"].wait(timeout=5)
+        which_psi = which('psi4')
+        if which_psi not in self.version_cache:
+            with popen([which('psi4'), '--version']) as exc:
+                exc["proc"].wait(timeout=5)
+            self.version_cache[which_psi] = exc["stdout"]
 
-        candidate_version = exc["stdout"]
+        candidate_version = self.version_cache[which_psi]
+
         if "undef" in candidate_version:
             raise TypeError(
                 "Using custom build without tags. Please pull git tags with `git pull origin master --tags`.")
@@ -73,7 +80,7 @@ class Psi4Executor(ProgramExecutor):
                 input_data["keywords"]["reference"] = "uhf"
 
             # Execute the program
-            success, output = execute(["psi4", "--json", "data.json"], {"data.json": json.dumps(input_data)}, ["data.json"])
+            success, output = execute([which("psi4"), "--json", "data.json"], {"data.json": json.dumps(input_data)}, ["data.json"])
 
             if success:
                 output_data = json.loads(output["outfiles"]["data.json"])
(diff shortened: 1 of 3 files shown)


qcengine/programs/psi4.py
```diff
@@ -53,36 +60,39 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
...
         self.found(raise_error=True)
-        import psi4
 
         # Setup the job
-        input_model = input_model.copy().dict()
-        input_model["nthreads"] = config.ncores
-        input_model["memory"] = int(config.memory * 1024 * 1024 * 1024 * 0.95)  # Memory in bytes
-        input_model["success"] = False
-        input_model["return_output"] = True
+        input_data = input_model.json_dict()
```
loriab: like this. makes sure nothing unserializable lurking and bysteps the copy at the same time

qcengine/programs/psi4.py
```diff
@@ -53,36 +60,39 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+        input_data["nthreads"] = config.ncores
+        input_data["memory"] = int(config.memory * 1024 * 1024 * 1024 * 0.95)  # Memory in bytes
+        input_data["success"] = False
+        input_data["return_output"] = True
 
-        if input_model["schema_name"] == "qcschema_input":
-            input_model["schema_name"] = "qc_schema_input"
+        if input_data["schema_name"] == "qcschema_input":
+            input_data["schema_name"] = "qc_schema_input"
```
loriab: as of 1.3, psi will take either (changes to qc_ for qcel 0.1.3's sake)

qcengine/programs/psi4.py
```diff
@@ -53,36 +60,39 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+                    output_data["extras"] = {}
 
-            output_data["extras"]["local_qcvars"] = output_data.pop("psi4:qcvars", None)
-            if output_data["success"] is False:
-                output_data["error"] = {"error_type": "internal_error", "error_message": output_data["error"]}
+                output_data["extras"]["local_qcvars"] = output_data.pop("psi4:qcvars", None)
+                if output_data["success"] is False:
+                    output_data["error"] = {"error_type": "internal_error", "error_message": output_data["error"]}
+            else:
+                output_data = input_data
```
loriab: is this a general principle that upon error, output_schemus should equal input_schemus and any new info should be routed to the error field?

Thinking about CBS-flavor output_schemus where one job does not converge. Possible to return every other subjob in the output_schemus and just skip the assembly step. But the notion of returning output=input+error can be good b/c in qcf, all the subjobs will still query as run and take no time.

qcengine/programs/psi4.py
dgasmith: Yea, but doesn't hurt anything at the moment and is still backward compatible.

qcengine/programs/psi4.py
dgasmith: In general yes. So for a failed CBS-flavor I would think we would pack all of the other jobs into `extra` still?

qcengine/util.py
```diff
@@ -474,3 +473,19 @@ def which(command, return_bool=False):
         return bool(ans)
     else:
         return ans
+
+
+def safe_version(*args, **kwargs):
+    """
+    Package resources is a very slow load
+    """
+    import pkg_resources
```
loriab: `pkg_resources` has widespread use. It's not really avoidable in an engine run, right? if so, not following how this is better than an outright import

loriab: that is, I renovated found and get_version for dftd3 and psi4 as a model, but rdkit and torchani should probably follow suit.

dgasmith: It will be used, but doesn't take an extra `0.4` seconds at init time. For example this will slow down QCFractal import by 0.4 seconds even it is never used.

qcengine/programs/psi4.py
loriab: ok, so principle is upon failure output_schemus == input_schemus with exceptions of success/error/extras fields. got it.

qcengine/util.py
loriab: hmpf, that is slow

PR #54, opened 2019-03-15
fix cased testing of psi4 keywords

loriab:
qcengine not respecting `REFERENCE` in psi4 input for open-shell. for posterity, test script below, but no good reason to inspect it now.

I patched it here, but the greater question is is case something the runners in qcengine need to be careful of or is it something that the qcel model should validate into a uniform case. Certainly one shouldn't mess with the case of keyword _values_, but keyword _names_ are generally uniform wrt case.

```
import pprint

inp = {'driver': 'energy',
# 'return_output': True,  # REMOVE FOR ENGINE
 'schema_name': 'qcschema_input',
 'schema_version': 1,
 'keywords': {'DF_SCF_GUESS': 0, 'REFERENCE': 'ROHF', 'SCF_TYPE': 'PK'},
 #'keywords': {'DF_SCF_GUESS': 0, 'reference': 'ROHF', 'SCF_TYPE': 'PK'},
 'model': {'basis': 'aug-cc-pvqz', 'method': 'scf'},
 'molecule': {'atom_labels': [''],
              'atomic_numbers': [3],
              'fix_com': True,
              'fix_orientation': True,
              'fragment_charges': [0.0],
              'fragment_multiplicities': [2],
              'fragments': [[0]],
              'geometry': [0.0, 0.0, 0.0],
              'mass_numbers': [7],
              'masses': [7.0160034366],
              'molecular_charge': 0.0,
              'molecular_multiplicity': 2,
              'name': 'Li',
              'provenance': {'creator': 'QCElemental',
                             'routine': 'qcelemental.molparse.from_string',
                             'version': 'v0.3.3'},
              'real': [True],
              'symbols': ['Li']},
}

# EITHER run Engine

import qcengine as qcng
ans = qcng.compute(inp, "psi4")
pprint.pprint(ans.dict())
scf = ans.dict()['return_result']
print('ENGINE')
print(qcng.util.which('psi4'), qcng.util.which_import('psi4'))
print(scf)

# Results
# /home/psilocaluser/gits/hrw-quaternary/objdir37/stage/bin/psi4 /home/psilocaluser/gits/hrw-quaternary/objdir37/stage/lib/psi4/__init__.py
# -7.365043815990735


# OR run Psi4

#import psi4
#from psi4.driver import json_wrapper
#ans = psi4.json_wrapper.run_json(inp)
#pprint.pprint(ans)
#scf = ans['return_result']
#print('RUN_JSON')
#print(psi4.__file__)
#print(scf)

# Results
# /home/psilocaluser/gits/hrw-quaternary/objdir37/stage/lib/psi4/__init__.py
# -7.432695429345809
```

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -76,7 +76,8 @@ class Psi4Executor(ProgramExecutor):
 
         if parse_version(self.get_version()) > parse_version("1.2"):
 
-            if (input_model.molecule.molecular_multiplicity != 1) and ("reference" not in input_data["keywords"]):
+            caseless_keywords = {k.lower(): v for k, v in input_model["keywords"].items()}
+            if (input_model.molecule.molecular_multiplicity != 1) and ("reference" not in caseless_keywords):
                 input_data["keywords"]["reference"] = "uhf"
 
             # Execute the program
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -76,7 +76,7 @@ class Psi4Executor(ProgramExecutor):
 
         if parse_version(self.get_version()) > parse_version("1.2"):
 
-            caseless_keywords = {k.lower(): v for k, v in input_model["keywords"].items()}
+            caseless_keywords = {k.lower(): v for k, v in input_model.keywords.items()}
             if (input_model.molecule.molecular_multiplicity != 1) and ("reference" not in caseless_keywords):
                 input_data["keywords"]["reference"] = "uhf"
 


dgasmith:
QCFractal has this [capability](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/models/common_models.py#L110) that we homogenize for the database. Not sure about here however.

loriab:
Sounds like a case of the perils of unexpected entry points. When one goes through QCF, keyword keys get normalized to lowercase and (on the other issue) molecules get normalized to 8 digits, which makes sense for database/eq structures. But when one bypasses QCF by hitting engine directly, engine's keyword testing assumptions are violated. Sounds like ultimately, qcf keyword validation needs to happen at engine level? But more importantly, GEOMETRY_NOISE needs to be a parameter passed down to qcel model.Molecule so that qcng and qcf can handle findif.

loriab:
I should note that for the moment, this engine trouble is easily avoided and the qcel trouble is one-liner easy, though not so clean, to avoid. So this can be handled post-ACS.

dgasmith:
Im not convinced that this is too much of an issue. QCEngine in general doesn't handle case on keywords unless a program demands it (e.g. a program must have keywords in upper case). QCFractal is a bit of a different beast as we need to hash keyword sets.

loriab:
It's not a common issue b/c engine isn't probing keywords much -- only implementing a quick "(auto)" for open-shell reference. But engine taking a valid json input and running a different calc (uhf) than that which the program would have run (rohf) with the same json is a definite issue.

PR #58, opened 2019-03-21
TeraChem: Added environment variables. QCEngine can now drive terachem 

ffangliu:
Added test cases for the executor. Tests are skipped if TeraChem environment variable not found.


loriab:
perhaps modify `TeraChemExecutor`'s `found` and `version` fns to model psi4's, which were recently updated. that tests on `$PATH` rather than environment. Then add terachem to https://github.com/MolSSI/QCEngine/blob/master/qcengine/testing.py#L47-L72 so you can use `@using_terachem` above the pytest tests, rather than the skipif.

ffangliu:
Remove the environment variable related stuff and TeraChem Executor can still work. Added the found and get_version functions as recommended by @loriab

PR #59, opened 2019-03-22
Molpro: (WIP) Added CCSD support to output parsing

sjrl:
Added CCSD support to output parsing. However, it looks like the ccsd keys (same as mp2 except with the ccsd label) need to be added to the supported keys. I believe in QCSchema?

- [ ] Add support to QCSchema (?) for CCSD quantities

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -106,7 +106,7 @@ class MolproExecutor(ProgramExecutor):
         # SCF maps
         scf_energy_map = {"Energy": "scf_total_energy"}
         scf_dipole_map = {"Dipole moment": "scf_dipole_moment"}
-        scf_extras = {"method": "molpro_scf_method"}
+        # scf_extras = {"method": "molpro_scf_method"}
 
         # MP2 maps
         mp2_energy_map = {
@@ -116,7 +116,7 @@ class MolproExecutor(ProgramExecutor):
             "triplet pair energy": "mp2_triplet_pair_energy"
         }
         mp2_dipole_map = {"Dipole moment": "mp2_dipole_moment"}
-        mp2_extras = {}
+        # mp2_extras = {}
 
         # CCSD maps
         ccsd_energy_map = {
@@ -126,7 +126,7 @@ class MolproExecutor(ProgramExecutor):
             "triplet pair energy": "ccsd_triplet_pair_energy"
         }
         ccsd_dipole_map = {"Dipole moment": "ccsd_dipole_moment"}
-        ccsd_extras = {}
+        # ccsd_extras = {}
 
         # Compiling the method maps
         supported_methods = {"HF", "RHF", "MP2", "CCSD"}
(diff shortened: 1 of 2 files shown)


dgasmith:
Yes, we have not yet added CCSD keys to the QCSchema. Should be a pretty simple add, you can open a PR modifying [here](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/results.py#L41) if you wish.

sjrl:
While I was working on this today, I realized I may have mapped the Molpro and Molssi quantities incorrectly. Do you know if `opposite_spin_correlation` is the same as `singlet pair energy` or `triplet pair energy` (which is the notation used in Molpro. Doing a quick search I couldn't easily see how the naming conventions match up.

loriab:
Yes, Molpro --> Psi4 (which in this case "wins" on precedent) needs a bit of math for same-spin/opposite-spin quantities.

```
            #}   
            if ($line =~ /CCSD correlation energy/) {
               @ltemp = split( /\s+/, $line);
               $tempc = @ltemp[4];
            }
            if ($line =~ /Total correlation energy/) {
               @ltemp = split( /\s+/, $line);
               $tempd = @ltemp[4];
            }
            if ($line =~ /MP2 triplet pair energy/) {
               @ltemp = split( /\s+/, $line);
               $tempf = @ltemp[5];
            }
            #if ($line =~ / TEMP3            = /) {  # only printed with special instructions
            #   @ltemp = split( /\s+/, $line);
            #   $temph = @ltemp[3];
            #}
            if ($line =~ /CCSD triplet pair energy/) {
               @ltemp = split( /\s+/, $line);
               $tempg = @ltemp[5];
            }
         }  

         $tempf *= 2/3;
         #$temph *= 2/3;
         $tempg *= 2/3;
```

sjrl:
Sorry I'm still a little confused. Are the final `$tempf` and `$tempg` quantities the conversion between triplet pair energy and the same spin energy?

loriab:
something like that. if you can use a scaling with 2, 3 to get values like https://github.com/psi4/psi4/blob/master/tests/pytest/test_mp2.py#L13-L24 for https://github.com/psi4/psi4/blob/master/tests/pytest/test_mp2.py#L52-L54, you're in good shape.

sjrl:
I think I have it `(1/3)*triplet + singlet = opposite` and `(2/3)*triplet = same`. I'll update my pull request with the correct conversion.

dgasmith:
This looks good! We need to crank out a new QCElemental release soon so that this can go through.

sjrl:
Okay sounds good! I'll keep an eye out for the QCElemental release.

dgasmith:
QCElemental v0.4 is now out. Do we need to do anything here to get this in?

sjrl:
It should be good to pull unless you'd like me to update it with the current master branch.

dgasmith:
Please do, it's been long enough that we want to run this through CI once more.

sjrl:
Okay everything looks good. Also just wanted to give a heads up, one of the testjobs I have added for Molpro does not pass, because it is unable to find the CCSD energy from the xml file when calling a gradient. I have an idea of how to fix this (maybe text parsing), but I was going to wait to fix that at a later date.

PR #60, opened 2019-03-22
half a gamess interface

loriab:
bring your own input file. known to run and parse HF (rhf/uhf/rohf) and FCI. much regex from Nuwan de Silva


loriab:
The gamess interface is no further along, but I'm recommending this be merged because it's not hurting anything and, more importantly, I've standardized the `found` and `get_version` fns across the programs. Also added version stubs for cfour and nwchem. For those and gamess, have to run a miniature dummy file to get the header printed to scrape for version.

qcengine/programs/util/pdict.py
```diff
@@ -0,0 +1,112 @@
+from decimal import Decimal, ROUND_FLOOR, ROUND_CEILING
+
+import numpy as np
+
+# This file has considerable history in the quantum chemistry package Psi4
```
dgasmith: Having a util folder and a `util.py` might be getting out of hand.

qcengine/programs/__init__.py
```diff
@@ -63,3 +66,6 @@ def list_available_programs() -> Set[str]:
 register_program(DFTD3Executor())
 register_program(TeraChemExecutor())
 register_program(MP2DExecutor())
+register_program(GAMESSExecutor())
+register_program(NWChemExecutor())
+register_program(CfourExecutor())
```
dgasmith: We should not register these until they are complete.

qcengine/programs/cfour.py
```diff
@@ -0,0 +1,59 @@
...
+from .executor import ProgramExecutor
+#from ...extras import provenance_stamp
+
+pp = pprint.PrettyPrinter(width=120, compact=True, indent=1)
+
+
+class CfourExecutor(ProgramExecutor):
+
+    _defaults = {
+        "name": "Cfour",
```
dgasmith: I always thought this was `CFOUR`?

qcengine/programs/gamess/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .runner import GAMESSExecutor
```
dgasmith: It doesn't seem we need a whole folder for this? No need to increase the depth if we can add everything to a single file.

loriab: The ones with parsing are going to get really long. What’s here for gamess is only half of a program that does  a few methods. I’d rather reconsider file layout at the end.

qcengine/programs/util/pdict.py
loriab: I agree. But these are internal (non exported) utils. What about renaming the outer util “executing” in keeping with recent trends? (Would need a qcel bump)

PR #61, opened 2019-04-16
TeraChem: modified input parser to generate xyz files in Angstrom

ffangliu:


qcengine/programs/terachem.py
```diff
@@ -84,14 +84,11 @@ def get_version(self) -> str:
...
         xyz_file = [] 
-        s = str(len(input_model.molecule.symbols))  
-        xyz_file.append(s+"\n")
-        for sym, geom in zip(input_model.molecule.symbols, input_model.molecule.geometry):
-            s = "{:<4s} {:>{width}.{prec}f} {:>{width}.{prec}f} {:>{width}.{prec}f}".format(sym, *geom, width=14, prec=10)
-            xyz_file.append(s)
-  
+        natoms = len(input_model.molecule.symbols)  
+        xyz_file.append(str(natoms)+"\n")
+        xyz_file = xyz_file + input_model.molecule.pretty_print().strip('\n').split('\n')[-natoms:]
```
dgasmith: @loriab Is this right string formatter to use here? 

Otherwise LGTM

loriab: something in the `input_model.molecule.to_string(dtype=??)` family would be best. what does it need to look like? (ghost atoms, fragments, fix_symmetry/com/orientation, in particular.) Ah, actually looks like lines 91--101 should be in a new `dtype=terachem` in https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/to_string.py#L137.

Really want to write geometry in Angstrom? That's fine if it's all terachem will accept. But it's cleaner physconst-wise to stick with Bohr.

ffangliu: TeraChem can accept Bohr by setting a keyword "unit bohr". By default, it assume the xyz file it reads in is Angstrom. I thought it could help avoid error if the file is written into angstrom. Alternatively I can make the input parser always add the keyword "unit bohr" when generating input file.

loriab: I'd go for Bohr.

Working on a formatter for terachem ...

If you fed terachem the below, will it reject b/c of the `au`? I hate to write files that don't convey the units.
```
2 au
sldfkjsldks
H 0. 0. 0.
H 2. 0. 0.
```

loriab: if terachem _doesn't_ hate the `au` in above, let me know so this can be cleaner.

in any case, see https://github.com/MolSSI/QCElemental/pull/55 . You should be able to `xyz_file = input_model.molecule.to_string(dtype='terachem')`. Feel free to take over, duplicate, or PR to that PR at qcel.

It's awkward that terachem won't accept coordinates in the input file. otherwise could write more over at qcel. are there any keywords for controlling orientation or symmetry?

ffangliu: > I'd go for Bohr.
> 
> Working on a formatter for terachem ...
> 
> If you fed terachem the below, will it reject b/c of the `au`? I hate to write files that don't convey the units.
> 
> ```
> 2 au
> sldfkjsldks
> H 0. 0. 0.
> H 2. 0. 0.
> ```

Hi Lori, this format with `au` works fine for TeraChem. I would try PR to that PR in qcel. 

There's no keyword for controlling orientation or symmetry. TeraChem does not enforce any symmetry or orientation.

loriab: great on the `au`! in that case, turn line 142 back into https://github.com/MolSSI/QCElemental/pull/55/files#diff-04c950a9fd69fdef4dff905341f9d4e4R76, please. even if the file is short-lived, I'll feel better not springing bohr in a format that's defined to be ang.

> There's no keyword for controlling orientation or symmetry. TeraChem does not enforce any symmetry or orientation.

Interesting. Out of curiosity, does this mean that terachem uses exactly the input orientation or (more common) it moves all mols to an internal standard orientation and returns all quantities at that orientation?

ffangliu: Yes, terachem uses exactly the input orientation, and returns all quantities at that orientation. Although during geometry optimization the geometry may be transformed into different types of internal coordinates.

PR #62, opened 2019-04-22
TeraChem: Changed the input parser to call qcelemental to_string method with bohr unit

ffangliu:
Change to use to_string and keep the bohr unit. Print "bohr" in the first line of the generated xyz file. Skip one test that compare the output json file (fail because I'm using the qcelemental github version to make sure the to_string method is available. The version is newer than the conda-forge qcelemental version, and won't succeed when comparing the provenance)

### DIFF
--- a/qcengine/programs/terachem.py
+++ b/qcengine/programs/terachem.py
@@ -84,14 +84,11 @@ class TeraChemExecutor(ProgramExecutor):
 
     def build_input(self, input_model: 'ResultInput', config: 'JobConfig', 
                     template: Optional[str]=None) -> Dict[str, Any]:
-        #Write the geom xyz file
+        #Write the geom xyz file with unit Angstrom
         xyz_file = [] 
-        s = str(len(input_model.molecule.symbols))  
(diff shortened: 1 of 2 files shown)


loriab:
btw, if you end up with a local branch with name `topicnew` that you want to attach to an existing PR associated with branch `topic` (thus preserving the discussion history) and don't mind that the latter gets overwritten remotely on your GH fork, you can `git push origin topicnew:topic`. I don't know if that was the situation that led to closing 61 and opening 62, but it's a handy procedure to know.

ffangliu:
> btw, if you end up with a local branch with name `topicnew` that you want to attach to an existing PR associated with branch `topic` (thus preserving the discussion history) and don't mind that the latter gets overwritten remotely on your GH fork, you can `git push origin topicnew:topic`. I don't know if that was the situation that led to closing 61 and opening 62, but it's a handy procedure to know.

Thanks for the tips for attaching to existing PR. I closed the older PR manually. I'll keep in mind the new way of attaching.

PR #63, opened 2019-04-25
remove which, which_import

loriab:
remove "which" fns from qcng after adding to qcel in MolSSI/QCElemental#58


loriab:
this is RTG after qcel v0.4.0 conda package available and CI retriggered and passing.

I have no urgency for a qcng release, but [this'll](https://github.com/conda-forge/qcengine-feedstock/blob/master/recipe/meta.yaml#L28) need bumping when that happens.

PR #64, opened 2019-04-30
Add the python binary directory to `which`'s PATH

Lnaden:
Appends the directory of the executing Python bin to the path of the
search for `which` function. This should help with a corner case
on clusters which don't export the PATH variable correctly on nodes
of the executing environment since the absolute Python binary's path
ensures proper imports.

Also cleaned up some unused imports and `__all__` statements.


loriab:
per #63 and MolSSI/QCElemental#58, `which`, `which_import`, `parse_version`, `safe_version` have departed for qcel, so this PR needs to split, methinks.

Lnaden:
Obsolete on this repo from #63 
Superseded by molssi/qcelemental#60

PR #65, opened 2019-05-07
Torchani Update

dgasmith:
Adds ANI1-x and ANI1-ccx models to torchani.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,6 @@ QCEngine
 
 Quantum chemistry program executor and IO standardizer ([QCSchema](https://github.com/MolSSI/QCSchema)) for quantum chemistry.
 
-Available Compute Programs:
- - [Psi4](http://www.psicode.org)
- - [RDKit](http://rdkit.org)
- - [TorchANI](https://github.com/aiqm/torchani)
-
-Available Procedures:
- - [Geometric](https://github.com/leeping/geomeTRIC)
-
-
 # Example
 
 A simple example of QCEngine's capabilities is as follows:
@@ -44,7 +35,7 @@ H  0.0  1.494  1.027
 These input specifications can be executed with the ``compute`` function along with a program specifier:
 
 ```python
->>> ret = qcng.compute(inp, "psi4", return_dict=False)
+>>> ret = qcng.compute(inp, "psi4")
 ```
 
 The results contain a complete record of the computation:
--- a/qcengine/programs/__init__.py
+++ b/qcengine/programs/__init__.py
@@ -4,13 +4,13 @@ Imports the various compute backends
 
 from typing import List, Set
 
+from .dftd3 import DFTD3Executor
+from .molpro import MolproExecutor
+from .mp2d import MP2DExecutor
 from .psi4 import Psi4Executor
 from .rdkit import RDKitExecutor
-from .torchani import TorchANIExecutor
-from .molpro import MolproExecutor
-from .dftd3 import DFTD3Executor
 from .terachem import TeraChemExecutor
-from .mp2d import MP2DExecutor
+from .torchani import TorchANIExecutor
 
 __all__ = ["register_program", "get_program", "list_all_programs", "list_available_programs"]
 
--- a/qcengine/programs/executor.py
+++ b/qcengine/programs/executor.py
@@ -3,6 +3,8 @@ from typing import Any, Dict, Optional
 
 from pydantic import BaseModel
 
+from ..util import which_import
+
 
 class ProgramExecutor(BaseModel, abc.ABC):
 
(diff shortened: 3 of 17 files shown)


qcengine/programs/__init__.py
```diff
@@ -4,13 +4,13 @@
...
+from .molpro import MolproExecutor
+from .mp2d import MP2DExecutor
 from .psi4 import Psi4Executor
 from .rdkit import RDKitExecutor
-from .torchani import TorchANIExecutor
-from .molpro import MolproExecutor
-from .dftd3 import DFTD3Executor
 from .terachem import TeraChemExecutor
-from .mp2d import MP2DExecutor
+from .torchani import TorchANIExecutor
```
loriab: did you mean to undo mp2d?

qcengine/programs/executor.py
```diff
@@ -21,23 +23,40 @@ class Config:
...
+
+        Parameters
+        ----------
+        raise_error : bool, optional
+            If True, raises an error if the program cannot be found.
+
+        Returns
+        -------
+        bool
+            Returns True if the program was found, False otherwise.
```
loriab: need inherited docstrings :-)

qcengine/programs/rdkit.py
```diff
@@ -25,17 +26,19 @@ class Config(ProgramExecutor.Config):
     def __init__(self, **kwargs):
         super().__init__(**{**self._defaults, **kwargs})
 
+    @staticmethod
+    def found(raise_error: bool=False) -> bool:
+        return which_import("rdkit", return_bool=True, raise_error=raise_error)
+
```
loriab: I think you want logic more like https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/psi4.py#L34-L39 . The idea is that sometimes you want to know yes/no found and sometimes you want to find-or-die, and this is the place to consolidate the latter's error msg.

qcengine/util.py
```diff
@@ -457,9 +457,12 @@ def which_import(plug, return_bool=False):
 
     """
     import pkgutil
-    plug_spec = pkgutil.find_loader(plug)
+    module_spec = pkgutil.find_loader(module)
+
+    if module_spec is None:
+        if raise_error:
+            raise ModuleNotFoundError(f"Could not find module '{module}'' in Python path.")
```
loriab: fwiw, I tried and rejected the 3-mode return from `which_import` of bool, path/None, raise_error in favor of handling only the first two at this level (remember `which` and `which_import` need to follow the same pattern). Most of the time, you want to raise a more specific error and so you'd end up catching and re-raising anyways. just my opinion.

qcengine/programs/executor.py
dgasmith: They do.

qcengine/programs/__init__.py
dgasmith: It moved.

qcengine/util.py
dgasmith: Hmm, the 99% case seems to be that you want to throw here? Easy to do in `which` as well. 

The Psi4 logic that you linked to above is reproduced in these lines. I didn't think rewriting the same thing over and over was healthy.

qcengine/programs/executor.py
```diff
@@ -21,23 +23,40 @@ class Config:
...
+        str
+            Return a valid, safe python version string.
+        """
         pass
 
 ## Computers
 
     def build_input(self, input_model: 'ResultInput', config: 'JobConfig',
-                    template: Optional[str] = None) -> Dict[str, Any]:
+                    template: Optional[str]=None) -> Dict[str, Any]:
```
loriab: Very minor, but is your auto-format working just from the yapf settings in setup.cfg? I turned on `yapf` autoformatting in an IDE and noticed it did exactly the opposite of the above, transforming into `asdf: List[str] = None`. I've checked that it's properly set to use yapf, updated to yapf 0.27, and observed that it's picking up on the 119 column limit.

qcengine/util.py
loriab: Yes, often you want to throw, but you want a program specific message like https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/molecule.py#L236 (as opposed to the rather bad one I linked to for psi).

For the fallback case when the generic error will do, I'm good with the `raise_error` arg in `which`/`which_import` like you have. I'll implement that in the qcel version. But for most in-module calls to which*, I think we should be using the T/F and writing good error strings (e.g., `conda install psi4 -c psi4`. That was my aim in localizing all the checks to each engine's `.found` method.

Side question is should there be a generalized error? `which_import` would be `ModuleNotFoundError` but `which` is natively just `None`. And you don't want the error changing when go btwn `psi4` and `import psi4`. (These are some of the reasons I punted on raise_error before.)

dgasmith: I say `which` should just raise a `ModuleNotFoundError` (IMO). Functionally the same thing here.

loriab: Ok, that will require no changes to https://github.com/MolSSI/QCElemental/pull/63 then.

How pressing is this PR? Easiest would be to wait on new qcel release, then #63 merge, then rebase, then patch up the `found` messages. But that's days away.

qcengine/programs/executor.py
dgasmith: Ok turns out I need to update from 0.16 (!). So we now have the spaces... Not sure I am a fan of the spaces, but I guess we go with it.

qcengine/util.py
dgasmith: Not a huge hurry at the moment.

dgasmith:
Should be ready for review.

qcengine/programs/psi4.py
```diff
@@ -35,7 +34,7 @@ def found(raise_error=False) -> bool:
         is_found = which("psi4", return_bool=True)
 
         if not is_found and raise_error:
-            raise ModuleNotFoundError("Could not find Psi4 in the Python path.")
+            raise ModuleNotFoundError("Could not find Psi4 in the Python path. Please 'conda install psi4 -c psi4'.")
```
loriab: I think this whole fn (and probably its counterparts) collapses into:
```
def found(raise_error=False) -> bool:
    return which("psi4", return_bool=True, raise_error=raise_error, raise_msg='Please `conda install psi4 -c psi4`.')
```

README.md
```diff
@@ -10,15 +10,6 @@ QCEngine
 
 Quantum chemistry program executor and IO standardizer ([QCSchema](https://github.com/MolSSI/QCSchema)) for quantum chemistry.
 
-Available Compute Programs:
```
Lnaden: /why remove this block from the README? Should it not keep expanding as we get more?

dgasmith: Best to keep it in the primary docs (IMO) so that we do not have two copies running around. Happy to change this as desired.

Lnaden: Good reason. Keep it as is then

qcengine/programs/psi4.py
loriab: This didn't get addressed? It's a desirable simplification, I think.

PR #67, opened 2019-05-10
TeraChem: Improvement of output parser: Parse stdout into Result.

ffangliu:
Also improved how the last SCF iteration info is extracted to make sure it works for FON and other more complicated calculations than regular SCF.

### DIFF
--- a/qcengine/programs/terachem.py
+++ b/qcengine/programs/terachem.py
@@ -10,6 +10,7 @@ from .executor import ProgramExecutor
 from ..util import which, popen, parse_version, safe_version
 import qcengine.util as uti
 from qcelemental.models import FailedOperation
+import re
 
 
 class TeraChemExecutor(ProgramExecutor):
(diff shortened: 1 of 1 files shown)


loriab:
You may find regex very helpful for this type of parsing. In particular, the `NUMBER` string is great for float matching https://github.com/MolSSI/QCElemental/blob/68aa6846170bbfcea62d42706598965127d6f189/qcelemental/molparse/regex.py#L18. Can form lines like https://github.com/MolSSI/QCEngine/pull/60/files#diff-1ce502a98dede4b94883a44719956b3bR259

ffangliu:
Thanks for the suggestion! I'll try out the regex.

qcengine/programs/terachem.py
```diff
@@ -172,12 +150,20 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
                        gradients.append( float(x) )
 
         # Look for the last line that is the SCF info 
+        DECIMAL = r"""(
+          (?:[-+]?\d*\.\d+(?:[DdEe][-+]?\d+)?) |  # .num with optional sign, exponent, wholenum
+          (?:[-+]?\d+\.\d*(?:[DdEe][-+]?\d+)?)    # num. with optional sign, exponent, decimals
+        )"""
+
```
loriab: feel free to `from qcel.molparse.regex import NUMBER` and use it directly

ffangliu: Hi Lori, I didn't directly use the NUMBER definition of `qcel.molparse.regex` because it includes integer numbers, which I would like to exclude here. TeraChem prints FON numbers in between the SCF lines with format "int decimal int decimal int decimal", and real SCF lines are like "int decimal decimal decimal decimal decimal", and I would like to distinguish them.

loriab: makes sense, tks. perils of parsing that we're one `.` away from grabbing wrong number.

dgasmith: @loriab would it be good to put more regex values in QCEl?

loriab: probably so. I found a place in pubchem where I needed the equivalent of `DECIMAL`. More than once or twice and should probably be added to molparse.regex (in string form of course so that can be composed).

Issue #66, opened 2019-05-07
Error classification


dgasmith:
It would be good to add error classification to these models so that can downstream programs can make decisions on what should happen. Several examples:
 - `InputError` - (non-recoverable) error in the user input (e.g., incorrect keyword or method)
 - `SetupError` - (recoverable) example: scratch directory is not writeable 
 - `ConvergenceError` - (recoverable) likely requires options tweaking to enhance iterative convergence.
 - `RandomError` - (recoverable) random seg fault or the like.

Recoverable/non-recoverable in a distributed computing sense where an upstream manager can make the decision to resubmit.

My initial thought here is that we build these as Exception classes so that the `compute` command can capture them and either properly process them into proper JSON error message or let them raise. I am usually not a fan of custom error classes, but here is a good case where there is a variety of different behaviors that you want to elicit depending on the type of error provided.

It would be good to kick around the different error types for a bit before implementing.

loriab:
I _am_ usually a fan of custom error classes for flow management, so I like this idea.

Categories look good, though I suspect they're what the upstream manager wants to know more than than what the throwing program knows about its situation.
* `ResourceError` - (recoverable or non-recoverable) req'd memory not allocatable (rc), available disk not sufficient (nrc if same resource; rc if alternate resource)
* `RandomError` - (rc) PSIO error
* `InputError` - (nrc) what were once `psi4.ValidationError`s

I thought pymatgen was going to be more helpful than it turned out:

https://github.com/materialsproject/pymatgen/blob/76a66990113cf9a11f908e1aba643f19c6abca68/pymatgen/io/abinit/events.py#L767-L877

https://github.com/materialsproject/pymatgen/blob/76a66990113cf9a11f908e1aba643f19c6abca68/pymatgen/io/abinit/tasks.py#L1299-L1304

dgasmith:
Yea, I had poked through there as well without finding too much.

Is the `ResourceError` a `SetupError` or vice versa? Not sure if we need to delineate between the two, either way they are recoverable and rerun on another worker I think.

loriab:
I was thinking `ResourceError` was a more concrete name for `SetupError`.

dgasmith:
So we have:

- InputError - (non-recoverable) error in the user input (e.g., incorrect keyword or method)
- ResourceError - (recoverable) example: scratch directory is not writeable, not enough memory
- ConvergenceError - (recoverable) likely requires options tweaking to enhance iterative convergence.
[...]

PR #69, opened 2019-05-20
Custom Exceptions

dgasmith:
Begins custom exceptions as discussed in #66. This is going to be a somewhat rough process I believe as we have many different categories here and integrating into our current ValueErrors will be interesting. In addition we try to split this into two types of errors:
1) Unknown python exceptions which are handled as we currently have them (traceback + unknown_error type in the JSON return)
2) Known python exceptions where the traceback is omitted and only the known error message is returned.

We also get to simplify the executors somewhat as we can localize issues.

This needs a deep review from @loriab, I am anticipating many small issues/questions here.

### DIFF
--- a/qcengine/exceptions.py
+++ b/qcengine/exceptions.py
@@ -1,40 +1,33 @@
-
 import traceback
 
-from qcelemental.models import ComputeError, FailedOperation
 
 class QCEngineException(Exception):
     """
     Base QCEngine exception, should never be called explicitly.
     """
     error_type = "base_error"
-    header = "QCEngine Base Error:"
+    header = "QCEngine Base Error"
 
-    def __init__(self, message, input_data, return_data):
+    def __init__(self, message: str):
 
         # Call the base class constructor with the parameters it needs
         super().__init__(message)
 
         # Now for your custom code...
-        self.input_data = input_data
-        self.return_data = return_data
+        self.raw_message = message
         self.traceback = traceback.format_exc()
 
-    def format_failed_operation(self):
-        rdata = self.return_data.copy()
+    @property
+    def error_message(self) -> str:
+        return f"{self.header}: {self.raw_message}"
 
-        message = f"{self.header}\n{str(self)}"
-        rdata["error"] = ComputeError(error_type=self.error_type, error_message=self.message)
-
-        return FailedOperation(input_data=self.input_data, **rdata)
 
 class UnknownError(QCEngineException):
     """
     Unknown QCEngine error, the type was not able to be specified.
     """
     error_type = "unknown_error"
-    header = "QCEngine Unknown Error:"
-
+    header = "QCEngine Unknown Error"
 
 
 class InputError(QCEngineException):
@@ -42,7 +35,7 @@ class InputError(QCEngineException):
     Incorrect user parameters, not recoverable. Also may indicate a version issue.
     """
     error_type = "input_error"
-    header = "QCEngine Input Error:"
+    header = "QCEngine Input Error"
 
 
 class ResourceError(QCEngineException):
@@ -51,7 +44,7 @@ class ResourceError(QCEngineException):
     Recoverable on different compute hardware.
     """
     error_type = "resource_error"
-    header = "QCEngine Resource Error:"
+    header = "QCEngine Resource Error"
 
 
 class ConvergenceError(QCEngineException):
@@ -59,7 +52,7 @@ class ConvergenceError(QCEngineException):
     Failed iteration convergence error, likely recoverable with tweak parameters.
     """
     error_type = "convergence_error"
-    header = "QCEngine Convergence Error:"
+    header = "QCEngine Convergence Error"
 
 
 class RandomError(QCEngineException):
@@ -67,4 +60,4 @@ class RandomError(QCEngineException):
     Likely recoverable errors such as segmentation faults or disk io problems.
     """
     error_type = "random_error"
-    header = "QCEngine Random Error:"
+    header = "QCEngine Random Error"
--- a/qcengine/exceptions.py
+++ b/qcengine/exceptions.py
@@ -49,7 +49,7 @@ class ResourceError(QCEngineException):
 
 class ConvergenceError(QCEngineException):
     """
-    Failed iteration convergence error, likely recoverable with tweak parameters.
+    Failed iteration convergence error, likely recoverable with tweaked parameters.
     """
     error_type = "convergence_error"
     header = "QCEngine Convergence Error"
--- a/qcengine/exceptions.py
+++ b/qcengine/exceptions.py
@@ -1,0 +1,52 @@
+from qcelemental.models import ComputeError, FailedOperation
+
+
+class UnknownError(Exception):
+    """
+    Unknown QCEngine error, the type was not able to be specified.
+    """
+    error_type = "unknown_error"
+
+    def __init__(self, message, input_data, return_data):
+
+        # Call the base class constructor with the parameters it needs
+        super().__init__(message)
+
+        # Now for your custom code...
+        self.input_data = input_data
+        self.return_data = return_data
+
+    def format_failed_operation():
+        rdata = self.return_data.copy()
+        rdata["error"] = ComputeError(error_type=self.error_type, error_message=self.message)
+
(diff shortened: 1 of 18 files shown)


qcengine/programs/dftd3/dashparam.py
```diff
@@ -480,7 +482,7 @@ def from_arrays(name_hint=None, level_hint=None, param_tweaks=None, dashcoeff_su
 
     # << 1 >> use name_hint and/or level_hint to determine intended dispersion level
     if name_hint is None and level_hint is None:
-        raise ValueError("""Can't guess -D level without name_hint ({}) or level_hint ({})""".format(
+        raise InputError("""Can't guess -D level without name_hint ({}) or level_hint ({})""".format(
```
loriab: glad to see these return to custom errors

qcengine/programs/mp2d.py
```diff
@@ -240,7 +241,7 @@ def mp2d_harvest(jobrec, modulerec):
             break
     else:
         if not ((real_nat == 1) and (jobrec['driver'] == 'gradient')):
-            raise ValueError('Unsuccessful run. Possibly -D variant not available in dftd3 version.')
+            raise ResourceError('Unsuccessful run. Possibly -D variant not available in dftd3 version.')
```
loriab: Should be `UnknownError` like below. "-D variant" sentence doesn't belong here.

qcengine/programs/psi4.py
```diff
@@ -93,28 +95,33 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
...
 
         # Dispatch errors, PSIO Errors are not recoverable for future runs
         if output_data["success"] is False:
-
-            if "PSIO Error" in output_data["error"]:
-                raise ValueError(output_data["error"])
+            error_message = output_data["error"]["error_message"]
+
+            if "PSIO Error" in error_message:
+                raise ResourceError(error_message)
```
loriab: from slack discussion, PSIO is turning into `RandomError`? I agree.

qcengine/exceptions.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Unknown QCEngine error, the type was not able to be specified.
+    """
+    error_type = "unknown_error"
+    header = "QCEngine Unknown Error"
+
+
+class InputError(QCEngineException):
+    """
+    Incorrect user parameters, not recoverable. Also may indicate a version issue.
```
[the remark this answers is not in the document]
dgasmith: For this in the bottom comment we have two sets of version errors here:
 1) The version is valid for QCEngine, but a particular method or parameter is only available in a higher version number. This would result in an InputError and be very hard for us to distinguish between a normal input and version issue.
2) The version is too old for QCEngine which throws a ResourceError error as below.

qcengine/procedures/base.py
```diff
@@ -27,7 +28,17 @@ def get_procedure(name: str) -> 'BaseProcedure':
...
-    return procedures[name.lower()]
+
+    name = name.lower()
+
+    if name not in procedures:
+        raise InputError(f"Procedure {name} is not registered to QCEngine.")
+
+    ret = procedures[name]
+    if not ret.found():
+        raise ResourceError(f"Procedure {name} is registered with QCEngine, but cannot be found.")
```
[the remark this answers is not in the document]
dgasmith: See above, we might want to change the `ret.found()` to check for versions as well.

qcengine/programs/mp2d.py
```diff
@@ -97,7 +98,7 @@ def run_json(jobrec):
             'error_type': 'ValueError',
             'error_message': """MP2D produces max gradient, not {jobrec['driver']}"""
         }
-        raise ValueError("""MP2D produces max gradient, not {jobrec['driver']}""")
+        raise InputError(f"""MP2D can valid driver options are 'energy' and 'gradient', not {jobrec['driver']}""")
```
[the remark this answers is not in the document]
dgasmith: Sure, I was confused why you have the above. The meta handlers automatically set the error information however.

qcengine/programs/psi4.py
dgasmith: Hmm, could be Resource or Random. Do the error strings differ?

loriab: Apparently PSIO errors should have all the variation one could want (https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libpsio/error.cc). In practice, my mind knows only one PSIO error. So long as it's not on infinite retry, I think a recoverable error is safe and helpful.

qcengine/procedures/base.py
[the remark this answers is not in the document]
loriab: that's starting to sound more like `find_package(module EXACT 1.2 REQUIRED)` syntax https://github.com/psi4/psi4/blob/master/cmake/FindPythonModule.cmake . I'd avoid that complication if possible, I think.

qcengine/exceptions.py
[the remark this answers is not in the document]
loriab: I agree with that separation. Especially when I think of multiple clusters where the software may be heterogeneous, rather than the normal single-cluster of homogeneous software and heterogeneous hardware.

qcengine/procedures/base.py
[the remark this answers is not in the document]
dgasmith: Apologies, I wasn't thinking exact. More in the frame of `Psi4.__version__ >= 1.2.1`. For example it could find Psi4 v1.1, but that should fail.

qcengine/programs/psi4.py
dgasmith: Attempt to refine this particular error based off those error strings.

dgasmith:
@loriab can this get a final review?

qcengine/programs/dftd3/runner.py
```diff
@@ -291,14 +293,14 @@ def dftd3_harvest(jobrec, modulerec):
...
-            raise ValueError("Cannot process ATM results from DFTD3 prior to v3.2.1.")
+            raise ResourceError("Cannot process ATM results from DFTD3 prior to v3.2.1.")
         elif re.match(r""" E6\(ABC\) /kcal,au:""", ln):
             atm = Decimal(ln.split()[-1])
         elif re.match(' normal termination of dftd3', ln):
             break
     else:
         if not ((real_nat == 1) and (jobrec['driver'] == 'gradient')):
-            raise ValueError('Unsuccessful run. Possibly -D variant not available in dftd3 version.')
+            raise ResourceError('Unsuccessful run. Possibly -D variant not available in dftd3 version.')
```
loriab: The "-D variant not available" is only a minor chance since been available since 2016. So probably r`aise UnknownError('Unsuccessful gradient collection.')` like below. But that can be patched up later.

qcengine/programs/psi4.py
```diff
@@ -93,28 +95,38 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+                    # Likely a random error, worth retrying
+                    raise RandomError(error_message)
+            elif "SIGSEV" in error_message:
+                raise RandomError(error_message)
+            elif "TypeError: set_global_option" in error_message:
+                raise InputError(error_message)
+            elif "RHF reference is only for singlets" in error_message:
+                raise InputError(error_message)
+            else:
+                raise UnknownError(error_message)
```
loriab: I'll try to keep an eye on these patterns in ordinary Psi4 runs.

qcengine/programs/torchani.py
```diff
@@ -88,19 +87,13 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
         # Build species
         species = "".join(input_data.molecule.symbols)
         unknown_sym = set(species) - {"H", "C", "N", "O"}
         if unknown_sym:
-            ret_data["error"] = ComputeError(
-                error_type="input_error",
-                error_message="The '{}' model does not support symbols: {}.".format(
-                    input_data.model.method, unknown_sym))
-            return FailedOperation(input_data=input_data.dict(), **ret_data)
+            raise InputError(f"TorchANI model '{input_data.model.method}' does not symbols: {unknown_sym}.")
```
loriab: +"support" symbols

PR #71, opened 2019-05-22
Molpro can now be run by QCEngine

sjrl:
Molpro can now be run by QCEngine and it passes the tests in `test_molpro.py` when QCEngineRecords is present. Except for `ccsd_water_gradient` test because the CCSD energy is missing from the xml file (I'm still working on fixing this).

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -5,8 +5,7 @@ Calls the Molpro executable.
 import xml.etree.ElementTree as ET
 from typing import Any, Dict, Optional
 
-#from qcelemental.models import ComputeError, FailedOperation, Provenance, Result
-from qcelemental.models import Result
+from qcelemental.models import Result, FailedOperation
 
 from ..util import which
 from .executor import ProgramExecutor
@@ -30,7 +29,35 @@ class MolproExecutor(ProgramExecutor):
         super().__init__(**{**self._defaults, **kwargs})
 
     def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
-        pass
+        """
+        Run Molpro
+        """
+        self.found(raise_error=True)
+
+        # Check Molpro version
+        # if parse_version(self.get_version()) < parse_version("1.5"):
+        #     raise TypeError("TeraChem version '{}' not understood".format(self.get_version()))
+
+        # Setup the job
+        job_inputs = self.build_input(input_data, config)
+
+        # Run Molpro
+        exe_outputs = self.execute(job_inputs)
+        exe_success, proc = exe_outputs
+
+        # Determine whether the calculation succeeded
+        output_data = {}
+        if not exe_success:
+            output_data["success"] = False
+            output_data["error"] = {"error_type": "internal_error",
+                                    "error_message": proc["stderr"]
+                                    }
+            return FailedOperation(
+                success=output_data.pop("success", False), error=output_data.pop("error"), input_data=output_data)
+
+        # If execution succeeded, collect results
+        result = self.parse_output(proc["outfiles"], input_data)
+        return result
 
     def build_input(self, input_model: 'ResultInput', config: 'JobConfig',
                     template: Optional[str] = None) -> Dict[str, Any]:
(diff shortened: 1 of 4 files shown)


qcengine/programs/molpro.py
```diff
@@ -5,8 +5,8 @@
 import xml.etree.ElementTree as ET
 from typing import Any, Dict, Optional
 
-# from qcelemental.models import ComputeError, FailedOperation, Provenance, Result
-from qcelemental.models import Result
+from qcelemental.models import Result, FailedOperation
+import qcengine.util as uti
```
dgasmith: `from ..util import execute`

qcengine/programs/tests/test_molpro.py
```diff
@@ -27,20 +28,38 @@ def test_molpro_output_parser(test_case):
...
+    inp = qcel.models.ResultInput.parse_raw(data["input.json"])
+
+    # Run Molpro
+    result = qcng.get_program('molpro').compute(inp, qcng.get_config(local_options={"ncores": 4}))
+    assert result.success is True
+
+    # Get output file data
+    output_ref = qcel.models.Result.parse_raw(data["output.json"])
+
+    if result.success:
```
dgasmith: The assert above will not let the computation get to this point.

qcengine/programs/tests/test_molpro.py
```diff
@@ -27,20 +28,38 @@ def test_molpro_output_parser(test_case):
...
+
+@using_molpro
+@pytest.mark.parametrize('test_case', molpro_info.list_test_cases())
+def test_molpro_executor(test_case):
+    # Get input file data
+    data = molpro_info.get_test_data(test_case)
+    inp = qcel.models.ResultInput.parse_raw(data["input.json"])
+
+    # Run Molpro
+    result = qcng.get_program('molpro').compute(inp, qcng.get_config(local_options={"ncores": 4}))
```
dgasmith: You can now call this via `qcng.compute(inp, 'molpro', local_options=config)`, no need to dive into the data.

qcengine/programs/tests/test_molpro.py
sjrl: The test for Terachem also has this redundancy. Should I change it there too?

dgasmith: Yes please!

qcengine/programs/tests/test_molpro.py
sjrl: I've updated the command. In this case I'm specifying the number of cores because it isn't necessarily advantageous for the Molpro testjobs to run with all available cores. In some cases it can prevent them from passing if they are run on a machine with lots of cores and a small amount of memory.

sjrl: Also as a side note, it looks like the `local_options` in `qcng.compute(inp, 'molpro', local_options=config)` expects a dictionary of type `Dict[str, str]`, but at least for the `ncores` option a `Dict[str, int]` needs to be specified. Should this be updated to `Dict[str, any]` or something similar or am I missing something?

qcengine/programs/tests/test_molpro.py
sjrl: Do you mind if I sneak the update to Terachem in this PR? Or should I make a separate PR?

loriab: Fine to add it here, imo.

qcengine/programs/tests/test_molpro.py
```diff
@@ -27,20 +28,36 @@ def test_molpro_output_parser(test_case):
     output_ref = qcel.models.Result.parse_raw(data["output.json"]).dict()
     output_ref.pop("provenance", None)
 
-    # TODO add `skip` to compare_recusive
+    # TODO add `skip` to compare_recursive
```
loriab: `compare_recursive` takes a `forgive` list where it won't fail even if missing or mismatch.

qcengine/programs/tests/test_molpro.py
dgasmith: Whoops, yes should be `Dict[str, Any]`. Note that you will need to import Any from typing.

qcengine/programs/tests/test_molpro.py
sjrl: I have added the changes

qcengine/programs/tests/test_molpro.py
sjrl: I don't think I actually put the `skip` comment there and I'm unsure of what we are worried about not matching.

PR #73, opened 2019-05-22
half a Cfour interface

loriab:
* Like GAMESS, this is the execution and parsing sides of the Cfour interface, not the input generation (that's next).
* 35 qcdb cfour test cases pass with it.
* there's still lots o'references to non `qcel.models` `Molecule`s

### DIFF
--- a/qcengine/programs/util/hessparse.py
+++ b/qcengine/programs/util/hessparse.py
@@ -1,0 +1,47 @@
+import numpy as np
+
+from qcelemental.util import filter_comments
+
+def load_hessian(shess, dtype):
+
+    # list o'lines w/o comments or blanks
+    shess = filter_comments(shess)
+    lhess = list(filter(None, map(str.strip, shess.splitlines())))
+
+    if dtype in ['fcmfinal', 'cfour']:
+        nat = int(lhess[0].split()[0])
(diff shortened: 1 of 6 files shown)


loriab:
Might not be the best notebook to launch. The psi4 binder goes to psiapi tutorial (which isn't really quick-start style either). Maybe launch to one that goes into some libmints fns on a SCF?

qcengine/programs/util/hessparse.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
```
dgasmith: Can we just have a single util file? I always found the multiple files under util to be troublesome unless the utils becomes truly huge.

loriab: Huh, I find the mixed-bag files like `qcengine/util.py` to be troublesome. A dictionary subclass and a Hessian reader just don't sound like they belong together. I expect some general options fns next. Can we revisit this when Cfour and Gamess interfaces are ready to register? There's ppl who'll be working on this, and it'd be nice to have files just migrate rather than migrate and rechristen.

dgasmith: Sounds good.

PR #76, opened 2019-05-24
(WIP) get_program: Added check option and fixed Molpro test jobs

sjrl:
Purposefully only fixed the Molpro test jobs to see if the Terachem ones fail in CI.

### DIFF
--- a/qcengine/programs/tests/test_molpro.py
+++ b/qcengine/programs/tests/test_molpro.py
@@ -22,7 +22,7 @@ def test_molpro_output_parser(test_case):
     data = molpro_info.get_test_data(test_case)
     inp = qcel.models.ResultInput.parse_raw(data["input.json"])
 
-    output = qcng.get_program('molpro').parse_output(data, inp).dict()
+    output = qcng.get_program('molpro', check=False).parse_output(data, inp).dict()
     output.pop("provenance", None)
 
     output_ref = qcel.models.Result.parse_raw(data["output.json"]).dict()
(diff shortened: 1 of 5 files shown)


qcengine/programs/tests/test_molpro.py
```diff
@@ -22,7 +22,7 @@ def test_molpro_output_parser(test_case):
     data = molpro_info.get_test_data(test_case)
```
sjrl: It looks like qcer doesn't populate the list `test_case` (line 18) for Molpro and Terachem on the CI build. It does run for me locally.

sjrl: It looks like qcer doesn't populate the list `test_case` (line 18) for Molpro and Terachem on the CI build. It does run for me locally.

dgasmith: I think I fixed this in qcer, I restarted the tests to see.

dgasmith:
Looks like it is picking up on it now, but failing due to the issue that you mentioned that CCSD gradients in Molpro do not export the CCSD total energy to the XML files for some reason.

sjrl:
Yes that is correct. I've been meaning to make a fix through that by either doing text parsing or grabbing the final energy from a different place within the xml. Should I do that for this PR?

dgasmith:
If you can accomplish your goal via the XML, I say do it. Otherwise I would be a bit hesitant to add text parsing to the clean parser. In this case, you could potentially edit it out the tests records tests.

sjrl:
Okay I've updated the output parser to grab the final energy from under the molecule tree if it can't be found in the normal spot under the jobstep tree. For the CI to pass a key had to be removed from qcer which I've done in this PR (https://github.com/MolSSI/QCEngineRecords/pull/12).

sjrl:
I've updated the parser so we no longer need qcer PR (https://github.com/MolSSI/QCEngineRecords/pull/12)

PR #78, opened 2019-05-28
Molpro: Implementing get_version for Molpro engine

sjrl:
The `get_version()` function now works. Currently I'm enforcing that any version of Molpro below 2018.1 won't work with QCEngine (the version of Molpro I've been testing these output parsers on). I haven't tested that extensively. Should I comment out that restriction for now?

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, Optional
 
 from qcelemental.models import Result, FailedOperation
 from ..util import execute
-from qcelemental.util import which
+from qcelemental.util import which, safe_version, parse_version
 
 from .executor import ProgramExecutor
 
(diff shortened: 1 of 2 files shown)


sjrl:
Thanks, and I have turned off CI testing for <2018.1 as you suggested. (https://github.com/MolSSI/QCEngine/pull/78/files#diff-231eed50a5284fd44cce744d7984054dR58)

qcengine/programs/molpro.py
```diff
@@ -35,17 +35,23 @@ def found(self, raise_error: bool=False) -> bool:
...
+                                      outfiles=["version.out", "version.xml"])
+
+            if success:
+                tree = ET.ElementTree(ET.fromstring(output["outfiles"]["version.xml"]))
+                root = tree.getroot()
+                version_tree = root.find('molpro_uri:job/molpro_uri:platform/molpro_uri:version', name_space)
+                year = version_tree.attrib['major']
+                minor = version_tree.attrib['minor']
+                molpro_version = year + "." + minor
+                self.version_cache[which_prog] = safe_version(molpro_version)
```
dgasmith: We need an else here that raises an exception. Usually this is a `TypeError`.

sjrl: Sounds good! I'm a little confused about how this is treated as a `TypeError`. Could you give me an example of how the error message would read?

dgasmith: Away from a computer at the moment. Can you look at the psi4 example?

sjrl: Taking a look at the psi4 example the `TypeError` is for catching custom versions of the Psi4 program? 

```
candidate_version = self.version_cache[which_prog]
if "undef" in candidate_version:
    raise TypeError("Using custom build without tags. Please pull git tags with `git pull origin master --tags`.")
```

I don't know if this exactly relevant for Molpro since I believe no matter what you compile it will always have a version enabled. Also I don't see any other examples for any of the other backends so I'm not entirely sure what I should be catching here.

dgasmith: Right, so this would be similar here:
```python
if version < (2018, 1):
    raise TypeError("QCEngine requires Molpro version 2018.1 or great")
```

loriab: Yeah, I'd say merge as is and only raise error if there's a problem with version computing (as with psi and the tags). Version too low belongs to [`is_program_new_enough`](https://github.com/MolSSI/QCEngine/blob/master/qcengine/testing.py#L35) imo.

PR #81, opened 2019-06-01
util.execute improvements

loriab:
- [x] write/read binary files
- [x] run in preexisting directory

### DIFF
--- a/qcengine/util.py
+++ b/qcengine/util.py
@@ -16,7 +16,7 @@ import time
 import traceback
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from pydantic import ValidationError
 from qcelemental.models import FailedOperation
@@ -36,7 +36,8 @@ def model_wrapper(input_data: Dict[str, Any], model: 'BaseModel') -> 'BaseModel'
         try:
             input_data = model(**input_data)
         except ValidationError as exc:
-            raise InputError(f"Error creating '{model.__name__}', data could not be correctly parsed:\n{str(exc)}") from None
+            raise InputError(
+                f"Error creating '{model.__name__}', data could not be correctly parsed:\n{str(exc)}") from None
     elif isinstance(input_data, model):
         input_data = input_data.copy()
     else:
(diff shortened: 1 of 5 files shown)


qcengine/util.py
```diff
@@ -412,7 +433,7 @@ def scratch_directory(child: str = None, parent: str = None, messy: bool = False
 
 
 @contextmanager
-def disk_files(infiles: Dict[str, str], outfiles: Dict[str, None], cwd: Optional[str] = None) -> Dict[str, str]:
+def disk_files(infiles: Dict[str, Union[str, bytes]], outfiles: Dict[str, None], cwd: Optional[str] = None, as_binary: Optional[List[str]] = None) -> Dict[str, Union[str, bytes]]:
```
dgasmith: Can we  do `disk_files(infiles, outfiles, *, ...)`. I a bit worried about mixing up the order after looking at the signature in execute.

qcengine/util.py
```diff
@@ -436,28 +459,32 @@ def disk_files(infiles: Dict[str, str], outfiles: Dict[str, None], cwd: Optional
         lwd = Path.cwd()
     else:
         lwd = Path(cwd)
+    if as_binary is None:
+        as_binary = []
```
dgasmith: Should we do any kinds of checks to make sure the binary values exist in the outfiles?

loriab: "binary values exist"? any file can be opened as binary, so the usual file exists check should be sufficient, unless I'm misunderstanding.

qcengine/util.py
loriab: sure. I actually added them, then deleted as over-involved for fn not called directly. will re-add.

qcengine/util.py
dgasmith: More like: `outfiles = {"hello"}, as_binary={"nonexistent"}`

Could prevent some confusion if we do something like:
 `assert len(set(as_binary) - set(outfiles)) == 0`

loriab: `as_binary` list is going to overlap with both `infiles` and `outfiles` dictionaries, so the existing OS errors should still catch. you're worried about `as_binary` never getting iterated over, so a mismatch could happen? I suspect this is a programmer's option, not an end-point option, so errors will get caught pretty quick. otherwise, I guess could pop as consumed and raise an error at the end if any remain.

loriab: so asserted.

PR #82, opened 2019-06-03
Entos: Added entos engine

sjrl:
I've added entos as an engine. It includes all functions to be run by `qcng.compute()` and the ability to run `entos.build_input()`, `entos.execute()`, and `entos.parse_output()`. I've begun adding support for using Jinja templates so we can cover the different ways might want to provide custom template files. 

(A) ordinary build_input (need to define a base template)
(B) user wants to add stuff after normal template (A)
(C) user knows their domain language (doesn't use any QCSchema quantities)

Right now it only supports option (C). Also right now the output parser does text parsing, but this will be removed eventually since entos plans to output all of their results to a xml or json format.

### DIFF
--- a/qcengine/programs/entos.py
+++ b/qcengine/programs/entos.py
@@ -9,6 +9,7 @@ from ..util import execute, popen
 from qcelemental.util import which, safe_version, parse_version
 
 from .model import ProgramHarness
+import string
 
 
 class EntosHarness(ProgramHarness):
@@ -57,13 +58,16 @@ class EntosHarness(ProgramHarness):
         job_inputs = self.build_input(input_data, config)
 
         # Run entos
-        exe_success, proc = execute(job_inputs["commands"],
-                                    infiles=job_inputs["infiles"],
-                                    outfiles=["dispatch.out"],
-                                    scratch_location=job_inputs["scratch_location"],
-                                    timeout=None
-                                    )
-        proc["outfiles"]["dispatch.out"] = proc["stdout"]
+        exe_success, proc = self.execute(job_inputs)
+
+        # # Run entos
+        # exe_success, proc = execute(job_inputs["commands"],
+        #                             infiles=job_inputs["infiles"],
+        #                             outfiles=["dispatch.out"],
+        #                             scratch_location=job_inputs["scratch_location"],
+        #                             timeout=None
+        #                             )
+        # proc["outfiles"]["dispatch.out"] = proc["stdout"]
 
         # Determine whether the calculation succeeded
         output_data = {}
@@ -72,6 +76,7 @@ class EntosHarness(ProgramHarness):
             output_data["error"] = {"error_type": "internal_error",
                                     "error_message": proc["stderr"]
                                     }
+            print(output_data["error"])
             return FailedOperation(
                 success=output_data.pop("success", False), error=output_data.pop("error"), input_data=output_data)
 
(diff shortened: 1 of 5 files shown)


sjrl:
Hmm probably should have realized `jinja2` isn't part of the required dependencies for QCEngine. Does it make sense to add the dependency or should we wait on that for now?

loriab:
`jinja2` is pretty light (dep list below), so I think it's fine to have it as an optional dependency. Can model it on `networkx` in qcel. https://github.com/MolSSI/QCElemental/search?q=networkx&unscoped_q=networkx

```
  ca-certificates    pkgs/main/linux-64::ca-certificates-2019.1.23-0
  certifi            pkgs/main/linux-64::certifi-2019.3.9-py37_0
  jinja2             pkgs/main/linux-64::jinja2-2.10.1-py37_0
  libedit            pkgs/main/linux-64::libedit-3.1.20181209-hc058e9b_0
  libffi             pkgs/main/linux-64::libffi-3.2.1-hd88cf55_4
  libgcc-ng          pkgs/main/linux-64::libgcc-ng-8.2.0-hdf63c60_1
  libstdcxx-ng       pkgs/main/linux-64::libstdcxx-ng-8.2.0-hdf63c60_1
  markupsafe         pkgs/main/linux-64::markupsafe-1.1.1-py37h7b6447c_0
  ncurses            pkgs/main/linux-64::ncurses-6.1-he6710b0_1
  openssl            pkgs/main/linux-64::openssl-1.1.1c-h7b6447c_1
  pip                pkgs/main/linux-64::pip-19.1.1-py37_0
  python             pkgs/main/linux-64::python-3.7.3-h0371630_0
  readline           pkgs/main/linux-64::readline-7.0-h7b6447c_5
  setuptools         pkgs/main/linux-64::setuptools-41.0.1-py37_0
  sqlite             pkgs/main/linux-64::sqlite-3.28.0-h7b6447c_0
  tk                 pkgs/main/linux-64::tk-8.6.8-hbc83047_0
  wheel              pkgs/main/linux-64::wheel-0.33.4-py37_0
  xz                 pkgs/main/linux-64::xz-5.2.4-h14c3975_4
  zlib               pkgs/main/linux-64::zlib-1.2.11-h7b6447c_3
```

sjrl:
Okay so I've begun adding `jinja2` as an optional dependency. If you could check my last commit (https://github.com/MolSSI/QCEngine/pull/82/commits/e472591bd89cc618722fab8ad9848c42373d7255) to see if the logic I'm using makes sense I'd really appreciate it. Also it looks like should add something to the files `setup.cfg`, `setup.py`, and one of the `yaml` files in `devtools/conda-envs/`, but I'm not exactly sure how to follow the `networkx` example in this case.

loriab:
You'll want to add jinja to whichever one of the devools/conda-envs/ you'd like to "test" the templating for. The setup* files can wait, as they don't actually get run for dependencies in the usual workflow.

sjrl:
I'm not sure which one of the devtools/conda-envs to add to? Should I just add it to the `psi4.yaml`?

Lnaden:
You could make a new one and add a new test in the `.travis.yml`

sjrl:
I'd be happy to do that too. I know we potentially plan to expand the use of `jinja2` to some of the other engines. Is it worth creating a new conda-env for that? Otherwise it seems like it would be pretty easy to sneak into the `psi4.yaml`.

qcengine/programs/entos.py
```diff
@@ -0,0 +1,238 @@
...
+            # # Build dictionary for substitute
+            # sub_dict = {
+            #     "method": input_model.model.method,
+            #     "basis": input_model.model.basis,
+            #     "df_basis": input_model.keywords["df_basis"].upper(),
+            #     "charge": input_model.molecule.molecular_charge
+            # }
+
+            # Perform substitution to create input file
+            str_template = jinja2.Template(template)
```
dgasmith: Something I was curious about is if we truly need jinja here, it seems our template are simple enough that they can be handled via more canonical means like finding `{{molecule}}` fields. I think my original suggestion may be been slightly optimistic towards the range of capabilities we could code in.

qcengine/programs/entos.py
```diff
@@ -0,0 +1,238 @@
...
+    def found(self, raise_error: bool = False) -> bool:
+        return which('entos', return_bool=True, raise_error=raise_error, raise_msg='Please install via XXX')
+
+    def get_version(self) -> str:
+        self.found(raise_error=True)
+
+        which_prog = which('entos')
+        if which_prog not in self.version_cache:
+            with popen([which_prog, '--version']) as exc:
+                exc["proc"].wait(timeout=5)
```
dgasmith: We should bump timeout to ~15 seconds. There are some very slow supercomputers out there.

qcengine/programs/entos.py
```diff
@@ -0,0 +1,238 @@
...
+                                    outfiles=outfiles,
+                                    scratch_location=inputs["scratch_location"],
+                                    scratch_name=scratch_name,
+                                    timeout=timeout
+                                    )
+        proc["outfiles"]["dispatch.out"] = proc["stdout"]
+
+        # Determine whether the calculation succeeded
+        output_data = {}
+        if not exe_success:
```
dgasmith: You can use the new `raise UnknownError` commands rather than forming a full `FailedOperation` see Psi4/rdkit.

qcengine/programs/entos.py
```diff
@@ -0,0 +1,238 @@
...
+                                'df_basis': input_model.keywords["df_basis"].upper(),
+                                'charge': input_model.molecule.molecular_charge
+                                },
+                              'print': {
+                                'results': True
+                                }
+                              }
+
+            # Write gradient call if asked for
+            if input_model.driver == 'gradient':
```
dgasmith: `elif ..` here with an `else` exception at the end.

It looks like very little changes between inputs as well. Possible to reduce duplication?

qcengine/programs/entos.py
```diff
@@ -0,0 +1,238 @@
...
+        natom = len(input_model.molecule.symbols)
+        for idx, line in enumerate(output_lines):
+            fields = line.split()
+            if fields[:1] == ["energy:"]:
+                properties["scf_total_energy"] = float(fields[-1])
+            elif fields[:2] == ["Molecular", "Dipole:"]:
+                properties["scf_dipole_moment"] = [float(x) for x in fields[2:5]]
+            elif fields[:3] == ["SCF", "converged", "in"]:
+                properties["scf_iterations"] = int(fields[3])
+            elif fields == ["Gradient", "(hartree/bohr):"]:
```
dgasmith: Since Entos is new, I don't suppose we could convince them to drop JSON output or even CSV?

qcengine/programs/entos.py
sjrl: I like the idea of having a template and then being able to do a dictionary substitute. I know the `string.Template` module offers some support of this, but it is very limited. I think we can provide the range of capabilities we discussed with jinja2, but it'll take some tinkering to be sure.

dgasmith: Absolutely, jinja will totally cover our use case. But the question is do we need it? `string.Template` provides fairly reaching functionality.

Or even writing our own using `{{method}}` with double bracket replacement.

qcengine/programs/entos.py
sjrl: They are in the process of writing their output in XML and/or JSON. It isn't currently done, but once it is I'll update the output parser.

qcengine/programs/entos.py
sjrl: I suppose what I didn't like about `string.Template` is that I couldn't find a way for it to tell me what fields are present which would then need to be filled. I'd like that functionality so we can be flexible in what information is required from the user in the `input.json`. It's possible that we could write our own but is it more trouble than it's worth? Is this mostly so we wouldn't need to have QCEngine be dependent on `jinja2` for something that is probably fairly simple?

qcengine/programs/entos.py
sjrl: Yeah the structure between the two are slightly different, but I was able to reduce code duplication.

qcengine/programs/entos.py
dgasmith: Pretty much, we try to keep the dependancies as small as possible for our lower level libraries and code small extra bits where needed.

sjrl: Okay so what is the way forward on this? @dgasmith I can switch over to using `string.Template` for now to keep the functionality I've added and to remove the jinja dependency, but I don't think this PR should include writing our own version of templating.

dgasmith: I would vote to delay the jinja integration unless absolutely necessary. Lets work through this in the slack channel to figure out what we need.

qcengine/programs/entos.py
```diff
@@ -0,0 +1,220 @@
...
+        proc = self.execute(job_inputs)
+
+        if isinstance(proc, FailedOperation):
+            return proc
+        else:
+            # If execution succeeded, collect results
+            result = self.parse_output(proc["outfiles"], input_data)
+            return result
+
+    def execute(self, inputs, extra_outfiles=None, extra_commands=None, scratch_name=None, timeout=None):
```
dgasmith: Can you add typing info here from the base harness?

sjrl: There is no typing info currently for the execute function from the base harness.

dgasmith:
Is this ready to go or are there other changes desired?

sjrl:
This should be all good to go!

dgasmith:
In we go then!

PR #85, opened 2019-06-06
MP2D gradients

loriab:
* new mp2d seems to produce correct gradients, so allow running and enable tests
* refactor the mp2d interface so it looks more like a Harness and doesn't call bits of dftd3
* ~this isn't going to work until I build new mp2d conda packages~

- [x] Q: Below is partial directory structure. What if I moved dashparam.py up a level (ok to live with the programs or should it be shoved into a resources/ or util/ folder?) then made dftd3 a single file again?
```
programs/dftd3/dashparam.py  # called by dftd3 & mp2d
programs/dftd3/runner.py
programs/mp2d.py
```
current status
```
programs/empirical_dispersion_resources.py  # called by dftd3 & mp2d & psi4
programs/dftd3.py
programs/mp2d.py
```


dgasmith:
Could we call it `emp_disp_dashcommands.py` or something?

loriab:
> Could we call it emp_disp_dashcommands.py or something?

`dashparams.py`? I'd call it `empirical_dispersion.py` if I had to rename it. `_resources`, `_utils`, `_parameters` if it needed a purpose modifier. If it gets renamed (which it probably should), I just want to do it only once, as it's been a resource locatable under this name for many years.

dgasmith:
Sure, just some thought. No real preferences there, I do like the underscore to denote that it is a very internal file.

loriab:
This is working nicely for mp2d gradients in psi4, so ready for review. the mp2d.py file has been rearranged so much that the diff isn't really useful. Let's get that one in good order, then I'll make dftd3 match it.

PR #88, opened 2019-06-07
MolproHarness Update

sjrl:
- Added execute function
- Added ability to provide a template to build_input using stdlib functionality
- Updated input generator to handle symmetry and orientation keywords

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, Optional
 from qcelemental.models import Result
 from ..util import execute
 from qcelemental.util import which, safe_version, parse_version
-from ..exceptions import UnknownError
+from ..exceptions import UnknownError, InputError
 
 from .model import ProgramHarness
 import string
@@ -81,6 +81,9 @@ class MolproHarness(ProgramHarness):
 
     def execute(self, inputs, extra_infiles=None, extra_outfiles=None, extra_commands=None, scratch_name=None,
                 scratch_messy=False, timeout=None):
+        """
+        For option documentation go look at qcengine/util.execute
+        """
 
         infiles = inputs["infiles"]
         if extra_infiles is not None:
@@ -97,7 +100,7 @@ class MolproHarness(ProgramHarness):
         exe_success, proc = execute(commands,
                                     infiles=infiles,
                                     outfiles=outfiles,
-                                    scratch_location=inputs["scratch_location"],
+                                    scratch_directory=inputs["scratch_directory"],
                                     scratch_name=scratch_name,
                                     scratch_messy=scratch_messy,
                                     timeout=timeout
@@ -150,7 +153,7 @@ class MolproHarness(ProgramHarness):
                 input_file.append('')
                 input_file.append('{force}')
             else:
-                raise NotImplementedError('Driver {} not implemented for Molpro.'.format(input_model.driver))
+                raise InputError('Driver {} not implemented for Molpro.'.format(input_model.driver))
 
             input_file = "\n".join(input_file)
         else:
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -79,11 +79,12 @@ class MolproHarness(ProgramHarness):
             result = self.parse_output(proc["outfiles"], input_data)
             return result
 
-    def execute(self, inputs, extra_infiles=None, extra_outfiles=None, extra_commands=None, scratch_name=None, timeout=None):
+    def execute(self, inputs, extra_infiles=None, extra_outfiles=None, extra_commands=None, scratch_name=None,
+                scratch_messy=False, timeout=None):
 
         infiles = inputs["infiles"]
         if extra_infiles is not None:
-            infiles.extend(extra_infiles)
+            infiles.update(extra_infiles)
 
         outfiles = ["dispatch.out", "dispatch.xml", "dispatch.wfu"]
         if extra_outfiles is not None:
@@ -98,6 +99,7 @@ class MolproHarness(ProgramHarness):
                                     outfiles=outfiles,
                                     scratch_location=inputs["scratch_location"],
                                     scratch_name=scratch_name,
+                                    scratch_messy=scratch_messy,
                                     timeout=timeout
                                     )
 
@@ -121,21 +123,9 @@ class MolproHarness(ProgramHarness):
             #     input_file.append('file,2,{}'.format(input_model.keywords.get('wfu')))
             input_file.append('')
 
-            # Don't orient the molecule if asked to fix_com or fix_orientation
-            if input_model.molecule.fix_orientation or input_model.molecule.fix_com:
-                input_file.append('{orient,noorient}')
-
-            # Have no symmetry if asked to fix_symmetry
-            if input_model.molecule.fix_symmetry == 'c1':
-                input_file.append('{symmetry,nosym}')
-            else:
-                input_file.append('{symmetry,auto}')
-
-            input_file.append('')
-
             # Write the geom
-            xyz_file = input_model.molecule.to_string(dtype='molpro', units='Bohr')
-            input_file.append(xyz_file)
+            xyz_block = input_model.molecule.to_string(dtype='molpro', units='Bohr')
+            input_file.append(xyz_block)
 
             # Write the basis set
             input_file.append('basis={')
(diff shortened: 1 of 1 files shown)


sjrl:
I think this will still fail the CI since it depends on the master version of QCElemental.

qcengine/programs/molpro.py
```diff
@@ -68,82 +70,115 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+
+            # Don't orient the molecule if asked to fix_com or fix_orientation
+            if input_model.molecule.fix_orientation or input_model.molecule.fix_com:
+                input_file.append('{orient,noorient}')
+
+            # Have no symmetry if asked to fix_symmetry
+            if input_model.molecule.fix_symmetry == 'c1':
+                input_file.append('{symmetry,nosym}')
+            else:
+                input_file.append('{symmetry,auto}')
```
loriab: As I see it, there are 3 things one could reasonably expect to get back from `to_string(dtype=...)`:
* (a) Cartesian coordinate lines only, with perhaps the Natom line for XYZ
* (b) All the molecule info that can be got out of the schema and the dtype program will accept _that can be handed to the dtype program **contiguous** to the Cartesian lines_
* (c) All the molecule info ... will accept with the non-contiguous or options-like parts returned in a dictionary.

I started off with (a). Currently most dtypes are (b). I'm doing (c) for cfour. Glad to gather opinions. The symmetry and orient could be included with (b) or (c).

qcengine/programs/molpro.py
```diff
@@ -68,82 +70,115 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+
+        commands = inputs["commands"]
+        if extra_commands is not None:
+            commands = extra_commands
+
+        exe_success, proc = execute(commands,
+                                    infiles=infiles,
+                                    outfiles=outfiles,
+                                    scratch_location=inputs["scratch_location"],
+                                    scratch_name=scratch_name,
```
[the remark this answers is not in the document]
sjrl: Sounds good. Does it make sense to add a doc string for `Harness.execute()`? Or at least point people to the doc string of `util.execute` so they know what the different options mean?

qcengine/programs/molpro.py
sjrl: It looks like I've mostly done option (b) for the `to_string` method for Molpro. I think it makes sense that the `to_string` function with specific `dtype`s for each program handles all of the molecule info that is relevant for that program. Unsure about option (b) vs (c) though.

sjrl: Should I go ahead and make another PR for QCEl or should I wait for further discussion on this?

qcengine/programs/molpro.py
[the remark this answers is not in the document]
loriab: The latter, I think. That, in turns, points ppl to the the docstrings of the context managers. :-) Otherwise, keeping this in sync will be a mess.

qcengine/programs/molpro.py
loriab: I think, yes, go ahead and do another qcel PR for orient and symmetry. will need to be careful that the meanings map directly. isn't there a `symmetry,cs`, etc. in molpro for fixing C_s symmetry for an otherwise C_2v mol?

qcengine/programs/molpro.py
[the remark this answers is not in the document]
sjrl: Makes a lot of sense to me! Is there a good way to accomplish this task? Or could I just write a docstring saying "For option documentation go look at qcengine/util.execute"?

qcengine/programs/molpro.py
sjrl: That could very well be the case, unfortunately I'm not too familiar with the symmetry functionality in Molpro.

sjrl: This is now handled in the `to_string` method of `molecule`. PR https://github.com/MolSSI/QCElemental/pull/72

dgasmith: Great, should have a new release tomorrow and will rerun the CI.

Once this goes through is there anything preventing a merge?

sjrl: I think everything should be good to go. Oh Lori had a question above about the memory assignment.

loriab:
Ok, this passes. I'll bump the qcel pinning in the other PR.

qcengine/programs/molpro.py
```diff
@@ -68,82 +70,115 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
-        input_file.append('{{{:s}}}'.format(input_model.model.method))
-
-        # Write gradient call if asked for
-        if input_model.driver == 'gradient':
+        if template is None:
+            input_file = []
+            posthf_methods = {'mp2', 'ccsd', 'ccsd(t)'}
+
+            # Memory is in megawords per core for Molpro
+            memory_mw_core = int(config.memory * (1024 ** 3) / 8e6 / config.ncores)
```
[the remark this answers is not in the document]
loriab: Daniel has confirmed 10% cushion has already been deducted.

qcengine/programs/molpro.py
```diff
@@ -68,82 +70,105 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+
+            # Write appropriate driver call
+            if input_model.driver == 'energy':
+                input_file.extend(energy_call)
+            elif input_model.driver == 'gradient':
+                input_file.extend(energy_call)
+                input_file.append('')
+                input_file.append('{force}')
+            else:
+                raise NotImplementedError('Driver {} not implemented for Molpro.'.format(input_model.driver))
```
loriab: `NotImplementedError` not defined. should be `InputError`, I think.

qcengine/programs/molpro.py
```diff
@@ -177,22 +200,36 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
...
         }
-        ccsd_dipole_map = {"Dipole moment": "ccsd_dipole_moment"}
-        ccsd_extras = {}
+
+        # CCSD(T) maps
+        # TODO There are two instances of "correlation energy" in the xml. One for CCSD and one for CCSD(T)
+        #      Will need a bit more logic to separate between the two
+        ccsd_prt_pr_energy_map = {
+            "total energy": "ccsd_prt_pr_total_energy",
+            "correlation energy": "ccsd_prt_pr_correlation_energy"
```
loriab: be good to collect just the `(T)` energy, too, or instead. (T) = CCSD(T) – CCSD. It's the immediately computed value, so most accurate.

qcengine/programs/molpro.py
```diff
@@ -177,22 +200,36 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
...
+        }
+        ccsd_prt_pr_dipole_map = {"Dipole moment": "ccsd_prt_pr_dipole_moment"}
```
loriab: are any dipoles already in atomic units?

qcengine/programs/molpro.py
```diff
@@ -68,82 +70,105 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+            outfiles.extend(extra_outfiles)
+
+        commands = inputs["commands"]
+        if extra_commands is not None:
+            commands = extra_commands
+
+        exe_success, proc = execute(commands,
+                                    infiles=infiles,
+                                    outfiles=outfiles,
+                                    scratch_location=inputs["scratch_location"],
```
loriab: change scratch_location --> scratch_directory if #89 already merged

qcengine/programs/molpro.py
sjrl: I'm not collecting these values at the moment because of complications with the Molpro XML using the same name for CCSD and CCSD(T) quantities. There is a way to differentiate, but I haven't spent the time to elegantly fix the problem. I'll add the `(T)` to the dictionary so I don't forget.

sjrl: Also there isn't a keyword for just the `(T)` value in qcel. Is that something to add in the future?

qcengine/programs/molpro.py
sjrl: The dipole is already in atomic units.

qcengine/programs/molpro.py
sjrl: I believe `NotImplementedError` comes with the stdlib in python? Would you prefer that I used `InputError` which is defined in `qcengine/exceptions`?

dgasmith:
Can you rebase this? Look like we hit a conflict.

qcengine/programs/molpro.py
dgasmith: I think we probably want to add a (T) term to the schema and elemental.

qcengine/programs/molpro.py
dgasmith: Right, we want to use all `qcengine/exceptions` where possible. This allows categorization and restart capabilities downstream along with some extra features.

qcengine/programs/molpro.py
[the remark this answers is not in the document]
dgasmith: I think a doctoring is sufficient for now :)

qcengine/programs/molpro.py
sjrl: Should I raise an issue on Schema and/or Elemental for the time being?

sjrl:
Okay rebased and should be good to go.

qcengine/programs/molpro.py
dgasmith: Lets do the canonical Schema -> Elemental route.

PR #89, opened 2019-06-11
dftd3: make dftd3 harness-like

loriab:
ok use of `extras['local_keywords']`?

### DIFF
--- a/qcengine/programs/dftd3.py
+++ b/qcengine/programs/dftd3.py
@@ -93,7 +93,7 @@ class DFTD3Harness(ProgramHarness):
             inputs["outfiles"],
             scratch_messy=False,
             #env=inputs["env"],
-            scratch_location=inputs["scratch_location"],
+            scratch_directory=inputs["scratch_directory"],
             blocking_files=inputs["blocking_files"],
         )
         return success, dexe
@@ -148,7 +148,7 @@ class DFTD3Harness(ProgramHarness):
             "command": command,
             "infiles": infiles,
             "outfiles": ['dftd3_gradient', 'dftd3_abc_gradient'],
-            "scratch_location": config.scratch_directory,
+            "scratch_directory": config.scratch_directory,
             "input_result": input_model.copy(deep=True),
             "blocking_files": [os.path.join(pathlib.Path.home(), '.dftd3par.' + socket.gethostname())],
         }
--- a/qcengine/programs/dftd3.py
+++ b/qcengine/programs/dftd3.py
@@ -1,27 +1,23 @@
 """Compute dispersion correction using Grimme's DFTD3 executable."""
 
-import copy
-import json
 import os
 import pathlib
 import pprint
 import re
 import socket
 import sys
-import traceback
 from decimal import Decimal
-from typing import Dict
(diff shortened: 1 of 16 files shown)


qcengine/programs/dftd3.py
```diff
@@ -62,326 +61,224 @@ def get_version(self) -> str:
...
-    modulerec['env'] = {
-            'HOME': os.environ.get('HOME'),
-            'PATH': os.pathsep.join([os.path.abspath(x) for x in os.environ.get('PSIPATH', '').split(os.pathsep) if x != '']) + \
-                    os.pathsep + os.environ.get('PATH'),
-            'LD_LIBRARY_PATH': os.environ.get('LD_LIBRARY_PATH'),
+        return {
+            "command": command,
+            "infiles": infiles,
+            "outfiles": ['dftd3_gradient', 'dftd3_abc_gradient'],
+            "scratch_location": config.scratch_directory,
```
dgasmith: We should standardize location/directory one way or the other across the whole code base. I vote a bit for `scratch_directory` as this keyword is found in the large QCA ecosystem.

qcengine/programs/dftd3.py
```diff
@@ -62,326 +61,224 @@ def get_version(self) -> str:
...
+
+        retres = calcinfo[f'CURRENT {input_model.driver.upper()}']
+        if isinstance(retres, Decimal):
+            retres = float(retres)
+        elif isinstance(retres, np.ndarray):
+            retres = retres.ravel().tolist()
+
+        output_data = {
+            'extras': {
+                'local_keywords': input_model.extras['info'],
```
dgasmith: No restrictions on extras, the present info seems reasonable however.

qcengine/programs/dftd3.py
loriab: If it meant the place directly where the calc happens, I would agree. But `scratch_location` turns into `parent` https://github.com/MolSSI/QCEngine/blob/master/qcengine/util.py#L418 or "scratch directory directory." Is that what the generic QCA `scratch_directory` means?

qcengine/programs/dftd3.py
loriab: `local_keywords` is sort of "resolved keywords" and is the best record of what the calc did. I don't really care about the name of the key. Not following "No restrictions on extras."

loriab: nm, I had forgotten I asked the question at the top, so was trying to figure out the point of your comment.

qcengine/programs/dftd3.py
dgasmith: I mean from a QCA perspective it doesn't matter. It is the location to spawn tasks, if a further tmpdir is built it is quite irrelevant as the tmpdirs are always cleaned. There could be additional complications that I am not considering however.

loriab: I guess I use “scratch directory” two ways, too. The enclosing loc like /scratch/hohndoe/ or the qc run loc like /tmp/water.4444/. Along that like, I don’t really care directory or location. Can change.

loriab:
ready for review

PR #98, opened 2019-06-28
nwc: tce general cases fix

vivacebelles:
Adding back capability of NWChem's harvester to pull TCE options that have been able to in QCDB.

### DIFF
--- a/qcengine/programs/nwchem/harvester.py
+++ b/qcengine/programs/nwchem/harvester.py
@@ -217,19 +217,19 @@ def harvest_outfile_pass(outtext):
                 psivar[f'{cc_corr} CORRECTION ENERGY'] = mobj.group(1)
                 psivar[f'{cc_plain} CORRELATION ENERGY'] = mobj.group(2)
                 psivar[f'{cc_plain} TOTAL ENERGY'] = mobj.group(3)
-        #Other TCE
-        for cc_name in [r'CCSD', r'CI', r'LCC']:
+        #Process other TCE cases
+        for cc_name in [r'CISD', r'CISDT', r'CISDTQ', r'CCD', r'CCSD', r'CCSDT', r'CCSDTQ', r'LCCSD', r'LCCD']:
             mobj = re.findall(
                 r'^\s+' + r'Iterations converged' + r'\s*' +
-                r'^\s+' + r'(.*?)' + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'
-                r'^\s+' + r'(.*?)' + r' total energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*$', 
-                outtext, re.MULTILINE | re. DOTALL)
+                r'^\s+' + cc_name + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'+
+                r'^\s+' + cc_name + r' total energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*$', 
+                outtext, re.MULTILINE)
 
             for mobj_list in mobj:
-                print('matched %s' % mobj_list[0])
-                print(mobj_list)
-                psivar['%s CORRELATION ENERGY' % mobj_list[0]] = mobj_list[1]
-                psivar['%s TOTAL ENERGY' % mobj_list[2]] = mobj_list[3]
+                print('matched %s'% cc_name)
+                print(mobj)
+                psivar[f'%s CORRELATION ENERGY' % cc_name] = mobj_list[0]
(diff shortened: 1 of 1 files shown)


qcengine/programs/nwchem/harvester.py
```diff
@@ -218,17 +218,18 @@ def harvest_outfile_pass(outtext):
...
         #Other TCE
-        #mobj = re.findall(
-        #        r'^\s+' + r'Iterations converged' + r'\s*' +
-         #       r'^\s+' + r'(.*?)' + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'
-         #       r'^\s+' + r'(.*?)' + r' total energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*$', 
-          #      outtext, re.MULTILINE | re. DOTALL)
+        for cc_name in [r'CCSD', r'CI', r'LCC']:
+            mobj = re.findall(
+                r'^\s+' + r'Iterations converged' + r'\s*' +
+                r'^\s+' + r'(.*?)' + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'
```
loriab: I think the findall --> search, `(.*?)` --> `cc_name` patterns like mbpt and (t) above would be better here.

vivacebelles:
Spoke too soon! Need to adjust, so doesn't interfere with special TCE test cases.

qcengine/programs/nwchem/harvester.py
```diff
@@ -217,18 +217,19 @@ def harvest_outfile_pass(outtext):
...
-        #mobj = re.findall(
-        #        r'^\s+' + r'Iterations converged' + r'\s*' +
-         #       r'^\s+' + r'(.*?)' + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'
-         #       r'^\s+' + r'(.*?)' + r' total energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*$', 
-          #      outtext, re.MULTILINE | re. DOTALL)
+        #Process other TCE cases
+        for cc_name in [r'CISD', r'CISDT', r'CISDTQ', r'CCD', r'CCSD', r'CCSDT', r'CCSDTQ', r'LCCSD', r'LCCD']:
+            mobj = re.findall(
+                r'^\s+' + r'Iterations converged' + r'\s*' +
+                r'^\s+' + cc_name + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'+
```
loriab: [this](https://github.com/MolSSI/QCEngine/pull/98/files#diff-c0d24fbdb771ad09450a93a96fe06099L209) preferred because it doesn't hardcode the spaces before `correlation energy /` the way the above does.

qcengine/programs/nwchem/harvester.py
```diff
@@ -217,18 +217,19 @@ def harvest_outfile_pass(outtext):
...
 
-        #for mobj_list in mobj:
-         #   print('matched %s' % mobj_list[0])
-         #   print(mobj_list)
-         #   psivar['%s CORRELATION ENERGY' % mobj_list[0]] = mobj_list[1]
-         #   psivar['%s TOTAL ENERGY' % mobj_list[2]] = mobj_list[3]
+            for mobj_list in mobj:
+                print('matched %s'% cc_name)
+                print(mobj)
+                psivar[f'%s CORRELATION ENERGY' % cc_name] = mobj_list[0]
```
loriab: can use f-strings like the blocks above. and just `if mobj` when back to `re.search`

qcengine/programs/nwchem/harvester.py
```diff
@@ -217,18 +217,19 @@ def harvest_outfile_pass(outtext):
...
                 psivar[f'{cc_plain} TOTAL ENERGY'] = mobj.group(3)
-        #Other TCE
-        #mobj = re.findall(
-        #        r'^\s+' + r'Iterations converged' + r'\s*' +
-         #       r'^\s+' + r'(.*?)' + r' correlation energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*'
-         #       r'^\s+' + r'(.*?)' + r' total energy / hartree' + r'\s+=\s*' + NUMBER + r'\s*$', 
-          #      outtext, re.MULTILINE | re. DOTALL)
+        #Process other TCE cases
+        for cc_name in [r'CISD', r'CISDT', r'CISDTQ', r'CCD', r'CCSD', r'CCSDT', r'CCSDTQ', r'LCCSD', r'LCCD']:
+            mobj = re.findall(
```
loriab: The `for cc_name in [` loop is taking the place of the `re.findall`, so this should be ok as `re.search`

PR #101, opened 2019-07-11
TeraChem: Enable user's request of extra output files.

ffangliu:
Enabled extra files in TeraChem scratch folder to be requested by users, collected after program execution, and recorded in the Result object as extras.

### DIFF
--- a/qcengine/programs/terachem.py
+++ b/qcengine/programs/terachem.py
@@ -64,7 +64,7 @@ class TeraChemHarness(ProgramHarness):
         # Setup the job
         job_inputs = self.build_input(input_data, config)
         # Run terachem
-        exe_outputs = self.execute(job_inputs)
+        exe_outputs = self.execute(job_inputs,extra_outfiles=input_data.extras)
         exe_success, proc = exe_outputs
         # Determine whether the calculation succeeded
         output_data = {}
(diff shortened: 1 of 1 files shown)


qcengine/programs/terachem.py
```diff
@@ -189,13 +189,21 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
...
+        for extra in input_model.extras.keys():
+            input_model.extras[extra] = outfiles[extra]
+
         return Result(**{**input_model.dict(), **output_data})
 
     def execute(self, inputs, extra_outfiles=None, extra_commands=None, scratch_name=None, timeout=None):
-
+        binaries = []
+        for filename in extra_outfiles:
+            if filename in ['scr/ca0', 'scr/cb0','scr/c0']:
```
dgasmith: Should this be `src/*`? We could do something like check if `filename.startswith("src") or filename.startswith("/src") `

ffangliu: The scr folder has both binary files like (ca0, cb0, c0, the binary MO coefficients), and many non-binary files like molden, trajectory. I found that it seems like I need to tell the executor which ones are binaries to make sure it can be correctly collected. Otherwise, I got encoding errors.

dgasmith: Right, just curious if we needed to make additional files as binary.

ffangliu: Yes, I think these are the ones that are generated in all terachem calculations. Some special functionalities may also generate binary files, and I need to figure that our case by case.

dgasmith: Ok great! Just checking in.

Looks like your tests are hit by a Pydantic bug that we fixed, but isn't in the current release. I am going to override CI to merge this.

ffangliu: Great! Thanks!

PR #103, opened 2019-07-16
Automated retry capabilities for random errors

dgasmith:
Allows random errors to be retried (current default is zero retries). Geometry optimizations automatically set retries to 2.


qcengine/compute.py
```diff
@@ -72,20 +73,26 @@ def compute(input_data: Union[Dict[str, Any], 'ResultInput'],
...
         local_options = {**local_options, **input_engine_options}
         config = get_config(local_options=local_options)
 
-        output_data = executor.compute(input_data, config)
+        for x in range(retries + 1):
+            try:
+                output_data = executor.compute(input_data, config)
+                break
+            except RandomError:
+                metadata["reties"] += 1
```
loriab: reties typo

qcengine/compute.py
```diff
@@ -28,34 +28,35 @@ def _process_failure_and_return(model, return_dict, raise_error):
 
 def compute(input_data: Union[Dict[str, Any], 'ResultInput'],
             program: str,
-            raise_error: bool=False,
-            local_options: Optional[Dict[str, Any]]=None,
-            return_dict: bool=False) -> 'Result':
+            raise_error: bool = False,
```
loriab: suggest a `*,` here so only input_data and program can be positional

qcengine/compute.py
```diff
@@ -28,34 +28,35 @@ def _process_failure_and_return(model, return_dict, raise_error):
...
+    program : str
+        The program to execute the input with.
+    raise_error : bool, optional
         Determines if compute should raise an error or not.
-    capture_output:
-        Determines if stdout/stderr should be captured.
-    local_options:
+    retries : int, optional
+        The number of random tries to retry for.
+        Description
```
loriab: "Description"?

dgasmith:
Ok, this needed a serious overhaul to really make it work. Can @loriab and @Lnaden both review this? Fairly intricate tech and needs a few eyes on it.

qcengine/procedures/model.py
```diff
@@ -15,6 +15,9 @@ class Config:
         allow_mutation: False
         extra: "forbid"
 
+    def __init__(self, **kwargs):
+        super().__init__(**{**self._defaults, **kwargs})
```
Lnaden: This causes an infinite recursion when I test this type of construct locally, and where is `_defaults` defined?

qcengine/compute.py
```diff
@@ -76,16 +76,29 @@ def compute(input_data: Union[Dict[str, Any], 'ResultInput'],
...
+                output_data = executor.compute(input_data, config)
+                break
+            except RandomError as e:
+
+                if x == config.retries:
+                    raise e
+                else:
+                    metadata["retries"] += 1
+            except:
+                raise
```
loriab: is the last except/raise just to be explicit about flow control?

qcengine/programs/base.py
```diff
@@ -34,6 +34,16 @@ def register_program(entry_point: 'ProgramHarness') -> None:
...
 
 
+def unregister_program(name: str) -> None:
+    """
+    Unregisters a given program.
+    """
+
+    ret = programs.pop(name.lower(), None)
+    if ret is None:
+        raise KeyError(f"Program {name} is not registered with QCEngine")
```
Lnaden: Is a raise really required here? If a user wants to unregister and its not there, is there any harm? I could see leaving this to be explicit about whats present and not.

qcengine/compute.py
dgasmith: This will raise a non-Random Exception. For example there is not a good reason to run an InputError multiple times.

qcengine/programs/base.py
Lnaden: After discussion, I think leaving this is fine so it catches things like spelling errors and symmetric with the `register` fuction.

PR #105, opened 2019-07-17
Testing Azure Integration

Lnaden:
As on the tin, this tests Azure auto-building and badge status


Lnaden:
This may be a work in progress for a bit, so bear with me if all the checks don't trigger correctly and I have to make a number of pushes/commits.

PR #110, opened 2019-07-20
Psi4: Expands random error expectations

dgasmith:


jchodera:
I can confirm this fix drastically reduces failures on `lilac`.

PR #112, opened 2019-07-24
Molpro update

sjrl:
The MolproHarness has been updated to handle DFT and CCSD(T) energies and gradients. The output parser went through a large change to accommodate reading in CCSD and CCSD(T) quantities at the same time. And now it should be easier to expand the output parser to support more methods in the future. 

On my local machine I've checked that this passes with the PR https://github.com/MolSSI/QCEngineRecords/pull/15 for QCEngineRecords, although it might be good to run this through Travis as well.

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -27,6 +27,7 @@ class MolproHarness(ProgramHarness):
 
     # Set of implemented dft functionals in Molpro according to
     # https://www.molpro.net/info/release/doc/manual/node208.html (version 2019.2)
+    # TODO This isn't extensive because it definitely doesn't include component functionals (e.g. LYP, VWN5)
     _dft_functionals = {
         "B", "B-LYP", "BLYP", "B-P", "BP86", "B-VWN", "B3LYP", "B3LYP3", "B3LYP5", "B88X",
         "B97", "B97R", "BECKE", "BH-LYP", "CS", "D", "HFB", "HFS", "LDA", "LSDAC", "LSDC",
@@ -35,8 +36,10 @@ class MolproHarness(ProgramHarness):
         "M08-SO", "M11-L", "TPSS", "TPSSH", "M12HFC", "HJSWPBE", "HJSWPBEH", "TCSWPBE", "PBESOL"
     }
 
-    # Currently supported post-hf methods in QCEngine for Molpro
+    # Currently supported methods in QCEngine for Molpro
+    _scf_methods = {"HF", "RHF", "KS", "RKS"}
     _post_hf_methods = {'MP2', 'CCSD', 'CCSD(T)'}
+    _supported_methods = {*_scf_methods, *_post_hf_methods}
 
     class Config(ProgramHarness.Config):
         pass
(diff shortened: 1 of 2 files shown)


qcengine/programs/molpro.py
```diff
@@ -293,32 +311,39 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
...
-        if method in post_hf_methods and method_energy_map['total energy'] in properties:
-            final_energy = properties[method_energy_map['total energy']]
-        elif method in scf_methods and method_energy_map['Energy'] in properties:
-            final_energy = properties[method_energy_map['Energy']]
+        if method in self._post_hf_methods and molpro_map['total energy'][method] in properties:
+            final_energy = properties[molpro_map['total energy'][method]]
+        elif method in self._scf_methods and molpro_map['Energy'][method] in properties:
+            final_energy = properties[molpro_map['Energy'][method]]
         else:
+            # TODO This won't work for DFT case since mol_method will be name of DFT functional
```
dgasmith: Which cases will not work here and what is the data that will be missing?

qcengine/programs/molpro.py
```diff
@@ -201,49 +224,43 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
...
+            },
+            "Dipole moment": {
+                "HF": "scf_dipole_moment",
+                "RHF": "scf_dipole_moment",
+                "KS": "scf_dipole_moment",
+                "RKS": "scf_dipole_moment",
+                "MP2": "mp2_dipole_moment",
+                "CCSD": "ccsd_dipole_moment",
+                "CCSD(T)": "ccsd_prt_pr_dipole_moment"
+            }
```
dgasmith: +1 for making this a literal.

I probably asked this before, but are the SCF energy components available in the MOLPRO XML?

qcengine/programs/molpro.py
sjrl: So this is meant to serve as a backup way of grabbing the energy if for some reason it’s not provided in the xml normally. This happens in the case of calling a CCSD or a CCSD(T) gradient. So this shouldn’t be needed for a DFT calculation but right now it wouldn’t work since mol_method is the actual functional name and method at this point is RKS.

sjrl: I just made a fix for this through some logic rearrangement. Will push now.

qcengine/programs/molpro.py
sjrl: Actually yes they are I just noticed. At least for the DFT case the XC functional is available, but in a different part of the xml from where I grab the total energy. I can see if I can grab these components as well.

sjrl: So I was able to grab `nuclear_repulsion_energy` and the `scf_xc_energy`, but that's all I can get for the scf energy breakdown.

sjrl:
Since I've added the nuclear repulsion energy and the scf xc energy this will cause Travis to break because QCEngineRecords will need to be updated.

qcengine/programs/molpro.py
```diff
@@ -25,6 +25,34 @@ class MolproHarness(ProgramHarness):
...
+        "TREF2", "TREF", "TTEST", "GLE", "GREEN", "SRB88", "SRLYP", "LB94", "EI", "SAOP", "USER",
+        "INE", "ECERF2", "ECERFINTER", "ECERFLOCAL2", "ECERFLOCAL", "EXERFLOCAL", "FC", "FCFO",
+        "FCO", "FL", "XC-M11", "PBEXANAL", "PBECANAL", "PBESOLCANAL", "PBESOLXANAL", "EXSRLDA",
+        "EXSRLPBE", "ECSRLPBE", "ECSRLLPBE", "ECSQRTLPBE", "ECMUDIVLPBE", "EXERFPHS", "ECLERFMUPBE",
+        "ECERFERFCPBE", "ECSQRTLDA", "REVPBEX", "B", "B-LYP", "BLYP", "B-P", "BP86", "B-VWN", "B3LYP",
+        "B3LYP3", "B3LYP5", "B88X", "B97", "B97R", "BECKE", "BH-LYP", "CS", "D", "HFB", "HFS", "LDA",
+        "LSDAC", "LSDC", "KYP88", "MM05", "MM05-2X", "MM06", "MM06-2X", "MM06-L", "MM06-HF", "PBE", "PBE0",
+        "PBE0MOL", "PBEREV", "PW91", "S", "S-VWN", "SLATER", "VS99", "VWN", "VWN80", 'M05',
+        "M05-2X", "M06", "M06-2X", "M06-L", "M06-HF", "M08-HX","M08-SO", "M11-L", "TPSS",
+        "TPSSH", "M12HFC", "HJSWPBE", "HJSWPBEH", "TCSWPBE", "PBESOL"
```
loriab: I agree with you on dealing only with XC functionals, not X or C alone. Sure that "B" is an XC?

sjrl: Yes it looks like B is an alias for B88 which I confirmed by running an input using `{rks,B}`. 

Also there isn't a clean list of what's a full XC functional versus components (sorry for my lack of DFT knowledge on this one). If there is an easy way to prune this that would be great, but going back to the previous shorter list isn't a great option because some of the letters like `B` are aliases for the more canonical name `B88` which wasn't included in the previous list.

qcengine/programs/molpro.py
```diff
@@ -141,14 +175,16 @@ def build_input(self, input_model: 'ResultInput', config: 'JobConfig',
...
-            write_hf = input_model.model.method.lower() in posthf_methods
-            if write_hf:
+            # If post-hf method is called then make sure to write a HF call first
+            if input_model.model.method.upper() in self._post_hf_methods:
                 energy_call.append('{HF}')
-            # TODO Support DFT calls, need to check if method is a DFT XC functional and then write {rks,XC}
-            energy_call.append('{{{:s}}}'.format(input_model.model.method))
+            # If DFT call make sure to write {rks,method}
+            if input_model.model.method.upper() in self._dft_functionals:
+                energy_call.append('{{rks,{:s}}}'.format(input_model.model.method))
```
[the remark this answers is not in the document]
sjrl: Oh right! Totally blanked on f-strings, I'm happy to switch over.

qcengine/programs/molpro.py
dgasmith: Thanks, always worth pulling anything else that we can!

dgasmith:
@sjrl I think this is ready, can we merge?

sjrl:
Yup merge away!

PR #114, opened 2019-07-25
0.8.2 Docs and bugfix for required kwargs in Fractal's stack

Lnaden:
Something in what I think is a serialization step in one of our Fractal
Queue Adapters causes the function signature to lose its
`kwargonlydefaults` field which makes the function look like a required
kwarg which has no defaults and so cant be called correctly.


Lnaden:
Failures here are related to the updated Molpro and will be resolved by #112. Should not be blocking for this fix.

PR #116, opened 2019-08-05
Compute: Environments

dgasmith:
Adds a environment context manager to catch NumPy style parallelization with Python functions.

### DIFF
--- a/qcengine/util.py
+++ b/qcengine/util.py
@@ -266,6 +266,44 @@ def popen(args: List[str], append_prefix: bool = False,
             ret["stderr"] = error.decode()
 
 
+@contextmanager
+def environ_context(config: Optional['JobConfig'] = None, env: Optional[Dict[str, str]] = None) -> Dict[str, str]:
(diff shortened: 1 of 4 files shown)


qcengine/util.py
```diff
@@ -266,6 +266,44 @@ def popen(args: List[str], append_prefix: bool = False,
             ret["stderr"] = error.decode()
 
 
+@contextmanager
+def environ_context(config: Optional['JobConfig'] = None, env: Optional[Dict[str, str]] = None) -> Dict[str, str]:
```
loriab: interaction of this and [the `env` in `execute()`](https://github.com/MolSSI/QCEngine/pull/116/files#diff-138f99563678a91c2c82995c8537354cR320) not really explored yet, right?

dgasmith: Right, I don't think we can dip to that level yet. The idea here was only to tackle the `np.dot`-like calls with this manager.

I bet we could consider tying the two together eventually, but subprocess becomes very program-dependent.

PR #120, opened 2019-08-07
MSGPack updates

dgasmith:
Updates QCEngine to support the new Msgpack/NumPy-based QCElemental. This change requires https://github.com/leeping/geomeTRIC/pull/96 to pass.


dgasmith:
This will pass as 0.6.0 propagates on c-f. Please go ahead and review so we can merge quickly.

qcengine/procedures/geometric.py
```diff
@@ -27,6 +27,9 @@ def compute(self, input_data: 'OptimizationInput', config: 'JobConfig') -> 'Opti
 
         geometric_input = input_data.dict()
 
+        # Temporary patch for geomeTRIC
```
mattwelborn: It looks like geometric PR #96 is merged. Is this code still necessary?

qcengine/procedures/geometric.py
```diff
@@ -27,6 +27,9 @@ def compute(self, input_data: 'OptimizationInput', config: 'JobConfig') -> 'Opti
...
+        geometric_input["initial_molecule"]["symbols"] = list(geometric_input["initial_molecule"]["symbols"])
+
         # Older QCElemental compat, can be removed in v0.6
```
mattwelborn: Since this PR depends on QCElemental 0.6, this may be a good time to remove this compat code.

devtools/conda-envs/psi-nightly.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - pyyaml
   - py-cpuinfo
   - psutil
-  - qcelemental >=0.4.2
+  - qcelemental >=0.6.0
```
loriab: want to add msgpack to this lane and leave plain psi to json to get some testing variety?

qcengine/procedures/geometric.py
dgasmith: This compatibility code is required for backwards compatibility with `geometric`.

qcengine/procedures/geometric.py
dgasmith: This compatibility code is required for backwards compatibility with geometric, unless we pin a higher geometric version which isn't really required at the moment.

Issue #118, opened 2019-08-07
Document program detection


mattwelborn:
The documentation section "Environment Detection" should provide an example of how to see which programs were detected. I think that qcengine.list_available_programs() does this? (Same goes for procedures.)

loriab:
a `qcengine --info` command might be a good idea. print
* where qcengine is running from
* version of qcengine and qcel
* all interfaced programs (for this qcng version)
* all detected programs along with their version and location
* ditto procedures
* default `Config`

dgasmith:
Agreed, that some CLI data would be useful. Perhaps more inline with the other QCA CLI's:

- `QCEngine info` - Show information as listed above.
- `QCEngine run *program* *json_file/json_data*` - Executes the program from input
- `QCEngine run-procedure *procedure* *json*` - Executes the procedure.

mattwelborn:
@dgasmith 0.9.0 or no?

dgasmith:
I say no, but hopefully before the release. Does the proposed CLI seem reasonable?

mattwelborn:
I'm happy with the proposed CLI.

PR #132, opened 2019-08-14
Expanded CLI

mattwelborn:
This PR extends the CLI to encompass three commands:

* `qcengine info` displays information about the environment detected by QCEngine.
* `qcengine run` runs a program.
* `qcengine run-procedure` runs a procedure.

This PR closes issue #118.

# TODO
- [x] Determine what to test, and how to test it. Do we need to add a CLI test suite as in QCFractal?


qcengine/cli.py
```diff
@@ -4,29 +4,119 @@
...
 
 
 def parse_args():
     parser = argparse.ArgumentParser(description='A CLI for the QCEngine.')
+    subparsers = parser.add_subparsers(dest="command")
 
-    parser.add_argument("program", type=str, default=None, help="The program to call")
-    parser.add_argument("data", type=str, default=None, help="A JSON blob of the task description.")
+    info = subparsers.add_parser('info', help="Print information about QCEngine setup, version, and environment.")
+    info.add_argument('--version', action='store_true', help="Print version of qcengine and qcelemental.")
```
dgasmith: QCEngine/QCElemental case similar to above?

qcengine/cli.py
```diff
@@ -4,29 +4,119 @@
...
 __all__ = ["main"]
 
 
 def parse_args():
     parser = argparse.ArgumentParser(description='A CLI for the QCEngine.')
+    subparsers = parser.add_subparsers(dest="command")
 
-    parser.add_argument("program", type=str, default=None, help="The program to call")
-    parser.add_argument("data", type=str, default=None, help="A JSON blob of the task description.")
+    info = subparsers.add_parser('info', help="Print information about QCEngine setup, version, and environment.")
```
dgasmith: In QCFractal we use `show`. Any votes for `show` vs `info`? We should be consistent.

qcengine/cli.py
```diff
@@ -4,29 +4,119 @@
...
+
+def info_cli(args):
+
+    def info_version():
+        print(">>> Version information:")
+        print(f"QCEngine version:    {__version__}")
+        try:
+            import qcelemental
+            print(f"QCElemental version: {qcelemental.__version__}")
+        except ImportError as e:
```
dgasmith: This should always pass since QCElemental is a dependancy.

qcengine/cli.py
```diff
@@ -4,29 +4,119 @@
...
-    print(json.dumps(ret))
+    command = args.pop("command")
+    if command == "info":
+        info_cli(args)
+    elif command == "run":
+        ret = compute(json.loads(args["data"] if args["data"] != "-" else sys.stdin.read()), args["program"])
+        print(ret.json())
+    elif command == "run-procedure":
+        ret = compute_procedure(json.loads(args["data"] if args["data"] != "-" else sys.stdin.read()), args["procedure"])
+        print(ret.json())
```
dgasmith: Should we allow for something like `qcengine run --infile=input.msgpack --outfile=output.msgpack` with support for full extensions? Or is that a bit much for this pass?

mattwelborn: My target audience for the CLI was demoing/debugging and human interaction. For production input -- MsgPack input can't be generated by hand -- I think the API is the appropriate interface. Let me know if you agree.

qcengine/cli.py
mattwelborn: I think I prefer changing the `qcfractal-server user` CLI to `info` because `qcfractal-server user info` sounds more (English) idiomatic than `qcfractal-server user show`.

mattwelborn:
Thanks for the rapid comments!

qcengine/cli.py
dgasmith: I was thinking for the `qcfractal-server show ...` commands that either show the current fractal config or alembic info.

This could become `qcfractal-server show programs`. `info` is likely better running through a few examples.

qcengine/cli.py
dgasmith: This could also allow `.yaml` files which would be even easier to write by hand?

`is os.path.isfile(args["data"]); result.parse_file(args["data"]); else ...`

Would allow the user to do what they please.

qcengine/cli.py
mattwelborn: `info` was the decision.

qcengine/cli.py
mattwelborn: Do we support yaml parsing? 

https://github.com/MolSSI/QCElemental/blob/6ae258fd9ae326b19e9e12cf0dca43d26c6c282d/qcelemental/models/basemodels.py#L56

dgasmith: We do not, but using the `parse_file` could be nice as it would be trivial to add there and then it would be picked up everywhere.

mattwelborn: Added.

qcengine/cli.py
```diff
@@ -4,29 +4,147 @@
...
+    data_arg: str
+        Either a data blob or file name or '-' for STDIN
+
+    Returns
+    -------
+    ResultInput
+        An input for compute.
+    """
+    if data_arg == "-":
+        return ResultInput.parse_raw(sys.stdin.read())
```
dgasmith: This will work for now, but we will need to become more sophisticated as I believe `sys.stdin.read()` will always write out a str which will be interpreted as JSON.

qcengine/cli.py
```diff
@@ -4,29 +4,147 @@
...
-    parser.add_argument("data", type=str, default=None, help="A JSON blob of the task description.")
+    subparsers = parser.add_subparsers(dest="command")
 
-    return vars(parser.parse_args())
+    info = subparsers.add_parser('info', help="Print information about QCEngine setup, version, and environment.")
+    info.add_argument('--version', action='store_true', help="Print version of QCEngine and QCElemental.")
+    info.add_argument('--programs', action='store_true', help="Print detected and supported programs.")
+    info.add_argument('--procedures', action='store_true', help="Print detected and supported procedures.")
+    info.add_argument('--config', action='store_true', help="Print host, compute, and job configuration.")
+    info.add_argument('--all', action='store_true', help="Print all available information.")
```
dgasmith: To make this simpler I like the idea of a:
`QCEngine info config`

Can be implemented as:

```
info.add_argument("category", nargs="?", default="all", choices=["version", "programs", ...], help="The config category to show.")
```

qcengine/cli.py
mattwelborn: I agree. I think this is something of an upstream issue in QCElemental, since it implements the logic of detecting MsgPack vs JSON vs YAML maybe in the future. Currently, I think it uses the extension (datatype) for files (raw data). 

Given some STDIN, I don't know how to know if it's bytes or str. It might make sense to detect the input type in a more sophisticated manner. Does MsgPack have a header? How far do we have to read into a JSON vs YAML file to figure out what kind it is?

We could also just do something like:
```
data = None
for parser in [yaml_parser, msgpack_parser, json_parser]:
    try:
        data = parser(blob)
        break
    except ParserError:
        pass
else:
    throw ValueError("Cannot parse data")
```

dgasmith: What about `--input-encoding=None` where `None` would try everything? Might be nice to have a `--output-encoding` as well.

We might need to port over some CLI testing tech from Fractal.

dgasmith: This is getting large enough that I would vote to merge the current and make an issue to prevent this PR from getting too big.

mattwelborn: Okay, I will do the `info config` suggestion and then we can merge this.

qcengine/cli.py
mattwelborn: DOne.

qcengine/cli.py
mattwelborn: Created issues #135 and #136 for this.

PR #134, opened 2019-08-14
Doc update for 0.9.0

Lnaden:


docs/source/changelog.rst
```diff
@@ -13,6 +13,37 @@ Changelog
...
+++++++++++++
+
+- (:pr:`112`) The ``MolproHarness`` has been updated to handle DFT and CCSD(T) energies and gradients.
+- (:pr:`116`) An environment context manager has been added to catch NumPy style parallelization with Python functions.
+- (:pr:`117`) MOPAC and DFTD3 can now accept an ``extras`` field which can pass around additional
+  data, conforming to the rest of the Harnesses.
+- (:pr:`119`) Small visual improvements to the docs have been made.
+- (:pr:`120`) Lists inside models are now generally converted to numpy arrays for internal storage to maximize the
+  benefit of the new Msgpack feature from Elemental.
+- (:pr:`128`) NWChem Harvester's ``harvest_hessian`` function has been implemented.
```
loriab: I don't think what happened in 128 is worth mentioning. I don't understand it, but it wasn't an implementation. https://github.com/MolSSI/QCEngine/blob/3e5840572b9386510684e5fa28371c8e9ade834d/qcengine/programs/nwchem/harvester.py#L571-L576

mattwelborn: Previously, the function did nothing in the sense that it contained one line of code, returned None and had no side effects. As an example of the PR workflow, I had a fellow mark it as not implemented. This statement should be removed from the change log.

Lnaden: It was a bit hard to gauge what to do with all the fixes from the class we were teaching. Sure the fixes were all small, but they were things we wanted done, and in this case, the function does go from raising a `NotImplementedError` to doing something. I can remove the call out though if you think its not important enough.

I could also just bundle all the small fixes into one mention in the change long, kind of like what I did with Elemental.

Lnaden: Ah, okay. i'll remove it

loriab: either way. I feel no compunction about skipping minute changes. After all, most major PRs should be getting several bullets if we were thorough.

Lnaden: Thats a fair point. I've removed the offending line for this one and will a bit more lax in the future.

PR #137, opened 2019-08-15
Enable Azure CI with custom images for proprietary code

Lnaden:
This PR creates an Azure Pipeline which uses a custom Docker image we (MolSSI)
maintain and have exclusive access to through Azure Container Registry.

On this image we can house non-public code which will enable us to test
Engine through integrated CI on these codes securely.

I will say, there are lots of shenanigans and chain of about 20-25 different
commits of trying different things which got squished into this, but I can
confirm it works. Once this is merged in, I will activate it on the
Azure side for Engine itself and we can see it in action.

Right now, the image only contains a static MOPAC binary, but we could expand it
to include packages, or make other images and set up a strategy of different
containers to run through the pipeline. Thanks to Jonathan Moussa (@godotalgorithm)
for helping get the binary.

Here is what one of those runs looks like btw: https://dev.azure.com/MolSSI/QCArchive/_build/results?buildId=41

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,53 @@
+# Based on a Starter pipeline
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+# Optional to only trigger on Master branch changes
+# trigger:
+# - master
+
+# Define the continaers we want access to
+resources:
+  containers:
+    -   container: 'qcengineci' # Arbitrary name for reference in pipelines
+        image: qcarchive.azurecr.io/qcengineci:latest # pointer to the Image in our Azure CR
+        endpoint: "QCArchive Azure Container Registry" # Name of the Service Connection the pipeline is configured with
+
+variables:
+  conda_source: /opt/conda/etc/profile.d/conda.sh # Helper variable 
+
+jobs:
+- job: proprietary_ci
+  displayName: 'CI with Proprietary Code'
+  pool:
+    vmImage: 'ubuntu-latest'
+  # The ENTRYPOINT in these images are ignored, so we need to use the variable above to source out the conda environment each time.
+  # https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops&tabs=yaml#linux-based-containers
+  # And this is not really a Service Container either since the CMD/ENTRYPOINT does not provide a service to run along side everything else
+  # and because we want to actually run inside this(the) container(s).
+  container: qcengineci
+
+  steps:
+  - script: |
+      uname -a
+      free -m
+      df -h
+      ulimit -a
+    displayName: 'Diagnostics'
+    
+  - script: |
+      source $(conda_source)
+      conda env create -n test -f devtools/conda-envs/base.yaml
+    displayName: 'Create the conda environment'
+
+  - script: |
+      source $(conda_source)
+      conda activate test
+      python setup.py develop
+    displayName: 'Install QCEngine'
+
+  - script: |
+      source $(conda_source)
+      conda activate test
+      pytest -v .
+    displayName: 'Run the Tests'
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,53 @@
...
+  - script: |
+      source $(conda_source)
+      conda activate test
+      python setup.py develop
+    displayName: 'Install QCEngine'
+
+  - script: |
+      source $(conda_source)
+      conda activate test
+      pytest -v .
```
loriab: could add a `-rs` to get a summary of why stuff skipped

azure-pipelines.yml
```diff
@@ -0,0 +1,53 @@
...
+# Optional to only trigger on Master branch changes
+# trigger:
+# - master
+
+# Define the continaers we want access to
+resources:
+  containers:
+    -   container: 'qcengineci' # Arbitrary name for reference in pipelines
+        image: qcarchive.azurecr.io/qcengineci:latest # pointer to the Image in our Azure CR
+        endpoint: "QCArchive Azure Container Registry" # Name of the Service Connection the pipeline is configured with
```
dgasmith: Can you describe a bit about how the authentication is handled? Is this a private Azure pw?

Lnaden:
> this looks too simple for the power of what it does. do other things need to go in a README for documentation? 
> Can you describe a bit about how the authentication is handled? Is this a private Azure pw?

The YAML side of things is the simple part of this, I am working on a full write up for how everything else is handled as a markdown file. Stay tuned.

Lnaden:
I have renamed the YAML file to be a hidden file, and I have added in a full "How to set this all up from near-scratch" markdown file in the devtools folder:

https://github.com/MolSSI/QCEngine/blob/58088c34459cf96ea4dcd26b91f996897a86e306/devtools/azure_ci_images.md

If this addresses the concerns about security and how-tos, then I think we can merge this. If things are still unclear though, let me know and I will keep writting!

devtools/azure_ci_images.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Pre-requisites
+
+* Do everything in the previous section
+* Be made an "Owner" of ACR and an "Administrator" of the Pipelines team
+* Be signed into the account which meets these criteria.
+
+### Linking the services
+
+1. Sign into the account which meets the pre-requisites through the DevOps login.
```
loriab: `0.`

devtools/azure_ci_images.md
```diff
@@ -0,0 +1,231 @@
...
+
+* Docker
+* Access to the private software you want to include for CI testing
+* (For uploading) Be a "Contributor" (or otherwise Push access) to the ACR in question. See some of the "Authenticating 
+  a new users for ACR and Pipelines" for more details.
+* (For uploading) Install the [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest)
+  
+### Creating the Image
+
+1. Start a new [Dockerfile](https://docs.docker.com/engine/reference/builder/) with the following one lline header:
```
loriab: lline

devtools/azure_ci_images.md
```diff
@@ -0,0 +1,231 @@
...
+```yaml
+  steps:
+  - script: |
+      source /opt/conda/etc/profile.d/conda.sh
+      # The rest of the script...
+```
+
+### Running against multiple images
+
+One avenue for future iomprovements is having multiple images on the ACR we want to reference. For that, we can
```
loriab: improvements

devtools/azure_ci_images.md
Lnaden: Markdown doesn't care about numbered list ordering. I could have just put `1. 1. 1. 1111. 1. 11.` and it would have ordered them correctly in render. It does look a bit odd to the author though. I fixed it anyways.

Lnaden:
Thanks for the reviews. I'll merge this when it all passes and then point the CI here at engine.

PR #138, opened 2019-08-15
Update changelog to test the Azure triggers

Lnaden:
This is just a change log update, but it should trigger the Azure private CI builds. This is mainly me testing I have it all set up correctly now that the YAML file is in place

### DIFF
--- a/devtools/azure_ci_images.md
+++ b/devtools/azure_ci_images.md
@@ -36,6 +36,19 @@ ACR can only be done if the following conditions are met:
 Then, and only then, can a Pipeline build access images in the ACR. Further, no unauthorized user can see, manage, 
 or change these credentials.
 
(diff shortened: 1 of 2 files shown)


Lnaden:
*sigh* it can never be easy, can it?

Lnaden:
The problem here is triggering builds from Forks can introduce malicious code and trigger builds simply by opening a PR. The settled solution for now is to disable builds on PRs, and only allow builds on merged master changes, so code has to be approved and merged by Engine admins.

Lnaden:
Na, I had just starting working on other things while I waited for travis to finish.

Lnaden:
I did figure out the syntax so we can manually trigger builds. They wont show up on the PRs though.

PR #139, opened 2019-08-20
 Update the install docs

Lnaden:
Tested that pip and conda both work. Stripped out the from source
install.


docs/source/install.rst
```diff
@@ -1,54 +1,63 @@
...
+   >>> pip install qcengine
+
+
+Test the Installation
+---------------------
+
+.. note::
+
+   QCEngine is a wrapper for other quantum chemistry codes. The tests for QCEngine will only test the wrapper for a
+   given code if its detected in the PATH or current Python Environment, otherwise the tests for that package are
```
dgasmith: $PATH in with code wrappers?

PR #144, opened 2019-08-28
efp: start pylibefp interface

loriab:
- [ ] this one test passes but not ready for merging

### DIFF
--- a/qcengine/programs/tests/test_pylibefp.py
+++ b/qcengine/programs/tests/test_pylibefp.py
@@ -253,6 +253,7 @@ def test_total_1a(system_1, keywords_label, results_label):
     assert compare_recursive(expected_ene, res.extras['local_properties'], atol=atol)
 
 
+@using_pylibefp
 @pytest.mark.parametrize("keywords,errmsg", [
     ({'nonsense_key': 'harmless', 'elec_damp': 'nonsense'}, "invalid value for [screen/overlap/off] elec_damp: nonsense"),
     ({'elec': 'yEs'}, "invalid value for [T/F] elec: yEs"),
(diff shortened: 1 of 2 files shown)


qcengine/programs/pylibefp.py
```diff
@@ -0,0 +1,159 @@
...
+from .model import ProgramHarness
+
+pp = pprint.PrettyPrinter(width=120, compact=True, indent=1)
+
+
+class PylibEFPHarness(ProgramHarness):
+
+    _defaults = {
+        "name": "PylibEFP",
+        "scratch": False,
```
dgasmith: This is probably thread safe?

qcengine/programs/tests/test_pylibefp.py
```diff
@@ -0,0 +1,274 @@
+import copy
+import pytest
+
+from qcelemental.testing import compare, compare_recursive, compare_values
+
+import qcengine as qcng
+from qcengine.testing import using_pylibefp
+
+b2a = 0.529177
```
dgasmith: I assume efp uses its own conversion? Can we add a comment here about that.

loriab: I'll look it up, but this particular one is just to get the geometry in with both round numbers and to match the answers taken from libefp test suite. outputs are all in Eh, a0. So this conversion isn't a broad concern.

qcengine/programs/pylibefp.py
loriab: all in memory and all functions acting on same struct instance, so yes.

qcengine/programs/tests/test_pylibefp.py
dgasmith: Makes sense. Should be fine as long as we note it.

PR #145, opened 2019-08-28
CLI tests

mattwelborn:
This PR adds rudimentary tests for the CLI, testing:

- `qcengine info` and its options terminate successfully 
- `qcengine run` and `qcengine run-procedure` terminate successfully for string and filename input, with rudimentary output checks.

A future PR will needed to address piping STDIN for `run` and `run-procedure`, which will need to be done carefully to avoid breaking coverage.

## TODO

- [x] use process running functions from util.py
- [x] add coverage

### DIFF
--- a/qcengine/cli.py
+++ b/qcengine/cli.py
@@ -5,12 +5,13 @@ Provides a CLI for QCEngine
 import argparse
 import os.path
 import sys
-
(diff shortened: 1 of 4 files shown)


mattwelborn:
@dgasmith @loriab could I get another review of this please?

PR #150, opened 2019-09-06
Add Hessian driver to torchani.py

farhadrgh:
- This PR adds `hessian` driver to `torchani.py`

- I have also changed the minimum requirement for TorchANI version from 0.5 to 0.9. 

In [TorchANI 0.8 release](https://github.com/aiqm/torchani/releases/tag/0.8) we fixed the breaking changes recently introduce in PyTorch, and [TorchANI 0.9](https://github.com/aiqm/torchani/releases/tag/0.9) has all the large files removed. This reduces the size of TorchANI installed with pip to ~57MB. We made it optional for users to download the large files via a `download.sh` script.

### DIFF
--- a/qcengine/programs/torchani.py
+++ b/qcengine/programs/torchani.py
@@ -77,6 +77,7 @@ class TorchANIHarness(ProgramHarness):
             raise ResourceError("QCEngine's TorchANI wrapper requires version 0.9 or greater.")
 
         import torch
+        import torchani
         import numpy as np
 
         device = torch.device('cpu')
@@ -111,7 +112,6 @@ class TorchANIHarness(ProgramHarness):
             ret_data["return_result"] = np.asarray(derivative *
                                                    ureg.conversion_factor("angstrom", "bohr")).ravel().tolist()
         elif input_data.driver == "hessian":
-            import torchani
             hessian = torchani.utils.hessian(coordinates, energies=energy)
             ret_data["return_result"] = np.asarray(hessian)
         else:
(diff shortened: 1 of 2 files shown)


qcengine/programs/torchani.py
```diff
@@ -110,9 +110,13 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
             derivative = torch.autograd.grad(energy.sum(), coordinates)[0].squeeze()
             ret_data["return_result"] = np.asarray(derivative *
                                                    ureg.conversion_factor("angstrom", "bohr")).ravel().tolist()
+        elif input_data.driver == "hessian":
+            import torchani
```
dgasmith: Could you move this import up to the `import torch` section on line 80 so that this is all imported in one area?

farhadrgh: Sure, fixed in 9939fab

qcengine/programs/torchani.py
```diff
@@ -73,10 +73,11 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
 
         # Check if existings and version
         self.found(raise_error=True)
-        if parse_version(self.get_version()) < parse_version("0.5"):
-            raise ResourceError("QCEngine's TorchANI wrapper requires version 0.5 or greater.")
+        if parse_version(self.get_version()) < parse_version("0.9"):
+            raise ResourceError("QCEngine's TorchANI wrapper requires version 0.9 or greater.")
```
loriab: should the version check happen for testing, too? https://github.com/MolSSI/QCEngine/blob/master/qcengine/testing.py#L121-L126

farhadrgh: Yes, you are right, I missed that. Fixed it in 7d16e98

Issue #147, opened 2019-09-02
Create OpenMM engine harness


dotsdl:
We would like to be able to execute OpenMM workloads, if possible, using QCEngine. This should involve creating a `qcengine.programs.model.ProgramHarness` subclass. Use the existing `*Harness` implementations as inspiration.

This issue should define the scope for this implementation; it serves as the nexus for discussion on this addition.

dotsdl:
Currently working on this one.

dgasmith:
@j-wags. Is there a Smirnoff toolkit piece we could borrow here?

Pretty excited about this. Should have quite a few applications such as starting guesses and conformer.

jchodera:
Apologies for hijacking this thread, but we're now in a position where we need to implement OpenMM-based energy and gradient evaluation for small molecule force fields for QCEngine for use in benchmark comparisons of SMIRNOFF and other small molecule force fields for the [Open Force Field Initiative](http://openforcefield.org).

Our initial needs are to allow processing of small molecules through the same workflows we use for the Open Force Field Initiative, which are currently `TorsionDriveDataset` and `OptimizationDataset`.

Because we need a molecular topology to begin with, we would likely base this off of [`qcengine/programs/rdkit.py`](https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/rdkit.py), which creates an RDKit `Chem.Mol` object that can then be fed to downstream parameterization engines.

We would like to support energy and gradient computation using OpenMM, using the following force fields:
* Open Force Field SMIRNOFF-format force fields, assigned via the [Open Force Field Toolkit](https://github.com/openforcefield/openforcefield) using the RDKit/AmberTools backend
* GAFF force fields (1.* and 2.*) assigned via Antechamber via AmberTools

Here's what we're thinking about how dispatch would work:
* OpenMM execution would be accessed with `program : openmm` (should this be something else more specific to reflect small molecule parameterization will occur after RDKit processing?)
* The `model` dict would specify the force field as `method` (e.g. `gaff-1.8`, `gaff-2.0`, `smirnoff-1.1.0`) to be consistent with how the `rdkit` support specifies `method : uff`. 
*  `gaff-*` support would be provided via AmberTools and a distributed set of the appropriate GAFF versions excised from their corresponding AmberTools versions, but we would use the latest AmberTools. We can support a `chargemodel` keyword that specifies the model to pass to Antechamber, and possibly an optional `charges` array that overrides this.
[...]

PR #151, opened 2019-09-09
Review: OpenMM harness (addresses #147)

dotsdl:
Initial structure in place; working on input structure and laying out execution bits. Used the basic example from the OpenMM docs as the starting point, and existing `rdkit` and `psi4` harnesses for guidance.

Using @jchodera's comment as our target use case: https://github.com/MolSSI/QCEngine/issues/147#issuecomment-527691867

Any additional requirements for this harness should be listed in a comment on #147.


jchodera:
Wow, this is a great start! 

It looks like you're targeting a really different use case, way beyond a single molecule in vacuum.

In the short-term, we're really in need of something that simply focuses on evaluating the energies and gradients for a single molecule in vacuum---the equivalent of what is done with QM---for comparison/benchmarking of the new openforcefield force field against the QM reference data.

I highly recommend using the RDKit molecule processing from here to form an RDKit `Mol`:
https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/rdkit.py#L47-L80

After that, you can then shove it through the Open Force Field toolkit to get an OpenMM `System` directly from one of our force fields:
```python
# Create an OFF Molecule from the RDKit Molecule
from openforcefield.topology import Molecule
off_molecule = Molecule(rdkit_molecule)
# Load an OFF Force Field
from openforcefield.typing.engines.smirnoff import ForceField
off_forcefield = ForceField(offxml_filename)
# Assign parameters to the molecule in vacuum
openmm_system = off_forcefield.create_openmm_system(off_molecule.to_topology())
```
You can then perform an energy/gradient evaluation with OpenMM.

For computations with GAFF/GAFF2, it will be a bit more complicated, since we need to use Antechamber to generate charges and parameters, and then either convert those into an OpenMM XML force field residue template or convert them via `leap` into a `prmtop/inpcrd` to load into OpenMM. We have lots of code to wrap this already, and I'm happy to help there!

I'm working with Junmei Wang to get access to a historical repository of all GAFF versions so we can specify which version we want in particular.

(Feel free to move this comment to #147 if desired!)

dgasmith:
This looks like a great start! I echo what John is pointing out here that keeping something pretty simple and quantum-like is likely the simplest and most useful choice to work through here. `compute` is specifically for `energy/gradient/Hessian` computations.

Running short dynamics will be _super_ useful as well, but we would like want to pack this into a procedure.

dotsdl:
Changes made addressing comments from @jchodera, @dgasmith, and
@jaimergp:
* procesing molecule with rdkit -> openforcefield -> openmm
* limiting scope to `driver`s ('energy', 'gradient')
* `model` specifies forcefield as `method`; also has `offxml` and `url` fields

Next:
* need structure positions generated
* need gradient calculation (`state.getForces()`)
* need `rdkit` checking in `found` method
* need `url` handling for forcefield *somewhere* (either in `openforcefield` or here)

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+            simulation.context.setPositions(pdb.positions)
+
+            # Execute driver
+            try:
+                if input_data.driver == "energy":
+                    simulation.context.setPositions(pdb.positions)
+                    state = simulation.context.getState(getEnergy=True)
+                    ret_data["return_result"] = state.getPotentialEnergy()
+                elif input_data.driver == "gradient":
+                    # TODO: calculate forces with OpenMM
```
jchodera: This should be easy:
```python
# Get number of atoms
n_atoms = len(jmol.symbols)
# Set positions
context.setPositions(off_mol.conformers[0])
# Compute the forces
state = simulation.context.getState(getForces=True)
# Get the gradient as a simtk.unit.Quantity with shape (n_atoms, 3)
gradient = - state.getForces(asNumpy=True)
# Convert to hartree/bohr and reformat as 1D array
ret_data["return_result"] = (gradient / (unit.hartree/unit.bohr)).reshape([n_atoms * 3])
```

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+                    pass
+                else:
+                    raise InputError(f"OpenMM can only compute energy and gradient driver methods. Found {input_data.driver}.")
+            except:
+                raise
+            else:
+                success = True
+
+            # GET OUTPUT, PUT INTO FORM WE NEED
```
jchodera: What output is necessary here?

jchodera:
Hi! Just wanted to check in to see what the plans and timeline for moving forward with this might be.

qcengine/programs/openmm.py
dotsdl: I don't see `hartree` in the `simtk.unit` module. Did this used to be there? I also don't see it present among source shipped with the `openmm` 7.4.0 conda package, either.

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+            # Create OpenMM system in vacuum from forcefield, molecule
+            off_top = off_mol.to_topology()
+            openmm_top = off_top.to_openmm()
+            openmm_system = off_forcefield.create_openmm_system(off_top)
+            
+            # Need an integrator for simulation even if we don't end up using it really
+            integrator = LangevinIntegrator(300*units.kelvin, 1/units.picosecond, 0.002*units.picoseconds)
+
+            # Set platform to CPU explicitly
+            platform = openmm.Platform.getPlatformByName('CPU')
```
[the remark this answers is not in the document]
dotsdl: Can we also do something like the following to achieve this, with the ability to set the number of threads via the `JobConfig`, or do we strictly want to leave this to the parent environment variables of the process?

```
# get number of threads to use from `JobConfig; otherwise, try environment variable
nthreads = config.nthreads
if nthreads is None:
    nthreads = os.environ.get('OPENMM_CPU_THREADS')

# Set platform to CPU explicitly
platform = openmm.Platform.getPlatformByName('CPU')

# Initialize context
context = openmm.Context(openmm_system, integrator, platform)

# set number of threads
if nthreads:
    platform.setPropertyValue(context, 'Threads', nthreads)
```

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+            # likely an existing tool in the modules we're using since this is
+            # just a molecule floating in space with no solvent
+            simulation.context.setPositions(pdb.positions)
+
+            # Execute driver
+            try:
+                if input_data.driver == "energy":
+                    simulation.context.setPositions(pdb.positions)
+                    state = simulation.context.getState(getEnergy=True)
+                    ret_data["return_result"] = state.getPotentialEnergy()
```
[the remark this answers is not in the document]
dotsdl: I don't see `hartree` in the `simtk.unit` module. Did this used to be there? I also don't see it present among source shipped with the `openmm` 7.4.0 conda package, either.

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+
+        return mol
+
+    def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
+        """
+        Runs OpenMM on given structure, inputs, in vacuum.
+        """
+        self.found(raise_error=True)
+
+        from simtk.openmm.app import Simulation
```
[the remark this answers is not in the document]
dotsdl: Thanks @jchodera! Fixed!

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+            # Set platform to CPU explicitly
+            platform = openmm.Platform.getPlatformByName('CPU')
+
+            # Initialize simulation
+            simulation = Simulation(openmm_top, openmm_system, integrator, platform)
+
+            # TODO: need a way to generate positions from the topology
+            # likely an existing tool in the modules we're using since this is
+            # just a molecule floating in space with no solvent
+            simulation.context.setPositions(pdb.positions)
```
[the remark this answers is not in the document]
dotsdl: Excellent, added!

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+            openmm_system = off_forcefield.create_openmm_system(off_top)
+            
+            # Need an integrator for simulation even if we don't end up using it really
+            integrator = LangevinIntegrator(300*units.kelvin, 1/units.picosecond, 0.002*units.picoseconds)
+
+            # Set platform to CPU explicitly
+            platform = openmm.Platform.getPlatformByName('CPU')
+
+            # Initialize simulation
+            simulation = Simulation(openmm_top, openmm_system, integrator, platform)
```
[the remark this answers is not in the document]
dotsdl: Thanks! I've switched to using a `Context` object as you suggest.

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+            # Load an OFF Force Field
+            off_forcefield = ForceField(offxml)
+
+            # Create OpenMM system in vacuum from forcefield, molecule
+            off_top = off_mol.to_topology()
+            openmm_top = off_top.to_openmm()
+            openmm_system = off_forcefield.create_openmm_system(off_top)
+            
+            # Need an integrator for simulation even if we don't end up using it really
+            integrator = LangevinIntegrator(300*units.kelvin, 1/units.picosecond, 0.002*units.picoseconds)
```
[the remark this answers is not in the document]
dotsdl: Done! Switched to `VerletIntegrator`.

qcengine/programs/openmm.py
dotsdl: At this point, I don't believe we need anything further. I've cut this bit out (was lifted originally from the `psi4` harness implementation)

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+
+            # TODO: need a way to generate positions from the topology
+            # likely an existing tool in the modules we're using since this is
+            # just a molecule floating in space with no solvent
+            simulation.context.setPositions(pdb.positions)
+
+            # Execute driver
+            try:
+                if input_data.driver == "energy":
+                    simulation.context.setPositions(pdb.positions)
```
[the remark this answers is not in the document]
dotsdl: Indeed, thanks! This is actually a redundant line now, so removed.

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+        return mol
+
+    def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
+        """
+        Runs OpenMM on given structure, inputs, in vacuum.
+        """
+        self.found(raise_error=True)
+
+        from simtk.openmm.app import Simulation
+        from simtk.openmm import units
```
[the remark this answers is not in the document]
dotsdl: Done! Thanks for catching this.

dotsdl:
Implemented recommendations from @jchodera.

Also:
* added an attempt at properly specifying the number of threads we want OpenMM to use
* added rdkit checking in `OpenMMHarness.found` using `RDKitHarness.found`
* cut out unncessary postprocessing bits lifted from psi4 harness
* used `Provenance` object to fall in line with current `RDKitHarness` usage
* `Result` now includes both input data and ouput data, in line with `RDKitHarness` usage

Remaining items include:
* conversion of energies to units of hartree
* conversion of gradients to units of hartree/bohr
* url handling for forcefield *somewhere*

dotsdl:
Sorry for the delay @jchodera; I'm currently traveling, but definitely don't want to be the bottleneck for a much-needed component. I'll be able to iterate quickly to finish out the remaining items and hammer out any newly-noticed issues over the evenings to come. 

There are three items remaining from my view; I'm going to try and address travis next.

qcengine/programs/openmm.py
jchodera: That must be an oversight! It's not [here](https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/unit/unit_definitions.py#L234-L239), but could be easily added.

jchodera: cc: https://github.com/openmm/openmm/issues/2419

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,194 @@
...
+                # TODO: add pull of offxml data from a url
+                pass
+            else:
+                raise InputError("OpenMM requires either `model.offxml` or `model.url` to be set")
+
+            # Process molecule with RDKit
+            rdkit_mol = self._process_molecule_rdkit(jmol)
+
+            # Create an OFF Molecule from the RDKit Molecule
+            off_mol = Molecule(rdkit_mol)
```
dgasmith: Could we note raw import Molecule so it isn't confused without general QC* Molecule object?

qcengine/programs/openmm.py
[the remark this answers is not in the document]
dgasmith: @jchodera Will this work here?

qcengine/programs/tests/test_programs.py
```diff
@@ -200,3 +200,17 @@ def test_random_failure_with_success(failure_engine):
...
+    json_data["molecule"] = qcng.get_molecule("water")
+    json_data["driver"] = "energy"
+    json_data["model"] = {"method": "smirnoff-1.1.0",
+                          "url": None,
+                          "offxml": "test_forcefields/smirnoff99Frosst.offxml"}
+    json_data["keywords"] = {}
+
+    ret = qcng.compute(json_data, "openmm", raise_error=True)
+
+    assert ret.success is True
```
dgasmith: Since the units are in question here could you add a test to `test_procedures::179` which does a good job of checking these values?

qcengine/programs/tests/test_programs.py
```diff
@@ -200,3 +200,17 @@ def test_random_failure_with_success(failure_engine):
...
     assert ret.extras["ncalls"] == 2
+
+@testing.using_openmm
+def test_openmm_task():
+    json_data = copy.deepcopy(_base_json)
+    json_data["molecule"] = qcng.get_molecule("water")
+    json_data["driver"] = "energy"
+    json_data["model"] = {"method": "smirnoff-1.1.0",
+                          "url": None,
+                          "offxml": "test_forcefields/smirnoff99Frosst.offxml"}
```
dgasmith: Is the `offxml` field required here? or is `method` sufficient?

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,194 @@
...
+from ..exceptions import InputError, RandomError, ResourceError, UnknownError
+from ..util import execute, popen, temporary_directory
+from ..units import ureg
+
+from .rdkit import RDKitHarness
+
+
+class OpenMMHarness(ProgramHarness):
+
+    # TODO: verify correctness of these default params
```
dgasmith: `thread_safe` is True/False depending if I can evaluate two instances concurrent in the same Python interpreter. This is fine only if a program does not use global memory and uses something more than PID unique files. This is typically used by our orchestration layer to help determine scheduling.

dotsdl:
Currently, tests fail because these changes are dependent on openmm/openmm#2421 and MolSSI/QCElemental#144 making it into releases for each. I could put in some deliberate hacks to be removed once we have those releases, just to see if this functionality works as expected. Thoughts?

dgasmith:
No, lets push out a new QCElemental release now so that this works.

dgasmith:
QCElemental v0.10.0 is out. Tests still fail, but you should have the ability to run this through CI now without the previous bottleneck.

jchodera:
I know we're hoping to put an OpenMM bugfix release out soon, but it may be a good idea to work around the Hartree issue for now. It might be easy to do that by simply adding
```python
from simtk import unit
if not hasattr(unit, 'hartree'):
    unit.hartree_base_unit = unit.ScaledUnit(4.3597447222071e-18, unit.joule, "hartree", "Ha")
    unit.hartree = unit.hartrees = unit.Unit({unit.hartree_base_unit: 1.0})
```

qcengine/programs/openmm.py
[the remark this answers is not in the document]
jchodera: This should work!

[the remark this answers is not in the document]
jchodera: I think this should work!

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,194 @@
...
+                raise InputError("OpenMM requires either `model.offxml` or `model.url` to be set")
+
+            # Process molecule with RDKit
+            rdkit_mol = self._process_molecule_rdkit(jmol)
+
+            # Create an OFF Molecule from the RDKit Molecule
+            off_mol = Molecule(rdkit_mol)
+
+            # Load an OFF Force Field
+            off_forcefield = ForceField(offxml)
```
jchodera: Creating a new `ForceField` object for every configuration of every molecule we want to evaluate will add a huge amount of overhead. Is there a way we can either memoize/cache this creation or somehow refactor things to allow this engine to apply the force field to many molecules once the `off_forcefield` object is created?

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,194 @@
...
+
+            # Create OpenMM system in vacuum from forcefield, molecule
+            off_top = off_mol.to_topology()
+            openmm_system = off_forcefield.create_openmm_system(off_top)
```
jchodera: If we are asked to compute many gradients/energies for the same molecule, this step can also introduce a significant amount of overhead if we call it over and over again. Can we also memoize/cache the `openmm_system` object if different geometries for the same molecule are called over and over again, or is that difficult because the bonds and bond orders are being perceived from geometry? 

I imagine that we could use the (optional) SMILES field---if these are specified, it should be easy to memoize/cache the `System` object in terms of this SMILES field.

qcengine/programs/tests/test_programs.py
jchodera: How do we want this to work? 

What if we had `method` specify `smirnoff`, to distinguish this from adding support for the native OpenMM ffxml format later (or other formats in the future)?

Then we could have three possibilities:
* `filename` could specify a local path to a standard SMIRNOFF force field installed / distributed by the `openforcefield` toolkit
* `url` could specify a remote URL to a SMIRNOFF `.offxml` file
* `offxml` could specify the actual contents of a SMIRNOFF `offxml` file

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,112 @@
...
+        # TODO: SET WORKDIR TO SCRATCH
+        # Location resolution order config.scratch_dir, /tmp
+        parent = config.scratch_directory
+        with temporary_directory(parent=parent, suffix="_openmm_scratch") as tmpdir:
+
+            # TODO: check what kind of processing we need to do of this input
+            jmol = input_data.molecule
+
+            ff_files = input_data.force_field + input_data.water_model
+            forcefield = app.ForceField(*ff_files)
```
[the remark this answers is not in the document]
dotsdl: Thanks @jaimergp! We're now doing some multiple kwargs unpacking in the returned `Result` object. Wasn't aware this could be done now. :D

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,182 @@
...
+                     raise_msg='Please install via `conda install openmm -c omnia`.')
+
+    @staticmethod
+    def _process_molecule_rdkit(jmol):
+        from rdkit import Chem
+        from rdkit.Chem import AllChem
+
+        # Handle errors
+        if abs(jmol.molecular_charge) > 1.e-6:
+            raise InputError("RDKit does not currently support charged molecules.")
```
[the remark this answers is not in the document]
dotsdl: @dgasmith I don't have a strong opinion on where this goes; your view is more global than mine. It might make sense to put it as a standalone utility function in the `rdkit.py` program module and then import it here. We already import the `RDKitHarness` here since we call its `found` method, so this wouldn't be crossing streams any more than we already are.

qcengine/programs/openmm.py
dotsdl: Done! I switched to doing:
```python
import openforcefield.topology as offtop
from openforcefield.typing.engines import smirnoff
...
...
            # Create an Open Force Field `Molecule` from the RDKit Molecule
            off_mol = offtop.Molecule(rdkit_mol)

            # Load an Open Force Field `ForceField`
            off_forcefield = smirnoff.ForceField(offxml)
```
Is this what you had in mind?

qcengine/programs/openmm.py
dotsdl: Ah, I see. I'm actually not certain how OpenMM behaves in this regard, as I've never tried to run e.g. two `Simulation` instances in different threads on the same interpreter. I'm willing to bet the answer is `False` here. If anyone knows otherwise, please let us know.

jchodera: OpenMM should be thread-safe provided you are using separate `Context` objects in the threads you are trying to perform operations on concurrently.

qcengine/programs/openmm.py
dotsdl: I'm not aware of where QCEngine implements caching. Is this something that currently exists? If so, can you point me to where it is currently used? I can then use this as an example to start from.

qcengine/programs/openmm.py
dotsdl: See question above re: where QCEngine implements caching of computed components for later use. I can them try the approach you've laid out here and we can discuss if this meets our needs or introduces issues (e.g. bond orders). Thanks!

@dgasmith, do you have insights on the question of caching `compute` intermediates from `*Harness` instances?

qcengine/programs/tests/test_programs.py
dotsdl: @dgasmith I'm not sure what you mean here. Can you clarify?

qcengine/programs/openmm.py
jchodera: I don't think QCEngine naturally implements caching since generally the computations take minutes to hours. Since these OpenMM calculations take milliseconds, it may make sense to try to minimize the overhead here. We can certainly implement that later if needed.

qcengine/programs/openmm.py
dgasmith: Yup.

qcengine/programs/openmm.py
dgasmith: See the ANI Harness where we cache the pytorch graph.

qcengine/programs/openmm.py
dgasmith: You could consider building a small LRU cache where the you can use a `Molecule.get_hash() + offxml_uid` to build a unique identifier.

qcengine/programs/tests/test_programs.py
dgasmith: There are geometry optimization tests in `test_procedures`, it would be good to add an OpenMM one there to help verify units.

qcengine/programs/openmm.py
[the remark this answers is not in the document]
dgasmith: Can we add this static method to the RDKit Harness then?

qcengine/programs/openmm.py
jchodera: See [here](https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/torchani.py#L17-L75).

qcengine/programs/tests/test_programs.py
dgasmith: I like `smirnoff-1.1.0` as a full specifier to fully define a single method and can be supported in the database. The additional fields are useful for local computations I believe. A little messy would be `method:smirnoff + url or offxml`. If `Smirnoff-version` is specified url/offxml is ignore with warning?

jchodera: @dgasmith : The problem is that OpenMM is _way_ more general than just being able to compute energies with SMIRNOFF. In principle, we could add support for `gaff` versions, as well as `CGenFF`, or other small molecule force fields that can be computed in OpenMM. 

You could try to ask the `openforcefield` toolkit to resolve the parameter sets in order, and then fall back to something else (GAFF?) if not, but this logic gets complicated and there's no clear way to specify precendence.

We could _make_ this the "openforcefield" harness, rather than the OpenMM harness, but you'd be significantly limiting yourself for no clear reason.

Is there another clearer way we could specify "use the OpenMM engine and compute openforcefield energies with this force field" if you don't like the requirement of specifying either `filename`, `url`, or `offxml` when `"method" : "smirnoff"` is specified? This sounds simple, and we can have a simple precedence order, as well as warn if an unused property is specified.

Is the issue that QCArchive organizes data by `method`?

dgasmith: Yes, it ultimately depends which level of abstraction you are looking for here. Quantum chemistry is pretty easy to generally restrict itself to a few specific field (`method`, `basis`, `keywords`) to fully define its method. Structured data like in QCArchive becomes a bit restricting as its hard to change primary keys. There are certainly ways around this, but outside our design scope at the moment.

> Is there another clearer way we could specify "use the OpenMM engine and compute openforcefield energies with this force field" if you don't like the requirement of specifying either filename, url, or offxml when "method" : "smirnoff" is specified? 

Yes, this is a better phrased version of what I was suggesting. 

QCArchive really only knows about `method` and `basis` fields here. We can either lump all the information into `method` or use `basis` somewhat awkwardly. Additional fields like `url` `filename` etc are of course perfectly fine here at the Engine layer, but will not cooperate with Fractal layers and above.

dgasmith: @dotsdl A quick update here from talking to @jchodera  lets do the following:
1) `method` is always the typing engine (openff/openmm/etc)
2) `basis` is the forcefield name and version down to patch if available (Smirnoff-v1.0.0, UFF, etc)
3) `basis` need not be specified. If a `url` or `offxml` is supplied it will this and fill in the basis later. If the `url/offxml` leads to a non-versioned forcefield the `basis` name will be generated as `f"{forcefield_name}-{hash(forcefield_schema)}"`

I think this covers the bases.

qcengine/programs/openmm.py
dgasmith: Lets update this key to `{mol_hash}-{input_model.method}-{hash(forcefield_schema)}`. Here we will need to change the `mol_hash` to only the symbols + connectivity fields. Since we are using RDKit here anyways, I assume that there is some sort of uniqueness perception available at the right abstraction level (connectivity only). `Molecule.get_hash()` will return a unique hash whenever the cartesian geometry changes so this wouldn't work for a geometry optimization.

On the `lru_cache` I would recommend a depth of maybe ~10. I don't know how big they are in memory, but unless I am doing a geometry optimization I am unlikely to get multiple hits per instantiation.

dotsdl:
Tests passing on every build but the OpenMM-specific one thanks to MolSSI/QCElemental#150. Addressing comments here over the next few nights so we can move this one closer to the finish line.

dgasmith:
@dotsdl Great looking forward to having this in!

jchodera:
A working `openmm` should be available from the `omnia-dev` channel right now.

What about adding a `devtools/conda-envs/openmm.yaml` file based on [`rdkit.yaml`](https://github.com/MolSSI/QCEngine/blob/master/devtools/conda-envs/rdkit.yaml) that contains:
```YAML
name: test
channels:
  - omnia-dev
  - conda-forge
  - rdkit
dependencies:
  - openmm
  - rdkit

    # Core
  - python
  - pyyaml
  - py-cpuinfo
  - psutil
  - qcelemental >=0.9.0
  - pydantic >=0.30.1

    # Testing
  - pytest
  - pytest-cov
  - codecov
```
Also, shouldn't this be testing the `conda-forge` RDKit, rather than the `rdkit` channel RDKit?

jchodera:
You'd of course also have to add a `PROG=OPENMM` branch to the [`.travis.yml`](https://github.com/MolSSI/QCEngine/blob/master/.travis.yml) that would make use of this environment.

dotsdl:
We may be hitting some fresh dependency hell with using `rdkit` from `conda-forge`. I'm not sure quite yet how to interpret the conda errors I'm seeing on Travis for the `OPENMM` build.

qcengine/programs/openmm.py
dotsdl: Great! Switching `thread_safe` to `True`, and adding this proviso as an inline comment for reference.

dotsdl:
@jchodera troubleshooting this now. Is `omnia-dev` a private conda channel?

Lnaden:
The `dev` is a Conda label build from a different github repo, you should be able to access the OpenMM `dev` builds from `conda install -c conda-forge -c omnia/label/dev openmm`

dotsdl:
Ah, thanks @Lnaden! Unfortunately I'm still getting what looks like the same result as when specifying `-c omnia-dev`:

```bash
> conda create -n molssi3 -c omnia/label/dev  openmm
Collecting package metadata (current_repodata.json): done
Solving environment: failed with current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:



Package numpy conflicts for:
openmm -> numpy
Package python conflicts for:
openmm -> python[version='2.7.*|3.5.*|3.6.*|3.7.*']
Package fftw3f conflicts for:
openmm -> fftw3f

```

Does the `openmm` package on `omnia-dev` have dependency spec issues? I'll admit I have very little experience troubleshooting conda issues.

jchodera:
@dotsdl: You'll need to include both the `omnia` and `omnia-dev` channels to satisfy the dependencies for now. 

Don't use the `omnia/label/dev` label---we've deprecated that.

We're nearing the home stretch with conda-forge! We just need to resolve an outstanding issue with OpenCL:
https://github.com/conda-forge/conda-forge-enhancement-proposals/issues/17

dotsdl:
@jchodera awesome, that seems to be the trick! Thank you! I'm also working on the caching components above. I believe I also have enough info to finish out implementing the spec you and @dgasmith decided for inputs. I know we're all looking forward to shipping this.

Lnaden:
Everything on the omnia channel and its labels requires `conda-forge` as a higher order dependency. So I think the fix to the create line there is 

`conda create -n molssi3 -c conda-forge -c omnia/label/dev  openmm`

However, the other thing you can do is save a YAML *similar* to the one John posted [above](https://github.com/MolSSI/QCEngine/pull/151#issuecomment-547143246) and then run the command 

`conda env create -n molssi3 -f openmm.yaml` (openmm.yaml is the file you saved)

The problem you were seeing with dependencies is from being slightly incomplete. This YAML should work (at least it does on my laptop)

```yaml
name: test
channels:
  - conda-forge
  - omnia-dev
  - omnia
dependencies:
  - openmm
  - rdkit

    # Core
  - python
  - pyyaml
  - py-cpuinfo
  - psutil
  - qcelemental >=0.9.0
  - pydantic >=0.30.1

    # Testing
  - pytest
  - pytest-cov
  - codecov
```

Lnaden:
Also, apparently disregard what I said about the `dev` label (i've struck that in my previous comment)

dgasmith:
@dotsdl I think this is ready to go if you can update the test env!

jchodera:
Awesome! Can't wait to get this into a new release.

qcengine/programs/openmm.py
dotsdl: @dgasmith do you know of a "uniqueness perception available at the right abstraction level (connectivity only)" that makes sense to use here? My knowledge of the tooling available to give the right representation with the guarantees we need is limited here.

dotsdl:
@dgasmith, @jchodera, thanks! There are a few outstanding items that I want to sprint on this week. I am aiming to have them ready for review by EOD 11/17:
1. geometry optimization test for OpenMM
2. tests for each of the options available for specifying which forcefield to use
3. caching of OpenForceField `Forcefield` (key: `hashlib.sha1(<xml>).hexdigest()`)
4. caching of OpenMM `System` (key: `{mol_hash}-{input_model.method}-{hashlib.sha1(forcefield_schema).hexdigest()}`)
5. Getting thread specification to work properly (see [TODO](https://github.com/dotsdl/QCEngine/blob/feature-openmm/qcengine/programs/openmm.py#L116)). We may need to resort to an environment variable; guidance on how best to require env variables in QCEngine appreciated.

jchodera:
> geometry optimization test for OpenMM

Is the OpenMM engine supposed to do geometry optimization, or is that driven by GeomeTRIC?

qcengine/programs/openmm.py
dgasmith: I think this should just be symbols + connectivity? Something like:
```
>>> import qcelemental as qcel
>>> caffiene = qcel.models.Molecule.from_data("pubchem:benzene")
	Searching PubChem database for benzene (single best match returned)
	Found 1 result(s)
>>> import json
>>> hash(json.dumps({"connectivity": caffiene.connectivity, "symbols": list(caffiene.symbols)}))
-1973773897279072458
```

You may need to add charge/multiplicity here too, the fields are whatever is used to generate the OpenMM system. Also note that this particular molecule doesn't have connectivity so you may need to add that as perceived by OpenMM as well.

dgasmith:
@jchodera In this case we would need to add OpenMM to the optimization procedures as well. This becomes a bit complex and we would need each step of the geometry optimization and energy from OpenMM. We have yet to do this and do not have an example of using a native geometry optimizer.

dotsdl:
> In this case we would need to add OpenMM to the optimization procedures as well. This becomes a bit complex and we would need each step of the geometry optimization and energy from OpenMM. We have yet to do this and do not have an example of using a native geometry optimizer.

Would this be better to handle as a separate PR? It's not clear to me how big a lift this addition will turn out to be.

dgasmith:
I agree this is for another PR.

qcengine/programs/tests/test_programs.py
dotsdl: We have decided that geometry optimization tests for OpenMM are out of scope for this PR, but should be created as a separate PR.

qcengine/programs/tests/test_programs.py
dotsdl: @dgasmith and @jchodera, is there something to grab onto in a SMIRNOFF XML file indicating `forcefield_name` and `forcefield_version`? For the `offxml` path case we can (somewhat reliably) use `os.path.splitext(os.path.basename(offxml))[0]`, but for the `url` case this may be (even) less reliable. Looking at the [`openff-1.0.0.offxml`](https://raw.githubusercontent.com/openforcefield/openforcefields/master/openforcefields/offxml/openff-1.0.0.offxml) content, I don't see something internal to the structure out of which I can pull this info for populating `basis`.

We could avoid this problem by demanding that `basis` be populated. The danger here of course is that `basis` may not actually match what is given by `offxml` or `url`. I'm not sure how likely this is to be a problem in practice, however.

dgasmith: @j-wags is there a good way to get the openff xml version?

dotsdl: Related question: given our use of pydantic, is it even possible to alter `model.basis` by the time it enters `OpenMMHarness.compute`?

dgasmith: It cannot be changed, but it can be updated.

dgasmith:
Considering merging this with test skipped and breaking remaining questions out into issues. What do you think?

qcengine/programs/openmm.py
dotsdl: LRU cache implemented for OFF `Forcefield`.

dotsdl:
@dgasmith agreed, see my updated [checklist](https://github.com/MolSSI/QCEngine/pull/151#issuecomment-552598728).

The last item is CPU thread number control. Are harnesses able to rely on environment variables for some of their execution input? Currently [these lines](https://github.com/MolSSI/QCEngine/pull/151/commits/03b99b319612887a40f5e9106558a6063671f2b2#diff-562c11b8089e53d28872cf6560eab52eR193-R194) throw an exception, and I get the impression this is because OpenMM's only way to alter the number of threads it uses is via the `OPENMM_CPU_THREADS` environment variable.

If using env variables are fine, these lines should be able to be deleted and OpenMM should respect the env variable.

dotsdl:
This is ready for review, from my perspective, even with [this outstanding issue](https://github.com/MolSSI/QCEngine/pull/151#issuecomment-554839986).

jchodera:
Tagging @peastman regarding setting the number of threads via a Platform property (see directly above). Are we misunderstanding how we should set the number of CPU threads for the CPU platform?

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,239 @@
...
+            # Execute driver
+            try:
+                if input_data.driver == "energy":
+
+                    # Compute the energy of the configuration
+                    state = context.getState(getEnergy=True)
+
+                    # Get the potential as a simtk.unit.Quantity, put into units of hartree
+                    q = (state.getPotentialEnergy() / unit.hartree)
+                    ret_data["return_result"] = q.value_in_unit(q.unit)
```
dgasmith: We always need the energy, should go into `properties.return_energy`.

qcengine/programs/openmm.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if url:
+            # TODO: DOESN'T APPEAR TO BE A WAY TO GET A MEANINGFUL FORCEFIELD DESIGNATOR FROM THE XML CONTENTS ITSELF
+            # PERHAPS THIS SHOULD BE ADDED IN FUTURE RELEASES?
+            return os.path.splitext(os.path.basename(url))[0]
+
+    @staticmethod
+    def found(raise_error: bool = False) -> bool:
+        # this harness requires RDKit as well, so this needs checking too
+        rdkit_found = RDKitHarness.found(raise_error=raise_error)
```
dgasmith: This is nice :)

qcengine/programs/openmm.py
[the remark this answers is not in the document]
peastman: The code given above will not work.  Once the Context is created, the number of threads it uses can't be changed.  Instead, you can use any of three methods to set the number of threads.

1. Set the environment variable `OPENMM_CPU_THREADS`.
2. *Before* creating the Context set the default value for the `Threads` property: `platform.setPropertyDefaultValue('Threads', '4')`
3. Provide a value for that property to the Context constructor: `context = openmm.Context(openmm_system, integrator, platform, {'Threads':'4'})`

If you don't do any of those, it defaults to the number of logical cores in your processor.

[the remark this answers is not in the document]
dotsdl: Excellent, thanks @peastman! I've gone with option (3).

qcengine/programs/openmm.py
dotsdl: Done!

qcengine/programs/openmm.py
dotsdl: Thanks! :smiley:

dotsdl:
Comments resolved. Merging and resolving conflicts with current `master` now.

dotsdl:
`master` merged. Waiting for tests to complete. This work should be good to merge if there are no further concerns. Some follow-up items should be made into separate issues:
1. Caching of `openmm_system` upon creation. This should follow a model similar to our caching of `off_forcefield`, but we need to carefully consider how we generate the key/hash for the `openmm_system`. Whatever we choose to hash should be invariant to rotations or translations of the system, but should capture connectivity, charge states, etc.
2. Generation of `basis` if not specified. Ideally `basis` is generated from `forcefield_name` and `forcefield_version` from the OFFXML file, but these do not appear to be included in current OFFXML forcefields at this time.

dotsdl:
Tests pass, please let me know if anything further is needed!

peastman:
> we need to carefully consider how we generate the key/hash for the openmm_system.

One option for this is to serialize the System to XML with `XmlSerializer.serialize(system)` then compute a hash from the string.  This will detect any change to the System or the Forces it contains, but will be unaffected by changes to particle positions.

Issue #86, opened 2019-06-06
Add types to execute function of ProgramHarness base class


sjrl:
A reminder to add types to the `execute` function of the `ProgramHarness` base class and then propagate those changes to all Harnesses. 

https://github.com/MolSSI/QCEngine/blob/f76e1070b30c4c5c806c91c80df579affb55029a/qcengine/programs/model.py#L61

sjrl:
I just noticed that the bool type was added to the `extra_commands` option in the Molpro program harness, 
https://github.com/MolSSI/QCEngine/blob/971c7d2a150c776cf8fd275df2157a14d929dd3f/qcengine/programs/molpro.py#L111-L119
but it's currently being treated as if it were a list. 
https://github.com/MolSSI/QCEngine/blob/971c7d2a150c776cf8fd275df2157a14d929dd3f/qcengine/programs/molpro.py#L134-L137
It does not currently have a type set in the base `execute` function of the `ProgramHarness` so I wanted to ask how is the `extra_commands` option intended to be used?

dgasmith:
I think that must be a mistake. It should be `List[str]` no? @loriab

loriab:
Yes, `Optional[List[str]]`. It's also deceptively named because that list isn't of additional commands (plural) but of a single replacement command where the list is arguments (e.g., `['ls', '-l']`).

mattwelborn:
@sjrl Does the merging of PR #143 close this issue?

sjrl:
Not entirely. The `execute` function in the base class is still missing a `return` type. I think I missed adding it there since the `execute` functions in different ProgramHarnesses were returning slightly different objects. I think it would be safe to add the return type to be `-> Tuple[bool, Dict[str, Any]]`

PR #152, opened 2019-09-09
Model: Addressing issue 86 

sjrl:
Finished up issue https://github.com/MolSSI/QCEngine/issues/86 by adding a return type for the base execute function of ProgramHarness.


dgasmith:
Thanks Sebastian!

PR #153, opened 2019-09-11
Adding ensemble related quantities to extras in torchani.py

farhadrgh:
The following quantities will be recorded in `extras` when `energy` driver is called:

- `ensemble_energies`: An energy array of all members (models) in an ensemble of models.
- `ensemble_energy_avg`: The average value of energy array which is also recorded with as `energy` in QCEngine.
- `ensemble_energy_std`: The standard deviation of energy array.
- `ensemble_per_root_atom_disagreement`: The standard deviation scaled by the square root of N, with N being the number of atoms in the molecule. This is the quantity used in the query-by-committee (QBC) process in active learning to infer the reliability of the models in an ensemble, and produce more data points in the regions where this quantity is below a certain threshold (inclusion criteria).

### DIFF
--- a/qcengine/programs/torchani.py
+++ b/qcengine/programs/torchani.py
@@ -55,6 +55,14 @@ class TorchANIHarness(ProgramHarness):
         import torch
         import torchani
 
+        # graft a custom forward pass to Ensemble class
+        def ensemble_forward(self, species_input):
+            outputs = torch.cat([x(species_input)[1] for x in self])
+            species, _ = species_input
+            return species, outputs
+
+        torchani.nn.Ensemble.forward = ensemble_forward
+
         if name == "ani1x":
             self._CACHE[name] = torchani.models.ANI1x()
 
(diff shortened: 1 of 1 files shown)


qcengine/programs/torchani.py
```diff
@@ -96,17 +104,29 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+        _, energy_array = model((species, coordinates))
+        energy = energy_array.mean()
+        ensemble_std = energy_array.std()
+        ensemble_scaled_std = ensemble_std / np.sqrt(num_atoms)
+
         ret_data["properties"] = {"return_energy": energy.item()}
 
         if input_data.driver == "energy":
             ret_data["return_result"] = ret_data["properties"]["return_energy"]
+            ret_data["extras"] = {
```
dgasmith: This should be provided regardless of the driver? Maybe place it below near the provenance information below.

qcengine/programs/torchani.py
```diff
@@ -96,17 +104,29 @@ def compute(self, input_data: 'ResultInput', config: 'JobConfig') -> 'Result':
...
+                "ensemble_energies": energy_array.detach().numpy(),
+                "ensemble_energy_avg": energy.item(),
+                "ensemble_energy_std": ensemble_std.item(),
+                "ensemble_per_root_atom_disagreement": ensemble_scaled_std.item()
```
dgasmith: Can you add your notes in the PR to the items here so that we have a record of exactly what these mean? We can then move this into the documentation pages when we get to them.

farhadrgh: fixed in 6770714

qcengine/programs/torchani.py
farhadrgh: fixed in 6770714

PR #154, opened 2019-09-11
Molpro properties

sjrl:
Improvements to the Molpro ProgramHarnes

1. Added support for some calinfo_* properties (`nalpha`, `nbeta`, and `nbasis`)
2. Added support for unrestricted HF and DFT calculations. Did this by adding keyword `reference`
3. Started to add support for parsing local correlation calculations in Molpro

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -164,6 +164,12 @@ class MolproHarness(ProgramHarness):
         if template is None:
             input_file = []
 
+            # TODO Resolve keywords
+            unrestricted = False
+            if "reference" in input_model.keywords:
+                if input_model.keywords["reference"] == "unrestricted":
+                    unrestricted = True
+
             # Memory is in megawords per core for Molpro
             memory_mw_core = int(config.memory * (1024**3) / 8e6 / config.ncores)
             input_file.append("memory,{},M".format(memory_mw_core))
@@ -190,7 +196,7 @@ class MolproHarness(ProgramHarness):
             energy_call = []
             # If post-hf method is called then make sure to write a HF call first
             if input_model.model.method.upper() in self._post_hf_methods:  # post SCF case
-                if input_model.keywords.reference == "unrestricted":
+                if unrestricted:
                     energy_call.append('{UHF}')
                     energy_call.append('')
                     energy_call.append(f'{{{input_model.model.method}}}')
@@ -200,12 +206,12 @@ class MolproHarness(ProgramHarness):
                     energy_call.append(f'{{{input_model.model.method}}}')
             # If DFT call make sure to write {rks,method}
             elif input_model.model.method.upper() in self._dft_functionals:  # DFT case
-                if input_model.keywords.reference == "unrestricted":
+                if unrestricted:
                     energy_call.append(f'{{uks,{input_model.model.method}}}')
                 else:
                     energy_call.append(f'{{rks,{input_model.model.method}}}')
             elif input_model.model.method.upper() in self._scf_methods:  # HF case
-                if input_model.keywords.reference == "unrestricted":
+                if unrestricted:
                     energy_call.append('{UHF}')
                 else:
                     energy_call.append('{RHF}')
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -164,7 +164,7 @@ class MolproHarness(ProgramHarness):
         if template is None:
             input_file = []
 
-            # TODO Resolve keywords
+            # Resolving keywords
             unrestricted = False
             if "reference" in input_model.keywords:
                 if input_model.keywords["reference"] == "unrestricted":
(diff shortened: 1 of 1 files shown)


qcengine/programs/molpro.py
```diff
@@ -45,11 +45,20 @@ class MolproHarness(ProgramHarness):
...
+    #       RHF-RMP2 (like RHF-UMP2 using Molpro's naming convention for CC methods)
+    #       RHF-UCCSD
+    #       RHF-UCCSD(T)
+    #       RHF-RCCSD
+    #       RHF-RCCSD(T)
+    # - Unrestricted references (Only supported up to UMP2, no CC support)
+    #       UHF-UMP2
+    # NOTE: Unrestricted SCF methods must be specified by using keyword reference
+    _hf_methods: Set[str] = {"HF", "RHF"}
+    _restricted_post_hf_methods = {"MP2", "CCSD", "CCSD(T)"}  # RMP2, RCCSD, RCCSD(T)}
```
dgasmith: Any reason these do not have types?

qcengine/programs/molpro.py
```diff
@@ -151,12 +160,15 @@ def build_input(self, input_model: 'ResultInput', config: 'JobConfig',
         if template is None:
             input_file = []
 
+            # Resolving keywords
+            caseless_keywords = {k.lower(): v for k, v in input_model.keywords.items()}
+            unrestricted = False
+            if "reference" in caseless_keywords and caseless_keywords["reference"] == "unrestricted":
```
dgasmith: Did you want to autodetect based on multiplicity like the Psi4 driver?

qcengine/programs/molpro.py
```diff
@@ -329,37 +375,64 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
             del properties['ccsd_singlet_pair_energy']
             del properties['ccsd_triplet_pair_energy']
 
+        # Process data in molpro_variable_map
+        # Note: For the DFT case molecule_method is the name of the functional plus R or U in front
+        molecule = root.find('molpro_uri:job/molpro_uri:molecule', name_space)
+        molecule_method = molecule.attrib['method']
+        molecule_final_energy = float(molecule.attrib['energy'])  # Energy from the molecule tag in case its needed
+        # Loop over each variable under the variables tag to grab additional info from molpro_variable_map
+        for variable in molecule.findall('molpro_uri:variables/molpro_uri:variable', name_space):
```
dgasmith: Bleh, XML fun. I don't see a way to make this simpler however.

qcengine/programs/molpro.py
sjrl: Nope, just blanked on that. I'll add that in.

qcengine/programs/molpro.py
sjrl: No I was thinking to follow the Molpro default which `ROHF` is done by default for open-shell calculations with only `UHF` performed if requested.

dgasmith: Ah gotcha, can we add a comment here to this effect?

qcengine/programs/molpro.py
sjrl: Yes I agree, it's starting to feel somewhat unwieldy although I haven't thought of a good way to simplify. In the future we could write separate functions that parse the different sections of the xml (e.g. energies, dipoles, gradient, variables, basis, etc.). That would at least make it more readable.

qcengine/programs/molpro.py
sjrl: Yes definitely.

qcengine/programs/molpro.py
loriab: hmm, I've been considering adding extra info to `model.method` as a violation of the QCSchema contract.

* `tce-ccsd` in nwc trying to add routing info to method
* `df-mp2` once upon a time in p4 trying to add algorithm info to method
* `uccsd` trying to add reference info to method
* I thought I remembered a gamess example last week but can't find it now.

my solution has been that the qcng interface must be taught a new keyword -- `qc_module` or `reference`

qcengine/programs/molpro.py
```diff
@@ -45,11 +45,20 @@ class MolproHarness(ProgramHarness):
...
-    _post_hf_methods: Set[str] = {'MP2', 'CCSD', 'CCSD(T)'}
-    _supported_methods: Set[str] = {*_scf_methods, *_post_hf_methods}
-
+    # Different open-shell scenarios:
+    # - Restricted references
+    #       RHF-RMP2 (like RHF-UMP2 using Molpro's naming convention for CC methods)
+    #       RHF-UCCSD
+    #       RHF-UCCSD(T)
+    #       RHF-RCCSD
+    #       RHF-RCCSD(T)
```
loriab: btw, if you want to join in the "standard_suite" make-the-programs-match fun, https://lists.uni-mainz.de/sympa/arc/cfour/2019-09/msg00030.html may be handy for us all.

qcengine/programs/molpro.py
sjrl: I think I'd be in support of needing to specify exactly what's in QCSchema for `model.method`. Adding new keywords would then make it more consistent between different ProgramHarnesses (i.e. I would need to find the potentially different keyword that controls density fitting in Molpro and Psi4 instead of allowing `df-mp2` to be a valid `model.method` for Molpro and not Psi4.

PR #156, opened 2019-09-20
Skip broken intel-openmp 2019.5-281

mattwelborn:
This PR skips intel-openmp version 2019.5-281, which breaks geometric+psi4. See also ContinuumIO/anaconda-issues#11294.


dgasmith:
Unsure how this made it through, but you need a:
```
ret_data["extras"].update({
...
})
```
[here](https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/torchani.py#L154) instead of a set.

PR #158, opened 2019-09-23
Entos Update

sjrl:
- Updated the output parsing to read the json dictionary produced by entos
- Updated the input file generation

### DIFF
--- a/qcengine/programs/entos.py
+++ b/qcengine/programs/entos.py
@@ -26,7 +26,9 @@ class EntosHarness(ProgramHarness):
     version_cache: Dict[str, str] = {}
 
     # Energy commands that are currently supported
-    _energy_commands = {"dft", "xtb"}
+    _energy_commands = {"dft",
+                        # "xtb"
+                        }
 
     # List of DFT functionals
     _dft_functionals = {
@@ -38,10 +40,10 @@ class EntosHarness(ProgramHarness):
 
     # Available keywords for each of the energy commands
     _dft_keywords = {"df"}
-    _xtb_keywords = {}
+    # _xtb_keywords = {}
     _keyword_map = {
         "dft": _dft_keywords,
-        "xtb": _xtb_keywords
+        # "xtb": _xtb_keywords
     }
 
     class Config(ProgramHarness.Config):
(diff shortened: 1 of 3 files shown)


sjrl:
This will only pass the CI once PR https://github.com/MolSSI/QCEngineRecords/pull/18 goes through

qcengine/programs/entos.py
```diff
@@ -85,7 +107,7 @@ def execute(self,
             infiles.update(extra_infiles)
 
         # Collect all output files and extend with with extra_outfiles
-        outfiles = ["dispatch.out"]
```
dgasmith: Do we need to bump the minimum version to 0.6?

qcengine/programs/entos.py
```diff
@@ -24,6 +25,27 @@ class EntosHarness(ProgramHarness):
...
+                                  # "xtb"
+                                  }
+
+    # List of DFT functionals
+    _dft_functionals: Set[str] = {
+        "SLATER", "DIRAC", "SLATERD3", "DIRACD3", "VWN5", "VWN", "VWN1", "SVWN", "LDA", "BLYP", "BPW91",
+        "BLYPD3", "B88", "PBEX", "PBERX", "PBEC", "LYP", "PW91", "P86", "PBE", "PBER", "PBED3", "PBERD3",
+        "B3LYP3", "B3LYP", "B3LYP5", "PBE0", "PBE1PBE", "B3LYP3D3", "B3LYPD3", "B3LYP5D3", "PBE0D3",
+        "PBE1PBED3", "CAMB3LYP", "WB97X", "CAMB3LYPD3", "WB97XD3"
+    }
```
dgasmith: Ouch, any way to avoid this in the future you think?

qcengine/programs/entos.py
sjrl: Yes, I'll make that change.

qcengine/programs/entos.py
sjrl: Hmm potentially. Entos provides their documentation in an xml format which does contain this list of functionals under the `xc` option tag. It could be parsed from there as long as the documentation file is always accessible. I'm not sure if this will always be the case depending on how the code is distributed.

PR #160, opened 2019-09-27
Turbomole

eljost:
Dear developers,

here is my first try on the Turbomole-Harness. Currently supported are
* HF/DFT (at least the xcfuncs hardcoded in define, the libxc ones not yet)
* RICC2 and RIMP2
* Restricted/unrestricted modes (unrestricted can be forced for closed shell systems)
* RIJ/RIJK/MARIJ for speedup
* Dispersion correction D3Zero and D3BJ
* OMP/SMP parallelization (MPI parallelization not yet)

The basic structure is derived from the NWChem-Harness. All Turbomole (TM) code
can be found in the `qcengine/programs/turbomole` subfolder. Besides the __init__.py
there are currently three files:
1. define.py
    define is a utility to set up a TM calculation. It intended to be used interactivly.
    I coded up a wrapper to to use it "automatically" from python. `prepare_stdin`
    creates a STDIN that can be piped to `define` using `subprocess.Popen.communicate`.
    To do this I actually had to create a separate function `execute_define`, as
    `util.execute` does not seem to support the `communicate()`-stuff. Also the encoding
    of the STDOUT generated by `define` has to be handled in a special way, because
    TM has some quirks in this regard.
    Some stuff is still hardcoded (Convergence to 1e-8 au and max. 150 cycles) but that
    could be changed later on.

2. methods.py
    Really crude hardcoded dicts of supported keywords and methods.
    
3. harvester.py
    Parser for the output generated by TM. I already had working parsers
    in [my own code](https://github.com/eljost/pysisyphus/blob/master/pysisyphus/calculators/parser.py) but they rely on `pyparsing`. So I recoded the relevant
    stuff using only the `re` module of the stdlib.
    Right now they only parse the gradient and the relevant energies.

4. runner.py
    `build_input` prepares the calculation. In TM every calculation needs it's own
    directory that contains all relevant files (control, basis, auxbasis, mos/alpha/beta).
    To actually run the calculation you have to (potentially) call several executables.
    E.g. for a RI-CC2 calculation you first have to generate a reference WF with `dscf`
    or `ridft` and then call `ricc2` on top. Right now this workflow is not easily supported by `util.execute()`. There are 3 ways to do it:
    1. Create a shellscript containing all relevant commands and execute it as
       single command. Sounds a bit hacky but would probably work without problems,
       unless the chosen shell is missing ... (!#/bin/bash etc. )
    2. Execute the programes one-by-one and take care to reuse the scratch-folder. Would
       proably work.
    3. The (for me easiest) route: Add a optional (`bool`) `shell` argument to
       `util.execute` so the provided command is run in a shell.
    
    The beginnings for hessian calculations are also there, but the parser function is still missing. Memory settings are ignored right now.

Throught the files there are several `TODO` sprinkled in where I could need an opinion.
I guess an opinion would be especially relevant for `execute_define` and if it should
have a timeout. Right now there is no real error-checking and if define hangs the whole script will hang indefinitely, so this will have to be fixed somehow.

Added also several tests in a TM-specific file `test_turbomole.py` and in the standard HF test file `test_standard_suite_hf.py`.

Tests can be run by `pytest -k `test_turbomole`. I tested it on Turbomole V7.4 and V7.3.

So hopefully some of the code helps and I'm open to any critisism. If you have some ideas how to improve everything please tell me.

Best regards,
Johannes


qcengine/programs/turbomole/define.py
```diff
@@ -0,0 +1,209 @@
...
+
+def execute_define(stdin: str, cwd: Optional["Path"] = None) -> str:
+    # TODO: replace this with a call to the default execute provided by QCEngine
+    # if possible. May be difficult though, as we have to pipe in stdin and
+    # be careful with the encoding.
+
+    # We cant use univeral_newlines=True or text=True in Popen as some of the
+    # data that define returns isn't proper UTF-8, so the decoding will crash.
+    # We will decode it later on manually.
+    proc = Popen("define",
```
dgasmith: `with Popen(...) as proc:`, best to open these as context so that we do not need to drop them.

qcengine/programs/turbomole/define.py
```diff
@@ -0,0 +1,209 @@
...
+                 stdin=PIPE, stdout=PIPE, stderr=PIPE, cwd=cwd,
+                 # stdin=PIPE, stderr=PIPE, cwd=cwd,
+    )
+    # TODO: Add timeout? Unless the disk hangs this should never take long...
```
dgasmith: Yes please, something like 30 seconds? Anything is better than nothing.

qcengine/programs/turbomole/define.py
```diff
@@ -0,0 +1,209 @@
...
+    # TODO: If used with timeout an exception will be thrown. Then how do we
```
dgasmith: This data should be contained within the `error.error_message` field. You can also simply raise a message like [here](https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/psi4.py#L134) and everything will be handled automatically.

qcengine/programs/turbomole/runner.py
```diff
@@ -0,0 +1,214 @@
...
+                              input_model.molecule.molecular_multiplicity,
+                              geoopt,
+        )
+        with temporary_directory(suffix="_define_scratch") as tmpdir:
+            tmpdir = Path(tmpdir)
+            with open(tmpdir / "coord", "w") as handle:
+                handle.write(coord_str)
+            stdout = execute_define(stdin, cwd=tmpdir)
+            # The define scratch will be populated by some files that we want to keep
+            to_keep = "basis auxbasis coord control alpha beta mos".split()
```
dgasmith: Can we switch the splits to explicit lists when you have the chance? Or is this easy to do from the docs?

dgasmith:
@eljost Gentle poke here, I think its getting very close!

qcengine/programs/turbomole/runner.py
eljost: fixed in d0ddc48

qcengine/programs/turbomole/define.py
eljost: fixed in 049b4f8, b18a244 and 12854d6

qcengine/programs/turbomole/define.py
eljost: fixed in b18a244

qcengine/programs/turbomole/define.py
eljost: (hopefully?) fixed in 12854d6. I am not completly if i did it correctly ...

eljost:
So I create a PR here https://github.com/MolSSI/QCEngineRecords/pull/19 that should add some Records.

eljost:
Hmm, somehow I cant open the LGTM.com website to view the alerts... I wonder if my unversity blocks access?!

dgasmith:
Thats very odd, LGTM is open without access restrictions.  I wouldn't worry too much about the LGTM alerts.

qcengine/programs/tests/test_turbomole.py
```diff
@@ -0,0 +1,94 @@
...
+
+    assert res.driver == "gradient"
+    assert res.success is True
+
+    grad = res.return_result
+    grad_norm = np.linalg.norm(grad)
+    assert compare_values(ref_norm, grad_norm)
+
+
+def test_turbomole_ri_dsp(h2o):
```
dgasmith: This needs a `testing.using_turbomole` decorator.

qcengine/programs/turbomole/runner.py
```diff
@@ -0,0 +1,214 @@
...
+
+        retres = qcvars[f'CURRENT {input_model.driver.upper()}']
+        if isinstance(retres, Decimal):
+            retres = float(retres)
+        elif isinstance(retres, np.ndarray):
+            retres = retres.ravel().tolist()
+
+        output_data = {
+            'schema_name': 'qcschema_output',
+            'schema_version': 1,
```
dgasmith: The schema* data is automatically filled.

qcengine/programs/turbomole/runner.py
```diff
@@ -0,0 +1,214 @@
...
+            'properties': {},
+            'provenance': Provenance(creator="Turbomole", version=self.get_version(),
+                                     routine="turbomole"),
+            'return_result': retres,
+            'stdout': stdout,
+        }
+
+        # got to even out who needs plump/flat/Decimal/float/ndarray/list
+        # Decimal --> str preserves precision
+        output_data['extras']['qcvars'] = {
```
dgasmith: I don't think we need this anymore. We have serializations that can handle complex python types.

Double checking with @loriab.

qcengine/programs/turbomole/runner.py
```diff
@@ -0,0 +1,214 @@
...
+            qcvars['CURRENT GRADIENT'] = gradient
+
+        if hessian is not None:
+            qcvars['CURRENT HESSIAN'] = hessian
+
+        retres = qcvars[f'CURRENT {input_model.driver.upper()}']
+        if isinstance(retres, Decimal):
+            retres = float(retres)
+        elif isinstance(retres, np.ndarray):
+            retres = retres.ravel().tolist()
```
dgasmith: Definitely leave these as NumPy arrays, they will just be case back.

qcengine/programs/tests/test_turbomole.py
eljost: fixed ind bf18f59

qcengine/programs/turbomole/runner.py
eljost: fixed in 434936a

qcengine/programs/turbomole/runner.py
eljost: fixed in 61983d0

qcengine/programs/turbomole/runner.py
eljost: I commented this out for now in e851df6

eljost:
- Added tests to test_harness_canonical.py and fixed the bugs that arose along the way
- I reworked the define call a bit so that InputError is raised when it times out.
- The supplied methods is now checked against a list of methods that are acutally supported and InputError is raised when the requested method is not in the list
- Fixed several LGTM warnings

dgasmith:
Ok, lets go for it. Thanks @eljost for this PR!

PR #163, opened 2019-09-30
0.11 Changelog and GitHub Templates

Lnaden:
This updates the change log to 0.11.0.

Updates the Elemental pins to 0.9 (not yet released), so I don't suspect this will pass or even run until then.

I realized I have the month wrong on 0.10's release, so the link on the GitHub releases page for that release change log needs to be edited. This does not change the release itself, its just the URL link.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,42 @@
+# How to contribute
+
+We welcome contributions from external contributors, and this document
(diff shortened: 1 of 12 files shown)


Lnaden:
With the release of Elemental 0.9.0, this is passing now. I'll merge this and if all looks good for release, it can be cut in the morning.

PR #165, opened 2019-10-08
Psi4 Wavefunctions

dgasmith:
## Description
Updates Psi4 backend to use the new `qcschema` runner so that we can both avoid JSON serialization (msgpack instead) and run the object directly in memory. This has the added benefit that Wavefunction quantities are baked in.

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -132,12 +132,12 @@ class Psi4Harness(ProgramHarness):
 
             else:
 
-                if ("interactive" in input_model.extras) and input_model.extras["interactive"]:
+                if ("psiapi" in input_model.extras) and input_model.extras["psiapi"]:
                     import psi4
                     psi4.core.set_num_threads(config.ncores, quiet=True)
                     psi4.set_memory(f"{config.memory}GB", quiet=True)
                     output_data = psi4.schema_wrapper.run_qcschema(input_model).dict()
-                    output_data["extras"]["interactively_executed"] = True
+                    output_data["extras"]["psiapi_evaluated"] = True
                     success = True
                 else:
                     run_cmd = [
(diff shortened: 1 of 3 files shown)


qcengine/testing.py
```diff
@@ -116,6 +116,7 @@ def get_job(self):
     "dftd3": which('dftd3', return_bool=True),
     "geometric": which_import("geometric", return_bool=True),
     "psi4": is_program_new_enough("psi4", "1.2"),
+    "psi4_14": is_program_new_enough("psi4", "1.4a2.dev160"),
```
loriab: I'd call this `"psi4_runschema"` or something instead of `"psi4_14"`. but not terribly important.

qcengine/programs/psi4.py
```diff
@@ -57,64 +57,113 @@ def compute(self, input_model: 'ResultInput', config: 'JobConfig') -> 'Result':
...
-                        # older c. 1.3 message-only run_json
-                        output_data["error"] = {"error_type": "internal_error", "error_message": output_data["error"]}
             else:
-                output_data = input_data
-                output_data["error"] = {"error_type": "execution_error", "error_message": output["stderr"]}
 
-        # Reset the schema if required
-        output_data["schema_name"] = "qcschema_output"
+                if ("psiapi" in input_model.extras) and input_model.extras["psiapi"]:
+                    import psi4
```
loriab: I strongly expect that we'll want [this trio](https://github.com/psi4/psi4/blob/master/psi4/driver/schema_wrapper.py#L205-L207) here. ok to merge, and I'll find out for sure with DDD.

dgasmith: This should happen Psi4 side IMO and not here.

PR #170, opened 2019-10-31
Docs and Coverage

dgasmith:
## Description
Adds additional tests and builds some general testing functionality. Notes:
 - QCEngineRecords is now globally pinned to commit for easier use and so previous commits work.
 - Azure now uploads codecov details.
 - Utils have additional tests with a bug found in temporary directory (str returned not Path)!
 - Light docs overhaul, trying a table to give very quick feature overviews.

@sjrl Is the Entos wrapper production ready?

@loriab It might be worth getting some GAMMES/NWChem/CFOUR output files in QCEngineRecords at this point.

## Status
- [ ] Changelog updated
- [x] Ready to go


docs/source/program_overview.rst
```diff
@@ -0,0 +1,62 @@
+Program Overview
+================
+
+The general capabalities available through QCEngine for each program can be
+found below:
+
+Quantum Chemistry
+-----------------
+
++------------+------------+------------+-------------+--------------+
```
mattwelborn: Very happy to have this table.

sjrl: Me too! Although would it make sense to have separate check marks for E/G/H since I don't think Molpro currently has Hessian support through QCEngine?

sjrl:
I haven't run the entos interface against the current master version of entos so I'll do that tomorrow to make sure nothing has been broken and update the entos version accordingly. Otherwise it's currently good for DFT Energies and Gradients.

Also what is considered production ready in this case?

docs/source/program_overview.rst
dgasmith: @sjrl Good point and we have the room.

loriab:
I thought QCEngineRecords was relegated to second class compared to actually running the program in qcng CI proper and checking values? Sets of three files per each method, deriv, option (ae/fc) that are continuously changing slightly is  rather a pain.

dgasmith:
Getting CI up and running has been troublesome. I am not sure if we can handle the license issues. I would certainly prefer to run everything using the program itself, but cleaning up the records solution works for now.

PR #171, opened 2019-11-01
Molpro standard suite

sjrl:
- Added Molpro to some of the standard suite (hf, mp2, ccsd). However, Molpro only works in the hf standard suit right now since, a new keyword to turn off frozen-core will need to implemented for the Molpro Harness to pass the mp2 and ccsd standard suite. I've left those commented out for the time being.
- Added a note in the parse_output function outlining the spherical basis set ordering using in Molpro up to angular momentum i.

### DIFF
--- a/qcengine/programs/molpro.py
+++ b/qcengine/programs/molpro.py
@@ -70,6 +70,7 @@ class MolproHarness(ProgramHarness):
                      raise_error=raise_error,
                      raise_msg='Please install via https://www.molpro.net/')
 
+    # TODO Consider changing this to use molpro --version instead of performing a full execute
     def get_version(self) -> str:
         self.found(raise_error=True)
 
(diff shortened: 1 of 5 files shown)


qcengine/programs/molpro.py
```diff
@@ -428,15 +437,21 @@ def parse_output(self, outfiles: Dict[str, str], input_model: 'ResultInput') ->
...
 
+        # Initialize output_data by copying over input_model.dict()
+        output_data = input_model.dict()
+
         # Determining return_result
-        if "return_result" not in output_data:
+        if input_model.driver == "energy":
             output_data["return_result"] = final_energy
+        elif input_model.driver == "gradient":
+            output_data["return_result"] = properties["gradient"]
```
dgasmith: `output_data["return_result"] = properties.pop("gradient")`

sjrl:
No problem!

PR #175, opened 2019-11-16
MPI/MultiNode support

dgasmith:
## Description
This PR adds some hooks which will allow multimode support.
- [x] Adds `nnodes` to config.
- [x] Renames JobConfig to TaskConfig which matches our general nomenclature much better.
- [ ] Do we need an option to allow non-`mpi` cli options?

@lward

## Status
- [ ] Changelog updated
- [x] Ready to go

### DIFF
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -81,7 +81,7 @@ class NodeDescriptor(pydantic.BaseModel):
         extra = "forbid"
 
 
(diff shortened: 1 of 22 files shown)


dgasmith:
@WardLT does this seem reasonable?

dgasmith:
@loriab Can I get a quick review on this?

PR #178, opened 2019-11-22
WIP: Orca Harness.

muammar:
## Description

Orca Harness

## Status
- [ ] WIP
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/orca.py
+++ b/qcengine/programs/orca.py
@@ -1,0 +1,457 @@
+"""
+Calls the Orca executable.
+"""
+
+import string
+import xml.etree.ElementTree as ET
+from typing import Any, Dict, List, Optional, Set, Tuple
+
+from qcelemental.models import AtomicResult
+from qcelemental.util import parse_version, safe_version, which
+
+from ..exceptions import InputError, UnknownError
+from ..util import execute
+from .model import ProgramHarness
+
+
+class OrcaHarness(ProgramHarness):
+    _defaults: Dict[str, Any] = {
+        "name": "Orca",
+        "scratch": True,
+        "thread_safe": False,
+        "thread_parallel": True,
+        "node_parallel": False,
+        "managed_memory": True,
+    }
+    version_cache: Dict[str, str] = {}
+
+    # Set of implemented dft functionals in Molpro according to dfunc.registry (version 2019.2)
+    # fmt: off
+    _dft_functionals: Set[str] = {
+        "B86MGC", "B86R", "B86", "B88C", "B88", "B95", "B97DF", "B97RDF", "BR", "BRUEG", "BW", "CS1", "CS2", "DIRAC",
+        "ECERFPBE", "ECERF", "EXACT", "EXERFPBE", "EXERF", "G96", "HCTH120", "HCTH147", "HCTH93", "HJSWPBEX", "LTA",
+        "LYP", "M052XC", "M052XX", "M05C", "M05X", "M062XC", "M062XX", "M06C", "M06HFC", "M06HFX", "M06LC", "M06LX",
+        "M06X", "M12C", "MK00B", "MK00", "P86", "PBEC", "PBESOLC", "PBESOLX", "PBEXREV", "PBEX", "PW86", "PW91C",
+        "PW91X", "PW92C", "STEST", "TFKE", "TH1", "TH2", "TH3", "TH4", "THGFCFO", "THGFCO", "THGFC", "THGFL", "TPSSC",
+        "TPSSX", "VSXC", "VW", "VWN3", "VWN5", "XC-M05-2X", "XC-M05", "XC-M06-2X", "XC-M06-HF", "XC-M06-L", "XC-M06",
+        "XC-M08-HX", "XC-M08-SO", "XC-M11-L", "XC-SOGGA11", "XC-SOGGA11-X", "XC-SOGGA", "FRMTST", "LHF", "TLHF",
+        "LXBECKE", "ELP", "NULL", "YTEST", "TREF2", "TREF", "TTEST", "GLE", "GREEN", "SRB88", "SRLYP", "LB94", "EI",
+        "SAOP", "USER", "INE", "ECERF2", "ECERFINTER", "ECERFLOCAL2", "ECERFLOCAL", "EXERFLOCAL", "FC", "FCFO", "FCO",
+        "FL", "XC-M11", "PBEXANAL", "PBECANAL", "PBESOLCANAL", "PBESOLXANAL", "EXSRLDA", "EXSRLPBE", "ECSRLPBE",
+        "ECSRLLPBE", "ECSQRTLPBE", "ECMUDIVLPBE", "EXERFPHS", "ECLERFMUPBE", "ECERFERFCPBE", "ECSQRTLDA", "REVPBEX",
+        "B", "B-LYP", "BLYP", "B-P", "BP86", "B-VWN", "B3LYP", "B3LYP3", "B3LYP5", "B88X", "B97", "B97R", "BECKE",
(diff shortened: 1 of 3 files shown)


muammar:
> This pull request **introduces 5 alerts** when merging [97c63cc](https://github.com/MolSSI/QCEngine/commit/97c63ccb28d32bf725d97c5afe0e1d138b4e5ca1) into [f9fea80](https://github.com/MolSSI/QCEngine/commit/f9fea80e012ecebb51e9b44c37b270f070200526) - [view on LGTM.com](https://lgtm.com/projects/g/MolSSI/QCEngine/rev/pr-a1fa87ba11cadf81a17612e141e65b584b6bb064)
> 
> **new alerts:**
> 
> * 2 for Unused local variable
> * 2 for Unused import
> * 1 for Variable defined multiple times

yeah I know, good bot

muammar:
@dgasmith would it be possible to have the first review to know what steps to take forward? The only check not passing right now is the CI. I would not know how to do this for Orca because it has a very particular way to be installed :(.

qcengine/programs/orca.py
```diff
@@ -0,0 +1,297 @@
...
+    ) -> Dict[str, Any]:
+
+        if template is None:
+            input_file = []
+
+            # Resolving keywords
+            # caseless_keywords = {k.lower(): v for k, v in input_model.keywords.items()}
+
+            # Write the geom
+            xyz_block = input_model.molecule.to_string(dtype="orca", units="Angstrom")
```
loriab: there isn't an orca dtype at the moment. you can add one [here](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/to_string.py#L386) or work with me so I can write one.

also, running in bohr is strongly preferred since atomic units don't change with codata revisions.

qcengine/programs/orca.py
```diff
@@ -0,0 +1,297 @@
...
+            #     hf_type = "UHF"
+            #     dft_type = "UKS"
+
+            # Write energy call
+            energy_call = []
+
+            if input_model.model.method.upper() in self._post_hf_methods:  # post SCF case
+                energy_call.append(f"{{{hf_type}}}")
+                energy_call.append("")
+                energy_call.append(f"{{{input_model.model.method}}}")
```
loriab: would a post-scf ever get the coordinates or a basis set?

qcengine/programs/orca.py
```diff
@@ -0,0 +1,297 @@
...
+        #     if success:
+        #         tree = ET.ElementTree(ET.fromstring(output["outfiles"]["version.xml"]))
+        #         root = tree.getroot()
+        #         version_tree = root.find("molpro_uri:job/molpro_uri:platform/molpro_uri:version", name_space)
+        #         year = version_tree.attrib["major"]
+        #         minor = version_tree.attrib["minor"]
+        #         molpro_version = year + "." + minor
+        #         self.version_cache[which_prog] = safe_version(molpro_version)
+
+        return "4.2.1"
```
loriab: if orca doesn't do a `orca --version` or similar, often programs will print a version in their output file. try experimenting with a very minimal input that we can sneak the version out of the output header.

qcengine/programs/orca.py
muammar: Along with this PR there is another PR for QCElemental. I will create it soon. I will change to bohr instead. Let me research about it.

qcengine/programs/orca.py
muammar: I will create some inputs for post HF and update accordingly. I was just playing with HF and DFT.

qcengine/programs/orca.py
muammar: Yeah. I realized you do something like it for Molpro if I am not wrong. Let me come up with a hacky solution for it.

qcengine/programs/orca.py
```diff
@@ -0,0 +1,298 @@
+"""
+Calls the Orca executable.
+"""
+
+import string
+import cclib
```
dgasmith: `cclib` import needs to be internal to the functions since this will not be a global dependancy.

qcengine/programs/orca.py
```diff
@@ -0,0 +1,298 @@
...
+                energy_call.append("")
+                energy_call.append(f"{{{input_model.model.method}}}")
+            
+
+            elif input_model.model.method.upper() in self._dft_functionals:  # DFT case
+                input_file.append("! SP {} {}".format(input_model.model.method, input_model.model.basis))
+                input_file.append(xyz_block)
+
+            elif input_model.model.method.upper() in self._hf_methods:  # HF case
+                input_file.append("! SP {} {}".format(input_model.model.method, input_model.model.basis))
```
dgasmith: The HF and DFT cases seem to be identical? Might be best to combine them.

qcengine/programs/orca.py
```diff
@@ -0,0 +1,298 @@
...
+            "input_result": input_model.copy(deep=True),
+        }
+
+    def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") -> "AtomicResult":
+
+        data = cclib.io.ccread(io.StringIO(outfiles["stdout"]))
+
+        properties = {}
+        extras = {}
+        extras["ev_to_hartrees"] = 0.0367493
```
dgasmith: Is this from Orca itself? Otherwise best to use `qcelemental.constants.conversion_factor("ev", "hartree")`

qcengine/programs/orca.py
```diff
@@ -0,0 +1,298 @@
...
+
+        output_data["properties"] = properties
+        output_data["extras"].update(extras)
+        output_data["schema_name"] = "qcschema_output"
+        output_data["stdout"] = outfiles["stdout"]
+        output_data["success"] = True
+
+        return AtomicResult(**output_data)
+
+    def get_gradient(self, gradient_file):
```
dgasmith: Can this be a private function with `_get_gradient`?

muammar: Yes, no problem. Done.

yueyericardo:
Just for your information, I manually add this PR into my local [fork](https://github.com/yueyericardo/QCEngine/commit/8abc3380175f94d042d0ee7a35a8d13c74df5b4b). And orca is working correctly on my end.

yueyericardo:
Hi, I'm not familiar withthe QCArchive codebase and found difficulties when run orca with qcarchive. The problem is that `parsl` is good for multi-threading software, but it does not work well with software like orca that is compiled against OpenMPI.

I managed a workaround by adjusting the configuration as follows. From Parsl's perspective, this ensures:
A single task_command for each worker
56 mpi_ranks (equivalent to ntasks in SLURM).
Only 1 cores_per_worker in Parsl's view, translating to 1 cpus_per_task in SLURM.

```yaml
# manager.yaml
common:
 adapter: parsl
 tasks_per_worker: 1 
 cores_per_worker: 1  # this is a workaround for parsl because it determine cpus_per_task = cores_per_node / tasks_per_node, we need cpus_per_task=1 
 memory_per_worker: 192 
 max_workers: 20 
 scratch_directory: "$SCRATCH"

cluster:
 node_exclusivity: True
 scheduler: slurm
 scheduler_options: 
  - --ntasks-per-node=56  # this is because orca uses multi-processing instead of multi-threading
```

however, I would now need to mannually hack the orca.py to set the number of process
```python
            # parallel = "%pal nproc {} end".format(config.ncores)
            config_ncores = 56
            parallel = "%pal nproc {} end".format(config_ncores)
```
because the launch command`process_worker_pool.py  --max_workers=1 -a login1.frontera.tacc.utexas.edu -p 0 -c 1` set number of core equals 1 now......

Any guidance or improvements to fix this issue would be appreciated!
Thanks!

loriab (commented):
Thanks for pursuing this. Please go ahead and rebase so that CI can run.

PR #185, opened 2019-12-02
Entos: Added subcommands and more keywords

sjrl:
Added the HF subcommand to the entos harness and performed some general cleanup. 

- Removed soon to be deprecated options in entos
- Added more keywords for the available commands in the entos harness. Keyword names are copied from the https://www.entos.info/manual only applicable to the energy command. Will need more consideration for how to pass keywords to the gradient and hessian drivers.
- Added entos to the standard suite test for HF

### DIFF
--- a/qcengine/compute.py
+++ b/qcengine/compute.py
@@ -1,9 +1,12 @@
 """
 Integrates the computes together
 """
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, TYPE_CHECKING
 
-from pydantic.main import BaseModel
+if TYPE_CHECKING:
+    from pydantic.main import BaseModel
+    from qcelemental.models import AtomicResult
+
 from qcelemental.models import AtomicInput, FailedOperation, AtomicResult
 
 from .config import get_config
--- a/qcengine/compute.py
+++ b/qcengine/compute.py
@@ -3,7 +3,8 @@ Integrates the computes together
 """
 from typing import Any, Dict, Optional, Union
 
-from qcelemental.models import AtomicInput, FailedOperation
+from pydantic.main import BaseModel
+from qcelemental.models import AtomicInput, FailedOperation, AtomicResult
 
 from .config import get_config
 from .exceptions import InputError, RandomError
--- a/qcengine/programs/entos.py
+++ b/qcengine/programs/entos.py
@@ -77,7 +77,8 @@ class EntosHarness(ProgramHarness):
         "orbital_grad_threshold",
         "energy_threshold",
         "coulomb_method",
-        "exchange_method"}
+        "exchange_method",
+    }
     _dft_keywords_extra: Set[str] = _scf_keywords_extra.copy()
     _hf_keywords_extra: Set[str] = _scf_keywords_extra.copy()
     _xtb_keywords_extra: Set[str] = {}
@@ -86,15 +87,16 @@ class EntosHarness(ProgramHarness):
     _energy_commands: Dict[str, Any] = {
         "dft": _dft_keywords_extra,
         "hf": _hf_keywords_extra,
-        "xtb": _xtb_keywords_extra
+        "xtb": _xtb_keywords_extra,
     }
 
     class Config(ProgramHarness.Config):
         pass
 
     def found(self, raise_error: bool = False) -> bool:
-        return which("entos", return_bool=True, raise_error=raise_error,
-                     raise_msg="Please install via https://www.entos.info/")
+        return which(
+            "entos", return_bool=True, raise_error=raise_error, raise_msg="Please install via https://www.entos.info/"
+        )
 
     def get_version(self) -> str:
         self.found(raise_error=True)
@@ -134,14 +136,14 @@ class EntosHarness(ProgramHarness):
             return UnknownError(proc["stderr"])
 
     def execute(
-            self,
-            inputs: Dict[str, Any],
-            extra_infiles: Optional[Dict[str, str]] = None,
-            extra_outfiles: Optional[List[str]] = None,
-            extra_commands: Optional[List[str]] = None,
-            scratch_name: Optional[str] = None,
-            scratch_messy: bool = False,
-            timeout: Optional[int] = None,
+        self,
+        inputs: Dict[str, Any],
+        extra_infiles: Optional[Dict[str, str]] = None,
+        extra_outfiles: Optional[List[str]] = None,
+        extra_commands: Optional[List[str]] = None,
+        scratch_name: Optional[str] = None,
+        scratch_messy: bool = False,
+        timeout: Optional[int] = None,
     ) -> Tuple[bool, Dict[str, Any]]:
         """
         For option documentation go look at qcengine/util.execute
@@ -178,7 +180,7 @@ class EntosHarness(ProgramHarness):
         return exe_success, proc
 
     def build_input(
-            self, input_model: "AtomicInput", config: "JobConfig", template: Optional[str] = None
+        self, input_model: "AtomicInput", config: "JobConfig", template: Optional[str] = None
     ) -> Dict[str, Any]:
 
         # Write the geom xyz file with unit au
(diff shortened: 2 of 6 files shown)


qcengine/compute.py
```diff
@@ -3,7 +3,8 @@
 """
 from typing import Any, Dict, Optional, Union
 
-from qcelemental.models import AtomicInput, FailedOperation
+from pydantic.main import BaseModel
+from qcelemental.models import AtomicInput, FailedOperation, AtomicResult
```
sjrl: I can remove these (`BaseModel` and `AtomicResult`), I just added them in since PyCharm noticed they weren't defined when they are used to type declare the return type of `compute_procedure` and `compute` respectively

qcengine/programs/entos.py
```diff
@@ -73,18 +67,35 @@ class EntosHarness(ProgramHarness):
...
+        "energy_threshold",
+        "coulomb_method",
+        "exchange_method"}
+    _hf_keywords_extra: Set[str] = {
+        "ansatz",
+        "df",
+        "orbital_grad_threshold",
+        "energy_threshold",
+        "coulomb_method",
+        "exchange_method"}
```
mattwelborn: There is redundancy between `_{xtb,dft,hf}_keywords_extra` and particularly `hf`/`dft`. Consider creating an inheritance-like structure of sets and supersets.

qcengine/programs/entos.py
```diff
@@ -182,67 +194,69 @@ def build_input(
...
+                "hf": {
+                    "ao": input_model.model.basis,
+                    "charge": input_model.molecule.molecular_charge,
+                    "spin": float(input_model.molecule.molecular_multiplicity - 1),
+                },
                 "xtb": {
-                    "charge": input_model.molecule.molecular_charge
+                    "charge": input_model.molecule.molecular_charge,
                 }
             }
```
mattwelborn: Similar to above, there are DRY opportunities here.

qcengine/programs/entos.py
```diff
@@ -182,67 +194,69 @@ def build_input(
...
+                            **energy_keywords_extra
+                        },
                     },
-                    **print_results,
                 }
             else:
                 raise NotImplementedError(f"Driver {input_model.driver} not implemented for entos.")
 
             # Write input file
-            input_file = self.write_input_recursive(input_dict)
```
mattwelborn: Please add entos policy/version here.

qcengine/programs/entos.py
```diff
@@ -291,7 +305,10 @@ def write_input_recursive(self, d: Dict[str, Any]) -> List:
 
     def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") -> "AtomicResult":
 
-        dft_map = {"energy": "scf_total_energy", "n_iter": "scf_iterations"}
+        dft_map = {
+            "energy": "scf_total_energy",
+            "n_iter": "scf_iterations"
+        }
```
mattwelborn: Once again, DRY for `{dft,xtb,hf}_map`

qcengine/programs/tests/test_standard_suite_hf.py
```diff
@@ -37,6 +37,12 @@ def nh2():
     [
         pytest.param("cfour", "aug-pvdz", {"scf_conv": 12}, marks=testing.using_cfour),
         pytest.param("cfour", "aug-pvdz", {}, marks=testing.using_cfour),
+        pytest.param(
+            "entos",
+            "aug-cc-pVDZ",
+            {"coulomb_method": "direct_4idx", "exchange_method": "direct_4idx"},
+            marks=testing.using_entos
+        ),
```
mattwelborn: Awesome; didn't know entos actually supported "normal" integrals.

qcengine/compute.py
mattwelborn: Use the `if TYPE_CHECKING` pattern:
https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles

qcengine/programs/tests/test_standard_suite_hf.py
sjrl: Yup! It is a little slow. I did try to use a giant density fitting basis, but `aug-cc-pVTZ-JKFIT` didn't cut it and there is no `QZ` version.

qcengine/programs/entos.py
sjrl: Yup! Just got added below.

sjrl:
> Thanks for the contribution; we're getting closer to exposing the most useful functionality in entos.
> 
> Would you consider including generic keyword handling? @loriab I can't find the issue where this was discussed, but I think the agreed upon convention was to use double underscore. For example, one could pass a not-yet-implemented `dft` keyword into a `hessian` calculation by saying `hessian__dft__foo = bar`.

If we add generic keyword handling would it better to not have "implemented" keywords anymore and require that all keywords are passed via `hessian__dft__foo = bar`?

mattwelborn:
> If we add generic keyword handling would it better to not have "implemented" keywords anymore and require that all keywords are passed via `hessian__dft__foo = bar`?

I'm okay with having convenience keywords for energy methods to make it easier and cleaner to pass in commonly used options, as long as the clunky was also works.

sjrl:
Speaking of which, how many convenience keywords would you like for the energy_methods? It's fairly straightforward to add more keywords listed in the entos manual and I'd be happy to add more.

mattwelborn:
> Speaking of which, how many convenience keywords would you like for the energy_methods? It's fairly straightforward to add more keywords listed in the entos manual and I'd be happy to add more.

I guess add as many as you want to write tests for. I think you've hit the big ones, and I'm happy if you want to stop here.

sjrl:
> I guess add as many as you want to write tests for. I think you've hit the big ones, and I'm happy if you want to stop here.

Okay I'm happy to stop here for the time being. Should I go ahead and work on the general clunky way with the assumed double underscore convention? I'll follow the entos hierarchy such as `hessian__dft__foo = bar` and assume that the user understands how an native entos input works?

qcengine/testing.py
```diff
@@ -12,7 +12,7 @@
 import qcengine as qcng
 from qcelemental.util import which, which_import
 
-QCENGINE_RECORDS_COMMIT = "854d1fe"
+QCENGINE_RECORDS_COMMIT = "f81f0f2"
```
mattwelborn: a18e236 or 9f64c01

qcengine/programs/entos.py
```diff
@@ -124,14 +138,14 @@ def compute(self, input_data: "AtomicInput", config: "JobConfig") -> "AtomicResu
...
-        scratch_name: Optional[str] = None,
-        scratch_messy: bool = False,
-        timeout: Optional[int] = None,
+            self,
+            inputs: Dict[str, Any],
+            extra_infiles: Optional[Dict[str, str]] = None,
+            extra_outfiles: Optional[List[str]] = None,
+            extra_commands: Optional[List[str]] = None,
+            scratch_name: Optional[str] = None,
+            scratch_messy: bool = False,
```
dgasmith: Odd that black is changing this.

qcengine/programs/entos.py
```diff
@@ -291,34 +308,49 @@ def write_input_recursive(self, d: Dict[str, Any]) -> List:
...
             "density": "scf_density",
             "orbitals": "scf_orbitals",
             "fock": "fock",
         }
+        dft_map = {
+            **scf_map
+        }
 
-        xtb_map = {"energy": "scf_total_energy", "n_iter": "scf_iterations"}
+        hf_map = {
```
dgasmith: This is equivalent to `hf_map = scf_map.copy()`, seems cleaner.

qcengine/programs/entos.py
sjrl: Sorry I don't think black changed that. I believe that was the built in PEP8 enforcer from PyCharm. Should I undo the indent change?

dgasmith: Yes, we use `black` now to format everything. Looks like PyCharm can integrate: https://black.readthedocs.io/en/stable/editor_integration.html

sjrl: Thanks! Just reformatted using black.

qcengine/programs/entos.py
```diff
@@ -181,68 +194,59 @@ def build_input(
...
                     },
-                    **print_results,
                 }
             else:
                 raise NotImplementedError(f"Driver {input_model.driver} not implemented for entos.")
 
             # Write input file
-            input_file = self.write_input_recursive(input_dict)
+            input_file = [
+                f"entos_policy(version = '{self.get_version()}')",
```
sjrl: It looks like CI is failing when I added this line to the entos input file.  @mattwelborn do you think the entos_policy is worth keeping? If so I'll just add the `@using_entos` mark to the `test_entos_input_formatter`.

dgasmith:
Perfect, thanks! I think we just need to the QCEngineRecords hash change.

qcengine/programs/entos.py
sjrl: I added the @using_entos.

PR #187, opened 2019-12-04
Qchem: log parser

mattwelborn:
## Description
This PR adds a new output parser to the qchem harness for the case where only log files are available. This case arises often when ingesting old data; QCSCRATCH is rarely preserved. It depends on MolSSI/QCEngineRecords#23.

## Status
- [ ] Changelog updated
- [ ] Ready to go


qcengine/programs/qchem.py
```diff
@@ -259,3 +263,201 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+        if mobj:
+            provenance["wall_time"] = float(mobj.group(1))
+
+        mobj = re.search(r'Archival summary:\s*\n[0-9]+\\[0-9+]\\([\w\.\-]+)\\', outtext)
+        if mobj:
+            provenance["hostname"] = mobj.group(1)
+
+        properties = {}
+
+        mobj = re.search(r'\n\s*SCF\s+energy in the final basis set =\s+'+NUMBER+r'\s*\n', outtext)
```
dgasmith: Could we easily combine this with `parse_output`? Where `parse_output` would skip items that it doesn't have?

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,201 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+                    keywords[s[0].lower()] = s[1]
+                input["model"] = {}
+                input["model"]["method"] = keywords.pop('method').lower()
+                input["model"]["basis"] = keywords.pop('basis').lower()
+                if 'jobtype' in keywords:
+                    jobtype = keywords.pop('jobtype')
+                else:
+                    jobtype = 'sp'
+                _jobtype_to_driver = {'sp': 'energy', 'force': 'gradient', 'freq': 'hessian'}
+                input["driver"] = _jobtype_to_driver[jobtype]
```
dgasmith: Do not names items `input` to not overload baes methods.

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,201 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
         output_data["success"] = True
 
         return AtomicResult(**{**input_model.dict(), **output_data})
+
+    @staticmethod
+    def parse_logfile(outtext: str) -> AtomicResult:
+        """Parses a log file. Should only be used when QCSCRATCH is not available."""
+        warnings.warn("parse_logfile should only be used when the QCSCRATCH directory is not available.")
+
+        NUMBER = "((?:[-+]?\\d*\\.\\d+(?:[DdEe][-+]?\\d+)?)|(?:[-+]?\\d+\\.\\d*(?:[DdEe][-+]?\\d+)?))"
```
dgasmith: We need a common RE expression spot at some point.

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,201 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+                        i += 1
+
+                ret = np.zeros((maxrow, maxcol))
+                for row in mdict:
+                    for col in mdict[row]:
+                        ret[row, col] = mdict[row][col]
+                return ret
+
+            if method in _mp2_methods:
+                mobj = re.search(r'\n\s+Full Analytical Gradient of MP2 Energy \(in au.\)\s*\n'
```
dgasmith: Our re skills are increasing at a somewhat depressing rate.

mattwelborn: image

qcengine/programs/qchem.py
mattwelborn: Do you mean items that it does have? That is, we first parse the QCSCRATCH. Then we parse the logfile, but ignore any fields that were determined from QCSCRATCH?

qcengine/programs/qchem.py
mattwelborn: Okay, I made one.

qcengine/programs/qchem.py
dgasmith: `parse_logfile` would have three steps:
1) Parse the output for enough information to generate a input model.
2) Pass the output and input model through the canonical output parser.
3) Parse additional information that the canonical output parser will not pick up due to a lack of extra files.

mattwelborn: Would this be the strategy for parsing all output, or only when we ingest data?

dgasmith: Only when we ingest data, it should come with a warning this method is less accurate.

qcengine/programs/cfour/harvester.py
```diff
@@ -63,7 +63,7 @@ def harvest_outfile_pass(outtext):
     #          vcc/ecc
 
     # fmt: off
-    NUMBER = "((?:[-+]?\\d*\\.\\d+(?:[DdEe][-+]?\\d+)?)|(?:[-+]?\\d+\\.\\d*(?:[DdEe][-+]?\\d+)?))"
+    NUMBER = regex.NUMBER
```
loriab: fyi that qcel has a `NUMBER` https://github.com/MolSSI/QCElemental/blob/16b2de82eb603c8d868a9c5e27a696c769f1f6ae/qcelemental/molparse/regex.py#L18-L22

qcengine/programs/qchem.py
mattwelborn: done!

qcengine/programs/cfour/harvester.py
mattwelborn: Thanks for pointing this out. It requires `re.VERBOSE` which will break some of my existing regexes. Is it possible to convert a VERBOSE `re` to a standard one?

loriab: maybe could incorporate into the string itself with `(?aiLmsux)` from https://docs.python.org/3.8/library/re.html. I just add the `re.VERBOSE` or `re.X`

mattwelborn: I can certainly turn on verbose mode inline with `(?x)`, but I don't know how to turn it off.

mattwelborn: Nevermind, there is a `(?x:<regex>)` syntax that solves this issue.

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,371 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+        mobj = re.search(r"\n\s*SCF\s+energy in the final basis set =\s+" + NUMBER + r"\s*\n", outtext)
+        if mobj and qcscr_result["properties"].get("scf_total_energy", None) is None:
+            qcscr_result["properties"]["scf_total_energy"] = float(mobj.group(1))
+
+        mobj = re.search(r"\n\s*Nuclear Repulsion Energy =\s+" + NUMBER + r"\s+hartrees\s*\n", outtext)
+        if mobj and qcscr_result["properties"].get("nuclear_repulsion_energy", None) is None:
+            qcscr_result["properties"]["nuclear_repulsion_energy"] = float(mobj.group(1))
+
+        mobj = re.search(r"\n\s*There are\s+(\d+) alpha and\s+(\d+) beta electrons\s*\n", outtext)
+        if mobj and qcscr_result["properties"].get("calcinfo_nalpha", None) is None:
```
dgasmith: We should move this to the `parse_output`?

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,371 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+            qcscr_result["properties"]["calcinfo_nalpha"] = int(mobj.group(1))
+        if mobj and qcscr_result["properties"].get("calcinfo_nbeta", None) is None:
+            qcscr_result["properties"]["calcinfo_nbeta"] = int(mobj.group(2))
+
+        mobj = re.search(r"\n\s*There are\s+\d+ shells and\s+(\d+) basis functions\s*\n", outtext)
```
dgasmith: Same here, anything we don't currently parse in `parse_output` should be moved there I think.

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,371 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+        _mp2_methods = {"rimp2"}
+
+        method = input_dict["model"]["method"].lower()
+        if qcscr_result["properties"].get("return_energy", None) is None:
+            if method in _mp2_methods:
+                qcscr_result["properties"]["return_energy"] = qcscr_result["properties"]["mp2_total_energy"]
+            elif method in _scf_methods:
+                qcscr_result["properties"]["return_energy"] = qcscr_result["properties"]["scf_total_energy"]
+            else:
+                raise NotImplementedError(f"Method {method} not supported for energy driver.")
```
dgasmith: Not supported by the log file parser?

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,371 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+            "pbe-qidh",
+            "pbe0-2",
+            "pbe0-dh",
+            "ptpss-d3",
+            "dsd-pbeb95-d3",
+            "pwpb95-d3",
+        }
+        _mp2_methods = {"rimp2"}
+
+        method = input_dict["model"]["method"].lower()
```
dgasmith: Rough strategy here to have every method listed out, unsure if there is a better way. @loriab?

qcengine/programs/qchem.py
```diff
@@ -259,3 +263,371 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+        if (
+            mobj_aaaa
+            and mobj_bbbb
+            and qcscr_result["properties"].get("mp2_opposite_spin_correlation_energy", None) is None
+        ):
+            qcscr_result["properties"]["mp2_opposite_spin_correlation_energy"] = float(mobj_aabb.group(1)) + float(
+                mobj_bbaa.group(1)
+            )
+
+        _scf_methods = {
```
dgasmith: At the very least these might be better off as class methods.

qcengine/programs/qchem.py
```diff
@@ -249,13 +255,390 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+            r"\n\s+Dipole Moment \(Debye\)\s*\n\s+X\s+" + NUMBER + r"\s+Y\s+" + NUMBER + r"\s+Z\s+" + NUMBER + r"\s*\n",
+            outtext,
+        )
+        if mobj:
+            cf = constants.conversion_factor("debye", "e * bohr")
+            properties["scf_dipole_moment"] = [float(mobj.group(i)) * cf for i in range(1, 4)]
+
+        return properties, provenance
+
+    def parse_logfile(self, outfiles: Dict[str, str]) -> AtomicResult:
```
loriab: might be good to throw if qchem link0 encountered (or whatever lets them run multiple jobs per input if link0 is gaussian syntax)

qcengine/programs/qchem.py
mattwelborn: I'm hesitant to use this approach because it requires knowledge of method hierarchy. This is obvious when different methods have different costs, but will have a lot of edge cases such as SCS-MP2, solvation, excited states, etc.

The methods list comes from Q-Chem's manual, so I see it as being on them if one of these does not work.

mattwelborn:
> looking good!

Thanks for the careful review!

qcengine/programs/qchem.py
```diff
@@ -249,13 +255,390 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+                else:
+                    jobtype = "sp"
+                _jobtype_to_driver = {"sp": "energy", "force": "gradient", "freq": "hessian"}
+                input_dict["driver"] = _jobtype_to_driver[jobtype]
+                for key in keywords:
+                    if keywords[key] == "false":
+                        keywords[key] = False
+                    if keywords[key] == "true":
+                        keywords[key] = True
+                input_dict["keywords"] = keywords
```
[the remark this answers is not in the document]
mattwelborn: I think this goes the opposite direction, but the point is taken.

qcengine/programs/qchem.py
```diff
@@ -249,13 +255,390 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+        if mobj:
+            inputtext = mobj.group(1)
+
+            rem_match = re.search(r"\$rem\s*\n([^\$]+)\n\s*\$end", inputtext, re.DOTALL | re.IGNORECASE)
+            if rem_match:
+                rem_text = rem_match.group(1)
+                lines = rem_text.split("\n")
+                keywords = {}
+                for line in lines:
+                    s = line.split()
```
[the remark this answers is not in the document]
mattwelborn: strip is not needed because `split()` trims leading and trailing whitespace; comment filtering added

qcengine/programs/qchem.py
```diff
@@ -249,13 +255,390 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+                    keywords[s[0].lower()] = s[1]
+                input_dict["model"] = {}
+                input_dict["model"]["method"] = keywords.pop("method").lower()
+                input_dict["model"]["basis"] = keywords.pop("basis").lower()
+                if "jobtype" in keywords:
+                    jobtype = keywords.pop("jobtype")
+                else:
+                    jobtype = "sp"
+                _jobtype_to_driver = {"sp": "energy", "force": "gradient", "freq": "hessian"}
+                input_dict["driver"] = _jobtype_to_driver[jobtype]
```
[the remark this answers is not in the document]
mattwelborn: Rejecting optimizations for now

qcengine/programs/nwchem/germinate.py
```diff
@@ -3,15 +3,89 @@
...
+    "mpw1b95",
+    "mpwb1k",
+    "pw6b95",
+    "pwb6k",
+    "m05",
+    "m05-2x",
+    "vs98",
+    "m06",
+    "m06-hf",
+    "m06-L",
```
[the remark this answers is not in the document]
mattwelborn: nwchem changes are a result of black

dgasmith:
I think this should be ready to go after QCER commit update?

mattwelborn:
> I think this should be ready to go after QCER commit update?

Ready to go as long as you're satisfied w.r.t. the list of DFT functionals.

qcengine/programs/qchem.py
loriab: I completely agree that if any method doesn't run to "Have a nice day", it's on them. It's whether the final energy result for all is to be found at "SCF energy in the final basis set" that makes me nervous for methods as different as b3lyp, pbeh-3c, double hybrids, vv10's, etc.

mattwelborn: Fair point that's a bug; it should be "Total energy in the final basis set". (If Q-Chem fails to put the correct result for an SCF method in "Total energy in the final basis set", then I'm willing to say that's on them.) 

This does raise a question about whether the `scf_total_energy` should be  "Total energy in the final basis set" or  "SCF energy in the final basis set"...

dgasmith: In terms of schema?

mattwelborn: Yes.

dgasmith: Can you explain a bit more here? Why are basis sets entering this conversation when a basis is pinned by the input?

loriab: "basis" is just in the line that's scraped into `scf_total_energy`, which in turn for this whole list becomes the `return_energy`. I'm gloomier than Matt that all this list of methods is so consistently scraped with one line. unless I've lost track of the real issue ...

dgasmith:
Happy here. @loriab do any comments on the functional list? Not super happy there either, but unsure of a better way.

PR #188, opened 2019-12-06
Read from both stderr and stdout to prevent buffer deadlock

WardLT:
## Description
Fixes problems with the stderr buffer filling and leading to QCEngine stalling.

Changes `popen()` to pull from both stderr and stdout while the subprocess is running (using multithreading) and modified `execute()` so that it does not have its own "make sure buffer doesn't fill" code

## Todos

None

## Status
- [x] Make sure test suite passes (only did a few of them locally)
- [ ] Changelog updated
- [ ] Ready to go


WardLT:
My it never hang and prompt my interference again, indeed 🤞 

I was thinking about adding in an option to pass the stderr and stdout read by the subprocess forward to the systems' standard out. The plan would be to have threads that read from the buffer have an option to also write the out the buffer to the system's stdout. Should I go ahead and see if I can figure that out?

loriab:
Thanks for the sysio forwarding! I look forward to trying it out.

dgasmith:
@WardLT Can you fix merge conflicts then ready to go.

WardLT:
I've fixed the merge conflicts

PR #189, opened 2019-12-06
MPI task support

WardLT:
## Description
Adds ability for QCEngine to invoke node-parallel tasks.

Implementing this required adding more information to the task configuration object and some utility functions for making an MPI configuration given a task configuration.

I also implemented support for NWChem, which seems to be working. 🤞 

## Todos
- [x] Add ability to get the CPU info of compute nodes

## Status
- [ ] Changelog updated
- [ ] Ready to go


qcengine/config.py
```diff
@@ -27,6 +27,8 @@ def get_global(key: Optional[str] = None) -> Union[str, Dict[str, Any]]:
     import cpuinfo
     import psutil
 
+    # TODO (wardlt): Implement a means of getting CPU information from compute nodes on clusters
```
dgasmith: Can you describe what is missing here? `cpuinfo` pulls quite a bit of information.

qcengine/config.py
```diff
@@ -72,7 +74,33 @@ class NodeDescriptor(pydantic.BaseModel):
...
+        Information about the task configuration will be added to the command by use of
+        Python's string formatting. The configuration will be supplied as the following variables:
+        
+            {nnodes} - Number of nodes
+            {ranks_per_node} - Number of MPI ranks per node
+            {total_ranks} - Total number of MPI ranks
+        """
+    )
+
+    def __init__(self, **data: Dict[str, Any]):
```
dgasmith: Any reason to remove ` -> "BaseModel"` here?

qcengine/config.py
```diff
@@ -82,13 +110,16 @@ class Config:
 
 
 class TaskConfig(pydantic.BaseModel):
+    """Description of the configuration used to launch a task."""
 
     # Specifications
-    ncores: int  # Number of ncores per task
+    ncores: int  # Number of cores per node
```
dgasmith: This is truly core per task. We often run where ncores < ncores per node.

qcengine/tests/test_config.py
```diff
@@ -155,8 +164,18 @@ def test_config_local_njob_ncore_plus_memory(opt_state_basic):
...
+        qcng.config.get_config(hostname="something", local_options={"nnodes": 10})
+    assert 'https://qcengine.readthedocs.io/en/stable/environment.html' in str(exc.value)
+
+    # Test with an MPI run command
+    config = qcng.config.get_config(hostname="something",
+                                    local_options={"nnodes": 4,
+                                                   "mpirun_command": "mpirun -n {total_ranks} -N {ranks_per_node}"})
+    assert config.use_mpirun
+    assert config.mpirun_command.startswith('mpirun')
+    assert create_mpi_invocation('hello_world', config) == ['mpirun', '-n', '20', '-N', '5', 'hello_world']
```
dgasmith: This is a really nice way of handling this!

@taylor-a-barnes What do you think of this kind of automated api config?

qcengine/config.py
```diff
@@ -72,7 +74,33 @@ class NodeDescriptor(pydantic.BaseModel):
...
+    mpirun_command: Optional[str] = pydantic.Field(
+        None, description="""Invocation for launching node-parallel tasks with MPI
+        
+        The invocation need not specify the number of nodes, tasks, or cores per node.
+        Information about the task configuration will be added to the command by use of
+        Python's string formatting. The configuration will be supplied as the following variables:
+        
+            {nnodes} - Number of nodes
+            {ranks_per_node} - Number of MPI ranks per node
+            {total_ranks} - Total number of MPI ranks
```
dgasmith: An example here would be great, I didn't see at first how of this slotted together.

dgasmith: I also wonder if we need a validator that checks for each of parameters in the string so that we can validate it.

qcengine/util.py
```diff
@@ -220,22 +266,52 @@ def popen(
...
+
+    # Ready the output
     ret = {"proc": subprocess.Popen(args, **popen_kwargs)}
+
+    # Spawn threads that will read from the stderr/stdout
+    #  The PIPE uses a buffer with finite capacity. The underlying
+    #  process will stall if it is unable to write to the buffer
+    #  because the buffer is full. These threads continuously read
+    #  from the buffers to ensure that they do not fill.
+    #
```
dgasmith: How much is that output BTW? Curious how you hit this cap, we often dumps 20-30MB log files here and haven't had an issue yet.

qcengine/config.py
WardLT: The amount of information is fine. However, the cpuinfo is pulled from batch node and not the compute node if QCEngine is configured for MPI tasks. This means the provenance information is not for the CPU that actually ran the calculation

qcengine/config.py
WardLT: Yes, the `__init__` function actually returns `None`

qcengine/config.py
WardLT: Sure. I'll put in an example and a validator is a good idea. However, some MPI launch systems will not use all 3 of these (e.g., the `aprun` used by Cray does not use `nnodes`) so I might make sure either `nnodes` or `total_ranks` is used along with `total_ranks`.

qcengine/util.py
WardLT: The problem was with the `stderr`, which had maybe a few KBs of data. The MPI tasks I was using print a stack trace from each MPI rank to `stderr`, which probably explains why I was seeing this problem for the first time. This change (also in #188) adds the same kind of buffer unpacking that was already used for stdout.

qcengine/config.py
dgasmith: Ah, I see and makes sense. Can you add a further comment here to explain this?

qcengine/config.py
dgasmith:
```
>>> re.findall("{\w+}", "hello {this} is a {scan} of {paren} words")
['{this}', '{scan}', '{paren}']
```
All together `set(re.findall("{\w+}", ...)) < {"nodes", "ranks", ...}`?

dgasmith: @mattwelborn @loriab I see now that everything in life can be solved via `re`.

WardLT: Nice! That's some good regex magic! I I opted for writing a more explicit test using `not in`, but going to have to remember this one :)

qcengine/config.py
WardLT: Ah, I'll change it back and add some notes that this means "number of cores per node" for nnodes>1 because we only use 1 task per node.

qcengine/config.py
WardLT: Sure

qcengine/tests/test_config.py
taylor-a-barnes: @dgasmith @WardLT 

Generally speaking, this should work fine.  One thing you should probably plan for is how this api will handle issues related to MPI task pinning, especially in the context of hybrid OMP-MPI calculations.  Some nice examples of these complications are described [at NERSC](https://docs.nersc.gov/jobs/affinity/).  A few specific points to consider: 

1. How will you handle task affinity with respect to logical cores (see, for example, the srun "-cpus-per-task" and "-cpu-bind" options)?  Is it worthwhile to have a logical_cores_per_rank option?
2. Similarly, is it worthwhile to have a ranks_per_socket option for handling socket affinity?
3. How does the user specify a job that involves, for example, 16 MPI ranks with 4 OMP threads per rank? Can the api derive the relevant information from ncores and total_ranks?

I'll also point out that it is generally a best practice to use "mpiexec" rather than "mpirun"; I suggest that you make this change wherever possible.

WardLT: All good questions, @taylor-a-barnes !

My current approach to affinity related issues and handling hybrid parallelism is to allow the users to hard-code them in the `mpirun_command` (e.g., adding a `-cpu-per-task` flag, or a `-depth` ). As I don't have a good grasp of MPI launch systems, I kept the configuration simple.

That said, it would be wonderful to be able to set the number of logical cores per node. What do you think about a `ranks_per_node` configuration variable for `TaskConfig`? That could then be used to compute a `cores_per_rank` variable we can add to the mpirun command.

Also, I'll go ahead and rename things to mpiexec from mpirun

WardLT: @taylor-a-barnes , I've pushed a commit that adds the ability to set the number of "cores/threads per rank"

taylor-a-barnes: @WardLT Great! This tends to be one of the more overlooked and thus error-prone elements of launching MPI jobs, so I think users will find this helps them avoid a fair number of mistakes.

PR #190, opened 2019-12-08
Psi4: Fixes issue with HF3c execution

dgasmith:
## Status
- [ ] Changelog updated
- [x] Ready to go


qcengine/programs/tests/test_entos.py
```diff
@@ -3,7 +3,7 @@
 import qcelemental as qcel
 import qcengine as qcng
 from qcelemental.testing import compare_recursive
-from qcengine.testing import qcengine_records, using_entos
+from qcengine.testing import qcengine_records, using
```
loriab: can we standardize btwn `using` and `testing.using`? I slightly favor former.

after going through the PR, it seems former is indeed favored. any reason it can't be universal? there's only a couple files with latter.

qcengine/testing.py
```diff
@@ -140,59 +140,45 @@ def get_job(self):
...
+
+_using_cache = {}
+
+
+def using(program):
+
+    if program not in _using_cache:
+        import_message = f"Not detecting module {program}. Install package if necessary to enable tests."
+        skip = pytest.mark.skipif(has_program(program) is False, reason=import_message)
+        _using_cache[program] = skip
```
loriab: nice consolidation

qcengine/programs/psi4.py
```diff
@@ -74,6 +74,10 @@ def compute(self, input_model: "AtomicInput", config: "TaskConfig") -> "AtomicRe
         error_message = None
         compute_success = False
 
+        # Basis must not be None for HF3c
+        old_basis = input_model.model.basis
+        input_model.model.__dict__["basis"] = old_basis or ""
```
loriab: for my edification, what's the problem with `(none)` exactly? qcf?

qcengine/programs/tests/test_entos.py
dgasmith: I wanted to open an issue on this, but overall I think we need to standardize on `using` so that this strategy works for fixtures. I can spin out an issue to standardize this.

qcengine/programs/psi4.py
dgasmith: Psi4, the current JSON drive always attempts to set the basis. Psi4 options can only be strings.

qcengine/programs/tests/test_entos.py
loriab: sounds good. i'll standardize in p4 where we're borrowing using_* from qcng. though now that I think about it, that reform may be in the ddd branch.

PR #199, opened 2019-12-13
fix nwchem

loriab:
## Description
Sorry, I approved faulty code. Actually, all NWChem is broken with "NameError: name 'cc' is not defined". This fixes.

## Changelog description
- Fix typo breaking NWChem

## Status

- [ ] Code base linted
- [x] Ready to go


loriab:
Yes. I should learn how to generate qcer entries.

PR #202, opened 2019-12-16
Entos wavefunction properties

sjrl:
## Description
Support for wavefunction properties has been added for the entos Harness. One thing of note is that the basis set information does not exactly match the one found on basis set exchange for the same basis (e.g. def2-svp for Oxygen). I believe this is due to entos returning the normalized primitives in its output. 
- [x] No based on QCER discussion. ~~Should the normalized primitive information be included in the `extras` field?~~
- [x] Do quantities like the Fock and density matrices need to follow CCA spherical/cartesian basis ordering (for p functions and above)? 

## Changelog description
Support for wavefunction properties has been added for the entos Harness.

## Status
- [x] Code base linted
- [ ] Ready to go


qcengine/programs/entos.py
```diff
@@ -90,6 +121,27 @@ class EntosHarness(ProgramHarness):
...
+    # This map order converts entos ordering to CCA ordering
+    # Entos spherical basis ordering for each angular momentum
+    # S -->  0
+    # P --> +1,  0, -1
+    # D --> +2, +1,  0, -1, -2
+    # F --> +3, +2, +1,  0, -1, -2, -3
+    # G --> +4, +3, +2, +1,  0, -1, -2, -3, -4
+    # H --> +5, +4, +3, +2, +1,  0, -1, -2, -3, -4, -5
+    # I --> +6, +5, +4, +3, +2, +1,  0, -1, -2, -3, -4, -5, -6
+    _entos_to_cca_ao_order: Dict[str, Dict[int, List[int]]] = {
```
dgasmith: It this is just a reversal, maybe add an automatically generating one for arbitrary angular momentum? Granted unlikely to go past 6, but best not to artificially cap.

qcengine/programs/entos.py
```diff
@@ -296,20 +348,33 @@ def write_input_recursive(self, d: Dict[str, Any]) -> List:
...
+            "restricted": {
+                "orbitals": "scf_orbitals_a",
+                "density": "scf_density_a",
+                "fock": "scf_fock_a",
+                "eigenvalues": "scf_eigenvalues_a",
+                "occupations": "scf_occupations_a",
+            },
+            "unrestricted": {
+                "orbitals_alpha": "scf_orbitals_a",
+                "orbitals_beta": "scf_orbitals_a",
```
dgasmith: `orbitals_b`

qcengine/programs/entos.py
```diff
@@ -336,56 +398,134 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+                )
+
+            # Now parse wavefunction information
+            n_channels = entos_results["n_channels"]
+            if n_channels == 1:
+                wavefunction["restricted"] = True
+                for key in wavefunction_map["restricted"].keys():
+                    if key in entos_results:
+                        wavefunction[wavefunction_map["restricted"][key]] = entos_results[key]
+                        # TODO Uncomment if CCA ordering should be used
```
dgasmith: CCA ordering is always used, can you elaborate on this comment?

qcengine/programs/entos.py
sjrl: Yes this is just a reversal. And sounds good!

qcengine/programs/entos.py
sjrl: When I first made this PR, I hadn't reordered the AO functions to follow CCA ordering. So I've been in the process of writing the functions to perform this reordering. I'm happy to put the functions I've just written in a general place or do such general functions already exist in QCEngine?

sjrl:
Also another question. Should `scf_orbitals_*` be stored as column major or row major?

dgasmith:
Orbitals are (nao, nmo).

qcengine/programs/entos.py
sjrl: I am now using CCA ordering.

dgasmith:
@sjrl One question for you, can you think of a small test that can correctly check for orbital order?

Something like water makes even `p` hard to test, we should think of something like CHFClBr, but I think this starts breaking down at `d/f` orbitals and Br is pretty big. We do not need to check very high in angular momentum, but probably through `d` is wise.

sjrl:
To clarify, do we want to check the validity of the functions `reorder_ao_ indices_col()` and `reorder_ao_ indices_2d()`? Or the `_entos_to_cca_ao_order` map? 

For the former, a loose check I've been performing is calculating the SCF density from the transformed `scf_orbitals` to see if it matches the transformed `scf_density`. This at least tells me everything is consistent, but probably doesn't check whether the `reorder_ao_*` functions are doing exactly what we expect.

For the `_entos_to_cca_ao_order` I'm not sure how to check this short of loading the transformed wavefunction quantities into another program.

Maybe the best way to handle both of these questions is by adding another standard suite for wavefunction properties and perhaps comparing to the output from Psi4?

dgasmith:
Yea, I think we need to compare in two programs to be sure. We can project orbitals to a grid to check it that way in the extreme, but trying to make sure `dxx` is `dxx` in both programs is the key.

I am trying to figure out a molecule that has low enough symmetry that checking a `d` function is the same between two programs. We can handle the scaling factors and normalization, but the main item is trying to make sure the cartesian and spherical are correct.

dgasmith:
I am ok merging this soon, but we do need to think of a test to validate that we order items correctly. I think Hypofluorous acid is probably the simplest molecule that we can validate against `p` functions and perhaps some of `d` as well. I will make an issue of this, but does this kind of testing make sense?

sjrl:
Yes I agree this kind of testing makes sense. Hypofluorous sounds like it would be a good candidate if we want to test up to `d` functions. For `p` functions we could probably orient `H2` (in a `DZ` basis) on the different axes to check for `px`, `py`, and `pz`. What do you think?

Also before this merged I'll need to 1) push an update to QCEngineRecords to update `scf_orbitals_a` since they were shaped wrongly before and 2) I would like to put the `reorder_ao_indices_*` functions in a more general place than the entos program harness. Any suggestions on this front?

dgasmith:
2) There is a `/program/util` folder where this would find a good home I think.

sjrl:
Once PR https://github.com/MolSSI/QCEngineRecords/pull/28 is merged for QCER then Travis should pass.

PR #204, opened 2019-12-18
NWCHEM:  added keyword dft to xc_functionals list.  If method == "dft"

ahurta92:
## Description
NWChem: added "dft" to list of exchange functionals.  If method == "dft" let exchange functional be handled by keywords.  Works with mixed functionals like "xc becke88 lyp".  This doesn't deal with cases like 'dft__xc': 'xc vwn_1_rpa 0.19 lyp 0.81 HFexch 0.20  slater 0.80 becke88 nonlocal 0.72'.

## Changelog description
NWChem: added "dft" to list of exchange functionals and let xc option be handled by keywords
## Status

- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcengine/programs/nwchem/germinate.py
+++ b/qcengine/programs/nwchem/germinate.py
@@ -175,12 +175,14 @@ def muster_modelchem(method: str, derint: int, use_tce: bool) -> Tuple[str, Dict
         )
 
     elif method in _xc_functionals:
-        if method != "dft":
-            opts["dft__xc"] = method
+        opts["dft__xc"] = method
         if use_tce:
             mdccmd = f"task tce {runtyp}\n\n"
             opts["tce__"] = "dft"
         else:
+            mdccmd = f"task dft {runtyp}\n\n"
+
+    elif method != "dft":
             mdccmd = f"task dft {runtyp}\n\n"
 
     else:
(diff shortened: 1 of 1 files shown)


qcengine/programs/nwchem/germinate.py
```diff
@@ -174,7 +175,8 @@ def muster_modelchem(method: str, derint: int, use_tce: bool) -> Tuple[str, Dict
         )
 
     elif method in _xc_functionals:
-        opts["dft__xc"] = method
+        if method != "dft":
+            opts["dft__xc"] = method
```
dgasmith: A slight comment here, we may want to consider reordering this so that the `for` looks more like the following:
```python
elif method in _xc_functionals:
...
elif method == "dft":
....
```

This keeps the `_xc_functionals` list "pure" which may make a difference down the road. There is some duplicate `use_tce` commands however, but may be worth it.

qcengine/programs/nwchem/germinate.py
```diff
@@ -181,6 +181,9 @@ def muster_modelchem(method: str, derint: int, use_tce: bool) -> Tuple[str, Dict
         else:
             mdccmd = f"task dft {runtyp}\n\n"
 
+    elif method == "dft":
```
dgasmith: Do we need the same `use_tce` part as above?

PR #205, opened 2019-12-30
Higher Precision Gradients from NWChem

WardLT:
## Description
NWChem only prints gradients to an accuracy of 6 decimal places to stdout. I explored getting more accurate gradients from NWChem as a hope for allowing NWChem to work better with geoMETRIC. Reading gradients with higher accuracies did not seem to be the (only) issue with NWChem+geometric, but here is the result of that line of testing!

The full-accuracy versions of gradients are stored in checkpoint file of NWChem. The new code accesses and prints them using a Python script that runs during the NWChem execution.

See discussion here:
http://www.nwchem-sw.org/index.php/Special:AWCforum/st/id2155

I also implemented Hessian support while I was at it

## Changelog description
Improved gradient output and added Hessian support for NWChem

## Status

- [] Code base linted
- [ ] Ready to go


loriab:
fyi @vivacebelles

WardLT:
To add some more detail, I just realized that NWChem rotates the coordinate system of the molecule. Anyone have experience with rotating gradients and Hessians for other codes?

loriab:
Yup, I do it for Cfour. There's [alignment machinery in qcel](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molutil/align.py). Form an `AlignmentMill`, then there's methods for grad and hess (https://github.com/MolSSI/QCElemental/blob/52c81715d0975a444d419f85e2e681323755f566/qcelemental/models/align.py#L109-L141). Inputs are the internal geom and desired geom orientations, https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/cfour/harvester.py#L649-L653. Let me know if anything isn't clear.

WardLT:
Thanks, @loriab ! Rotating the gradients did the trick for getting NWC and geoMETRIC to work together!

qcengine/programs/nwchem/harvester.py
```diff
@@ -729,6 +748,17 @@ def harvest(in_mol: Molecule, nwout: str, **kwargs) -> Tuple[PreservingDict, Non
     # Parse the NWChem output
     outPsivar, outMol, outGrad, version, error = harvest_output(nwout)
 
+    # If available, read higher-accuracy gradients
+    #  These were output using a Python Task in NWChem to read them out of the database
+    if outfiles.get("nwchem.grad") is not None:
+        logger.debug("Reading higher-accuracy gradients")
+        outGrad = json.loads(outfiles.get("nwchem.grad"))
```
dgasmith: `out_grad` we try to stick to PEP8 variable names.

qcengine/programs/nwchem/runner.py
```diff
@@ -147,8 +148,31 @@ def build_input(
...
+  if ga_nodeid() == 0:
+      import json
+      grad = rtdb_get('{theory}:gradient')
+      with open('nwchem.grad', 'w') as fp:
+          json.dump(grad, fp)
+end
+
+task python
+            """
+            nwchemrec["infiles"]["nwchem.nw"] += pycmd
```
dgasmith: Could we do something similar with the Hessian?

qcengine/programs/tests/test_nwchem.py
```diff
@@ -43,3 +44,45 @@ def test_b3lyp(nh2):
...
+
+@using("nwchem")
+def test_hess(nh2):
+    resi = {"molecule": nh2, "driver": "hessian", "model": {"method": "b3lyp", "basis": "3-21g"}}
+    res = qcng.compute(resi, "nwchem", raise_error=True, return_dict=True)
+    assert compare_values(-3.5980754370e-02, res["return_result"][0], atol=1e-3)
+
+
+@using("nwchem")
+def test_gradient(nh2):
```
dgasmith: Can we add a Hessian test too?

qcengine/programs/nwchem/runner.py
WardLT: We don't need to. The Hessian is already printed to a file in higher precision.

qcengine/programs/tests/test_nwchem.py
WardLT: Is `test_hess` insufficient?

qcengine/programs/nwchem/harvester.py
WardLT: Sure, I'll make the changes

qcengine/programs/tests/test_nwchem.py
dgasmith: Ah, I am blind.

qcengine/programs/tests/test_nwchem.py
```diff
@@ -43,3 +44,45 @@ def test_b3lyp(nh2):
     assert res["properties"]["calcinfo_natom"] == 3
     assert res["properties"]["calcinfo_nalpha"] == 5
     assert res["properties"]["calcinfo_nbasis"] == 13
+
+
+@using("nwchem")
+def test_hess(nh2):
+    resi = {"molecule": nh2, "driver": "hessian", "model": {"method": "b3lyp", "basis": "3-21g"}}
+    res = qcng.compute(resi, "nwchem", raise_error=True, return_dict=True)
+    assert compare_values(-3.5980754370e-02, res["return_result"][0], atol=1e-3)
```
dgasmith: Unsure if this should work a `return_result` of a Hessian computation should be a 2-d array. I would also try to test more values than just a diagonal element to check the rotation as well.

qcengine/programs/nwchem/runner.py
dgasmith: Oh neat, can we add a comment that this isn't needed for the Hessian?

qcengine/programs/nwchem/harvester.py
```diff
@@ -660,12 +662,28 @@ def harvest_outfile_pass(outtext):
...
+    hess_conv = hess.replace("D", "E")
+
+    # Parse all of the float values
+    hess_triu = [float(x) for x in hess_conv.strip().splitlines()]
+
+    # The value in the Hessian matrix is the upper triangle
+    #  We first convert it to a 2D array
+    n = int(np.sqrt(8 * len(hess_triu) + 1) - 1) // 2  # Size of the 2D matrix
+    hess_arr = np.zeros((n, n))
+    hess_arr[np.triu_indices(n)] = hess_triu
```
dgasmith: You can do `hess_arr[np.tril_indices(n, -1)] = hess_triu`

dgasmith: Ah whoops this is actually

```python
idx = np.tril_indices(n, -1)
hess_arr[idx] = hess_arr.T[idx]
```

qcengine/programs/tests/test_nwchem.py
WardLT: I'm unsure, based on the `Array` type for `AtomicResult.return_result`. as well. I'll switch it to a 2D array, as that makes more sense to me.

qcengine/programs/nwchem/harvester.py
WardLT: Oh, that is much smarter! :D

qcengine/programs/tests/test_nwchem.py
WardLT: Ok, took me a bit to figure this out, but the base class for AtomicResult (ProtoModel) flattens all arrays when JSON serializing them. So, return_result will also be flattened regardless of what I do if I return a dictionary from `qcng.compute`

qcengine/programs/nwchem/harvester.py
WardLT: Okay, I had this wrong originally. TL;DR: Going to fix it. Long version...

After some further digging through NWChem source code, turns out the ".hess" file is the [lower matrix](https://github.com/nwchemgit/nwchem/blob/9d605c61bc3d4f18856c352f59ba0a5e75b0971b/src/stepper/stpr_gen_aux.F#L189).

`tril_indices` gives the correct behavior of having the "column" on the inner loop (0,0 -> 1,0 -> 1,1), whereas the `triu_indices` is equivalent to stepping over rows on the inner loop (0,0 -> 1,0 -> 2,0). 

Accordingly, the correct solution is similar to what I originally had but replacing `triu_indices` with `tril_indices`.

qcengine/programs/tests/test_nwchem.py
dgasmith: Ah, hmm. This must date back to some legacy code:
```python
    if return_dict:
        return json.loads(ret.json())
    else:
        return ret
```

We probably need to make an issue out of this where `return_dict` will return NumPy arrays which I think makes a bit more sense than the current formulation.

dgasmith: I do think we need to test a few more elements here to prevent the exact issue that you hit of triu/tril.

WardLT: Alright, the test now uses a non-dict return type - a 2D numpy array

WardLT:
I'm feeling paranoid, and am going to add in a test of a non-zero off-diagonal element

qcengine/programs/nwchem/harvester.py
```diff
@@ -727,17 +751,40 @@ def harvest(in_mol: Molecule, nwout: str, **kwargs) -> Tuple[PreservingDict, Non
...
+                    % (out_mol.nuclear_repulsion_energy(), in_mol.nuclear_repulsion_energy())
                 )
     else:
         raise ValueError("""No coordinate information extracted from NWChem output.""")
 
-    return outPsivar, None, outGrad, outMol, version, error
+    # If present, align the gradients and hessian with the original molecular coordinates
+    #  NWChem rotates the coordinates of the input molecule. `out_mol` contains the coordinates for the
+    #  rotated molecule, which we can use to determine how to rotate the gradients/hessian
+    amol, data = out_mol.align(in_mol, atoms_map=True, mols_align=True, verbose=0)
```
loriab: Just to head off an insidious bug -- `atoms_map=True` means user water in HOH order is never transformed by NWChem into OHH. So long as that's true, it's good to keep this bool as True, but I wanted to check.

WardLT: That's a good point. I am not sure about whether NWC will ever reorder atoms, so I'm going to push a commit that sets `atom_map=False`

loriab: Great. `False` will limit the size of mols that can be aligned in a timely manner, but grad/hess QC mols haven't been a problem yet.

PR #207, opened 2019-12-31
Minor issues involving >1 core per MPI rank

WardLT:
## Description
Fixes bugs with how the total number of ranks and memory size for NWChem are determined when there is more than one core per MPI rank

## Changelog description
Minor bug fixes for NWChem when more than core per MPI rank is used

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/util.py
+++ b/qcengine/util.py
@@ -43,7 +43,7 @@ def create_mpi_invocation(executable: str, task_config: TaskConfig) -> List[str]
     mpirun_str = task_config.mpiexec_command.format(
         nnodes=task_config.nnodes,
         ranks_per_node=task_config.ncores // task_config.cores_per_rank,
-        total_ranks=task_config.nnodes * task_config.ncores,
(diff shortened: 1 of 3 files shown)


qcengine/util.py
```diff
@@ -43,7 +43,7 @@ def create_mpi_invocation(executable: str, task_config: TaskConfig) -> List[str]
     mpirun_str = task_config.mpiexec_command.format(
         nnodes=task_config.nnodes,
         ranks_per_node=task_config.ncores // task_config.cores_per_rank,
-        total_ranks=task_config.nnodes * task_config.ncores,
+        total_ranks=task_config.nnodes * task_config.ncores // task_config.cores_per_rank,
```
dgasmith: Can you add a test here in `test_config.py`?

WardLT: Yes. It turns out I already had a test for changing cores_per_rank, but it had the wrong number of total ranks. I've fixed the test to have the correct values.

PR #210, opened 2020-01-03
Error correction for autosym issues in NWChem

WardLT:
## Description
NWChem has issues with symmetry detection, that [can be avoided](http://www.nwchem-sw.org/index.php/Special:AWCforum/st/id1195/%22sym_geom_project%3A_sym_center_ma....html) by altering the symmetry detection routine.

Here is a prototype for allowing QCEngine to diagnose the problem, apply the change, and re-run the executable. 

This kind of auto-corrections raises many issues that I'd like to discuss, such as:
- Is such a change to the input file, in the QCFractal sense, running a different calculation? Where do we draw the line on acceptable changes?
- Is it worth making this framework for "correctable errors" more general?
- Should we adopt a policy similar to random errors - a "CorrectableError" - and give the high level `compute` function control over how many mitigations are attempted (if any)?

What are your thoughts? 

## Changelog description
Error correction for autosym issues in NWChem

## Status

- [x] Code base linted
- [x] Ready to go


dgasmith:
Definitely worth making a more general framework. A few notes:
1) It seems like `protocols` could control the level and detail of autocorrection and restarting.
2) There are various buckets of autocorrection 
3) Being able to register corrections is a good design pattern for this.
4) Keeping track of which corrections were applied is crucial and in what order.
5) Some layers of this could impact Fractal, but as long as the QCEngine version and correct "level" is recorded it should work just fine.

WardLT:
Thanks for the advice. I'll play around with making a more generalized version of this and get back to you.

I'm leaning going the route of making a `CorrectableError` exception class that is caught in `qcng.compute()`, rather than deferring to the Harness (as I have it here)

WardLT:
I implemented a more general approach for the error correcting, which is ready for suggestions.

What would you recommend for keeping track of which corrections are made? 

`custodian` includes a specification scheme for marking what changes are made to an input file, but I'm not sure I want to go to that level of complexity. I do see the advantage of storing the actual changes made rather than only storing the name of the problem encountered (as shown now).  

My reasoning for storing "error name only" is that the names of the errors and QCEngine versions maintains the ability to re-generate the same inputs but without the need for a unified specification language that works for all QCEngine codes. We loose a human-interpretable list of the changes applied to the inputs, though I could probably add in human-readable descriptions of what changes are made.

loriab:
I had forgotten error correction happens in two repositories. Is this PR near RTG after rebase?

WardLT:
I think it should be OK. I'll sync with master and push a revised version

WardLT:
I'm starting to work on this again, but got distracted with a different PR.

WardLT:
I'm going to close this PR and start again. 

After more experience and thought, I'd like to revise how Harnesses should respond to errors. Instead of relying on the harness to implement correct code specific to each error type in the "input file writer" part, we should have the error objects themselves alter the input options passed to the input file writer (which will have no logic relating to error correction).This would mimic how Materials Project performs error correction and, if I recall correctly, was what @dgasmith was suggesting all along :)

dgasmith:
👍  Thanks for working through this, we had a window for a simpler approach and I'm glad you explored it. The MP approach will be a bit more complex, but I think we have seen signs that it works well so putting in the dev time there will produce a good result!

PR #212, opened 2020-01-06
CI for NWChem

### LINKED ISSUE #211: NWChem CI
While NWChem isn't on conda, but it is on several package managers. From their docs:

> Debian: https://packages.debian.org/search?keywords=nwchem
> Ubuntu: https://launchpad.net/ubuntu/+source/nwchem
> Fedora and EPEL: https://admin.fedoraproject.org/updates/search/nwchem
> Good search engine for NWChem Linux packages: http://pkgs.org/search/nwchem

It would be good to install this package and run it through our tests on Travis.

loriab:
closes #211 

## Description
try ubuntu's nwchem package, as recc by #211 

## Changelog description
- [x] Adds CI testing of NWChem 6.6 through Ubuntu xenial pkg

## Questions
- [ ] networkx is req'd for the NWChem interface for all but energies to align arrays, but at present it's only an optional requirement for qcel. There's a nice "install me" message, but in production users would probably like to know to install the pkg _before_ the QC job, not after. But I'm hesitant to put the check in `def found` because if ppl are only looking at the boolean, it'd be unintuitive for `which nwchem` to be present but nwchem harness not be available. Where to address this? require networkx? or have a distinction btwn nwchem-found and nwchem-ready?
- [ ] combine this CI lane with another?

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,7 +4,6 @@ coverage:
     - qcengine/programs/util/.*
     - qcengine/programs/cfour/.*
     - qcengine/programs/gamess/.*
-    - qcengine/programs/nwchem/.*
     - setup.py
   status:
     patch: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ install:
     elif [ $PROG == "OPENMM" ]; then
       python devtools/scripts/conda_env.py -n=test -p=$PYTHON_VER devtools/conda-envs/openmm.yaml
     elif [ $PROG == "NWCHEM" ]; then
-      python devtools/scripts/conda_env.py -n=test -p=$PYTHON_VER devtools/conda-envs/base.yaml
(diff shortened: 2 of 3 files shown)


dgasmith:
Awesome :) 

FYI @WardLT

dgasmith:
We have a similar issue with the Orca interface where CCLib may be required. Should `found` be more complex where you can say something like `found(binary_only=True)`? Seems awkward however.

loriab:
maybe something like the below for when main QC present but non-QC component missing.
```
>>> Program information
Available programs:
dftd3 v3.2.1
mp2d v1.1

Incomplete programs (see devtools/conda-envs for install help):
nwchem -- needs networkx
orca -- needs cclib

Other supported programs:
cfour entos gamess molpro mopac nwchem openmm psi4 qchem rdkit terachem torchani turbomole
```

Issue #213, opened 2020-01-07
Found, not found, and incomplete programs


dgasmith:
As was mentioned in #212, some programs require additional dependancies to function beyond the executable or Python project. Two examples are Orca (CCLib) and NWChem (networkx). The CLI support for this is fairly straightforward as shown in #212 by @loriab:

```
>>> Program information
Available programs:
dftd3 v3.2.1
mp2d v1.1

Incomplete programs (see devtools/conda-envs for install help):
nwchem -- needs networkx
orca -- needs cclib

Other supported programs:
cfour entos gamess molpro mopac nwchem openmm psi4 qchem rdkit terachem torchani turbomole
```

The main question is if we need to expand the `found()` syntax or some other resource to determine the difference between a "runnable" state and "found, but incomplete". A few proposals:

- Leave `found()` as is and have a new function `runnable()` that includes dependancies in its checks.
- Add an additional kwarg to `found(include_dependancies=True)`.
- Have `found()` return an `enum` or similar object that contains a variety of states `True`, `found_executable`, `found_dependancies`, `None`.

PR #215, opened 2020-01-10
psithon and psiapi

loriab:
## Description
~Adapt psi harness to choose psithon or psiapi for `found` and `version` fns. What I'm aiming for is to not require PATH _and_ PYTHONPATH environment edits to run psi4 (which calls qcng which calls psi4) when providing full path to exe.~

If psi4 is found through PATH _or_ PYTHONPATH, this sets the env in qcng to make both psithon and psiapi available.
- does not check if you had PATH pointing to one psi4 installation and PYTHONPATH to another. just don't do that.
- running in psiapi when psi4 in PATH only works for latest ddd branch b/c I had to add `psi --module` to access that importdir.
- result of this is that running tests in ddd no longer requires typing path to `bin/psi4` _and_ having PATH set _and_ having PYTHONPATH set. hoping this will avert lots of questions down the road.

wrt #213, the `found()` fn universally means fully operational, so let's keep that meaning and add extra checks for harness-dep programs into the found. For the setup guidance, let's just stop the generic "cannot be found" and instead propagate `found`'s nice messages and also link how to run `found` in the `qcengine info` page.

## Questions
- [ ] ~This PR works for my purposes except I have to hack around "psi4 is registered with QCEngine, but cannot be found." b/c psiapi vs psithon as a mode of operation (usually supplied in `AtomicInput`) isn't know at that stage. Any strong opinions on whether psithon and psiapi should be set up as separate harnesses based on whether PATH or PYTHONPATH had been provided by the user. An alternative is that both psithon and psiapi modes can be made available from one of PATH/PYTHONPATH so that `found` could look for both and figure out the mods to `env` that need to be passed to `popen`.~

## Changelog description
included

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/base.py
+++ b/qcengine/programs/base.py
@@ -103,7 +103,7 @@ register_program(QChemHarness())
 register_program(TeraChemHarness())
 register_program(TurbomoleHarness())
 
-# Semi-emperical
+# Semi-empirical
 register_program(MopacHarness())
 
 # AI
--- a/qcengine/programs/base.py
+++ b/qcengine/programs/base.py
@@ -66,8 +66,11 @@ def get_program(name: str, check: bool = True) -> "ProgramHarness":
         raise InputError(f"Program {name} is not registered to QCEngine.")
 
     ret = programs[name]
-    if check and not ret.found():
-        raise ResourceError(f"Program {name} is registered with QCEngine, but cannot be found.")
+    if check:
+        try:
+            found = ret.found(raise_error=True)
+        except ModuleNotFoundError as err:
+            raise ResourceError(f"Program {name} is registered with QCEngine, but cannot be found.") from err
 
     return ret
 
--- a/qcengine/programs/base.py
+++ b/qcengine/programs/base.py
@@ -68,7 +68,7 @@ def get_program(name: str, check: bool = True) -> "ProgramHarness":
     ret = programs[name]
     if check:
         try:
-            found = ret.found(raise_error=True)
+            ret.found(raise_error=True)
         except ModuleNotFoundError as err:
             raise ResourceError(f"Program {name} is registered with QCEngine, but cannot be found.") from err
 
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -34,34 +34,42 @@ class Psi4Harness(ProgramHarness):
         pass
 
     @staticmethod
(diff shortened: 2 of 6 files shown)


loriab:
This is now proven against Azure on the Psi4 DDD to allow the removal of lots o'`export` lines. I venture it's good to go.

qcengine/programs/psi4.py
```diff
@@ -34,6 +36,28 @@ class Config(ProgramHarness.Config):
...
+
+        if psiapi and not psithon:
+            psiimport = str(Path(which_import("psi4")).parent.parent)
+            env = os.environ.copy()
+            env["PYTHONPATH"] = psiimport
+            with popen(["python", "-c", "import psi4; print(psi4.executable[:-5])"], popen_kwargs={"env": env}) as exc:
+                exc["proc"].wait(timeout=30)
+            os.environ["PATH"] += os.pathsep + exc["stdout"].split()[-1]
+
+        if psithon or psiapi:
```
dgasmith: How costly is all of this logic? We have been having issues where trying to launch Psi4 concurrently on hundreds of nodes can swamp a NAS.

qcengine/programs/base.py
```diff
@@ -66,8 +66,11 @@ def get_program(name: str, check: bool = True) -> "ProgramHarness":
         raise InputError(f"Program {name} is not registered to QCEngine.")
 
     ret = programs[name]
-    if check and not ret.found():
-        raise ResourceError(f"Program {name} is registered with QCEngine, but cannot be found.")
+    if check:
+        try:
+            ret.found(raise_error=True)
+        except ModuleNotFoundError as err:
```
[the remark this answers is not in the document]
loriab: `ModuleNotFoundError` seemed the sensible thing to raise for `which_import` and `which` just mirrors it. Plus those are defined in qcel where there isn't `ResourceError`.

qcengine/programs/psi4.py
loriab: It's a hit. 0.4s for `psi4 --module`, 0.6s for `psi4.executable`. The latter isn't reduced by plain `import` instead of `subprocess` b/c `import psi4` is about 0.6s. Good news is that if you set it up the environment right beforehand (as a conda install would), then `Total time: 0.000456 s` for `qcng.get_program('psi4')`.

Issue #224, opened 2020-02-04
OpenMM energy calculation units


jthorton:
**Describe the bug**

When calculating the energy of a molecule using the OpenMM harness I get overly large energies. I think this is due to the unit conversion used resulting in hartrees/mol.

**To Reproduce**

This is an example that calculates the energy of a molecule in qcarchive using QCEngine
```python
import qcportal as ptl
from simtk import openmm, unit
from openforcefield.typing.engines.smirnoff import ForceField
from openforcefield.topology import Molecule
import qcelemental as qcel
import qcengine

# qcelemental converter
ureg = qcel.PhysicalConstantsContext("CODATA2014")

client = ptl.FractalClient()
# load a molecule from the qcarchive 
opts = client.get_collection('OptimizationDataset', 'SMIRNOFF Coverage Set 1')

# get the first entry from the set
entry = opts.get_entry(opts.df.index[0])

# grab the intial molecule
molecule = client.query_molecules(id=entry.initial_molecule)[0]
# use QCengine to compute the energy of the molecule

#set up the task
openmm_task = {"schema_name": "qcschema_input",
            "schema_version": 2,
            "molecule": molecule,
            "driver": "energy",
            "model": {"method": 'openmm', "basis": 'openff-1.0.0', 'offxml': 'openff-1.0.0.offxml'},
            "keywords": {"scf_type": "df"}}
# compue the energy
result = qcengine.compute(openmm_task, 'openmm')
print(result.return_result)
```
> -3.4178659499631775e+22
```python
# calculate again using openmm

# first make the offmol follwoing the method in qcengine
rdmol = qcengine.programs.rdkit.RDKitHarness._process_molecule_rdkit(molecule)
offmol = Molecule.from_rdkit(rdmol)
# load the force field 
ff = ForceField('openff-1.0.0.offxml')
#make the topology
off_top = offmol.to_topology()
# make the system
system = ff.create_openmm_system(off_top)
# make an integrator
integrator = openmm.VerletIntegrator(1.0 * unit.femtoseconds)
# set the platform
platform = openmm.Platform.getPlatformByName("CPU")
# create the context
context = openmm.Context(system, integrator, platform)

# Set positions from our Open Force Field `Molecule`
context.setPositions(offmol.conformers[0])

# Compute the energy of the configuration
state = context.getState(getEnergy=True)
# Get the potential as a simtk.unit.Quantity, put into units of hartree
q = state.getPotentialEnergy() / unit.hartree
[...]

PR #225, opened 2020-02-04
Fix openmm harness unit conversions for energy and gradient.

jthorton:
## Description

- [X] Fix for #224 
## Changelog description

Updated the OpenMM harness to use proper unit conversion.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/openmm.py
+++ b/qcengine/programs/openmm.py
@@ -21,6 +21,7 @@ from ..exceptions import InputError
 from ..util import temporary_directory
 from .model import ProgramHarness
 from .rdkit import RDKitHarness
+from ..units import ureg
 
 
 class OpenMMHarness(ProgramHarness):
@@ -204,9 +205,9 @@ class OpenMMHarness(ProgramHarness):
             state = context.getState(getEnergy=True)
 
             # Get the potential as a simtk.unit.Quantity, put into units of hartree
-            q = state.getPotentialEnergy() / unit.hartree
+            q = state.getPotentialEnergy() * ureg.conversion_factor("kJ / mol", "hartree")
 
-            ret_data["properties"] = {"return_energy": q.value_in_unit(q.unit)}
+            ret_data["properties"] = {"return_energy": q._value}
 
             # Execute driver
             if input_data.driver == "energy":
@@ -223,8 +224,9 @@ class OpenMMHarness(ProgramHarness):
                 gradient = state.getForces(asNumpy=True)
 
                 # Convert to hartree/bohr and reformat as 1D array
-                q = (gradient / (unit.hartree / unit.bohr)).reshape([n_atoms * 3])
-                ret_data["return_result"] = q.value_in_unit(q.unit)
+                coef = ureg.conversion_factor("kJ / mol", "hartree") * ureg.conversion_factor("nanometer", "bohr")
+                q = (gradient * coef).reshape([n_atoms * 3])
+                ret_data["return_result"] = q._value
             else:
                 raise InputError(
                     f"OpenMM can only compute energy and gradient driver methods. Found {input_data.driver}."
--- a/qcengine/programs/openmm.py
+++ b/qcengine/programs/openmm.py
@@ -204,10 +204,14 @@ class OpenMMHarness(ProgramHarness):
             # Compute the energy of the configuration
             state = context.getState(getEnergy=True)
 
-            # Get the potential as a simtk.unit.Quantity, put into units of hartree
-            q = state.getPotentialEnergy() * ureg.conversion_factor("kJ / mol", "hartree")
+            # Get the potential, this is assumed to be a simtk.unit.Quantity, in kJ/mol
+            q = state.getPotentialEnergy()
+            # convert to hartree using the qcelemental conversion factors
+            # note we use the protected attribute ._value which is just the float of the energy
+            # to avoid any differences in internal conversion factors between OpenMM and qcelemental
+            final_energy = q._value * ureg.conversion_factor("kJ / mol", "hartree")
 
-            ret_data["properties"] = {"return_energy": q._value}
+            ret_data["properties"] = {"return_energy": final_energy}
 
             # Execute driver
             if input_data.driver == "energy":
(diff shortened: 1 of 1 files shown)


jthorton:
Not sure why the tests are now failing as I only changed the Openmm harness, but the return value was not tested before.

qcengine/programs/openmm.py
```diff
@@ -204,9 +205,9 @@ def compute(self, input_data: "AtomicInput", config: "TaskConfig") -> "AtomicRes
             state = context.getState(getEnergy=True)
 
             # Get the potential as a simtk.unit.Quantity, put into units of hartree
-            q = state.getPotentialEnergy() / unit.hartree
```
dgasmith: We use the programs conversion factors so that the values are always the same. I would suggest keeping this here as everyones kJ is slightly different as it rarely follows canonical CODATA.

jthorton: That makes sense but the return value from this is not what we expect. I think energies should be in hartrees but from my example, in issue #224 you can see that they are not. Should I raise this on the openmm github instead?

dgasmith: I am not familiar enough with the OpenMM unit system to say, but I think you will need to divide my Avagodros number or similar here to make the units work out. @jchodera @j-wags

jthorton: Yes, I believe that is what the difference is in this case:
```python
# get in hartrees using simtk.unit.AVOGADRO_CONSTANT_NA
result.return_result / unit.AVOGADRO_CONSTANT_NA
```
> -0.05675498965565169

compared to the value using qcelemental
> -0.05602423977045684

Lnaden: See my comments on the issue for what I understand is happening in the unit system:

https://github.com/MolSSI/QCEngine/issues/224#issuecomment-581959798

dgasmith: Wow! This is why we _always_ want to use internal units. It should be much more consistent.

Lnaden: If I recall correctly (and if I understand what you're attempting to do), the internal units for OpenMM is in a [`kJ/mol` basis](http://docs.openmm.org/latest/userguide/theory.html#units), so internally, the double returned by `state.getPotentialEnergy()` is the internal one, just wrapped in a `Unit` class, so if you wanted to use a different unit conversion kit, that should still give you as close to the correct internal units as possible.

jthorton: If that's correct and allows us to use the qcelemental conversion method I would vote for that as I feel it is much cleaner to implement and understand whats going on.

Lnaden: I don't know all of the details of OpenMM internals, so a bit of caution here. I know it does a conversion to the internal unit system which I linked above, and decently sure it returns that as well unmodified. 

If you want just the raw number out of a `Unit` class, you can access it through one of two means (assume the object is `X`)

* X._value
* X.value_in_unit(x.unit)

The former just spits out the numeric float attached to the unit, but has the problem of accessing a non-public method, the later uses only public methods, but might result in a small adjustment of the value. I don't know if there is a special check for "if the same unit" which bypasses multiplying by the `simtk.unit` conversion factors. However, since the base units of energy and amount are kJ and mole respectively, I think their conversion factor is literally 1, so its doing `X * 1 / 1` internally (I think), just returning the double anyways.

If it was anything other than the internal units returned by `getPotentialEnergy`, I would be worried about the 2nd approach.

You should not have to worry about `getPotentialEnergy`'s unit returns changing. The internal OpenMM unit system has been the same for a long time and I don't think there is any appetite to do something different in the future. At least, no one has suggested changing it, ever.

jthorton: Thanks for the detailed explanation of how the units are working, I did a quick test and it looks like requesting the energy explicitly in KJ/mol gives the same value as `q._value` (option 1 which is what I have used in this PR):

```python
# now try using the in units of method
q = state.getPotentialEnergy() 
q.value_in_unit(q.unit)
```
> -149.01023036563316

```python
# compare with the _value
q = state.getPotentialEnergy() 
print(q._value)
```
> -149.01023036563316

Lnaden: Sure, but I think the only reason it works is because the base conversion the `unit` module is doing is `* 1 / 1`. Personally, I would assume that the units coming out of the `getPotentialEnergy` function are always kJ/mol (a safe assumption), and just use `._value` to avoid any and all possibility of some oddity with the `unit` module I don't know, then use QCElemental's CODATA to do all the conversions. 

It would behoove you to make a few detailed comments about what you are doing there since its a private method access and makes a few assumptions that won't be obvious to other Engine devs/users who look at the code and not this issue. Try and minimize future development issues going forward?

Lnaden:
Thanks for the comments, I think it will help if someone ever needs to come back in and look at this

qcengine/programs/openmm.py
dgasmith: The error here is 0.5 kcal/mol between QCElemental/OpenMM units which is pretty scary according to the first numbers. Can we track this down?

Test failures are from another source and not you.

dgasmith:
I need to patch this in #226, but this really leaves the open question of where the difference between QCElemental and OpenMM units is coming from. It seems very large.

jthorton:
I am away from my computer till tomorrow now but I can double check the values when I get in. I notice your change in #226 uses the simtk unit method which should be okay but you might need to check the gradient conversion it comes out in kj/mol nanometer from openmm.

dgasmith:
There is a test for the gradients now to make sure they are correct.

qcengine/programs/openmm.py
jthorton: Ahhh my mistake I think I have posted the energy using the qcelemental conversion for a different conformer, I have recalculated it for the same conformer and get 
> -0.05675499938859474 for `qcelemental`

and 

> -0.05675498965565169 for `simtk.unit`

which matches what I posted above and is much closer :+1:

jthorton:
Should I close the PR now then?

jchodera:
> I need to patch this in #226, but this really leaves the open question of where the difference between QCElemental and OpenMM units is coming from. It seems very large.

@dgasmith : Which CODATA unit set does QCElemental use? We're debating about when to adopt the new CODATA recommendations (and if we can help persuade some other codes to do so) here:
https://github.com/openmm/openmm/issues/2427

Perhaps we could try to standardize between a few packages?

dgasmith:
We currently pin QCArchive to CODATA2014 values, but we added CODATA2018 to QCElemental and may round up soon. It would be fantastic if we can standardize on a single unit implementation. It gets a bit tricky as pint doesn't natively handle many unit conversation found in CMS, see [here](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/physical_constants/ureg.py).

dgasmith:
Closed by #226.

qcengine/programs/openmm.py
dgasmith: Thanks for following up on this. I was pretty concerned about the original numbers, but this looks fine!

Lnaden: That makes me feel much better too!

loriab:
@jchodera if it'd be handy, qcel can pull in older codata sets so that ratios of 2006/2018 physconst are available.

Issue #192, opened 2019-12-08
Generate `basis` if not specified in OpenMMHarness


dotsdl:
During #151, it was decided that the `basis` field in `AtomicInput` for the `OpenMMHarness` need not be specified. When not specified, it would be generated from the contents of `url` or `offxml`. This is partially implemented, but in a fragile way that is not desirable, in `OpenMMHarness._generate_basis`.

Ideally, the `basis` contents would be generated as:
* `f"{forcefield_name}-{hash(forcefield_schema)}"` for non-versioned forcefields
* `f"{forcefield_name}-{forcefield_version}"` for versioned forcefields

However, there does not appear to exist a way to pull `forcefield_version` from the XML contents of an e.g. SMIRNOFF force field. This will likely require this addition to future releases of SMIRNOFF force fields.

j-wags:
@dgasmith brought up this Issue this morning, and we had a productive conversation. Basically, we thought a scheme like this would be good starting point for discussion

### Use examples

Desired force field | Program | Method | Basis | Keywords
-- | -- | -- | -- | --
openff-1.0.0 | openforcefield | parsley | 1.0.0 | constraints:true
openff_unconstrained-1.0.0 | openforcefield | parsley | 1.0.0 | none
Parsley with some simple bugfix (1.0.1) | openforcefield | parsley | 1.0.0 | none
Sage | openforcefield | sage | 1.1.0 | None
Sage | openforcefield | sage | Latest? | None
Rdkit UFF | rdkit | uff | None | None
AMBER GAFF | ambertools | gaff | 2.11 | {ES/NB cutoff, etc}

The "openforcefield" program string would be used to mean "the openforcefield toolkit and all dependencies, including OpenMM and openforcefields (the conda package containing Parsley)"

### The "constraints" keyword

Parsley's technical (actual) name is "openff-1.0.0.offxml", and it is intended for long-timestep MD. There is another "Parsley" that is intended for small molecule operations, which differs only by the removal of hydrogen bond length constraints, called `openff_unconstrainted-1.0.0.offxml". We understand that this is confusing, and it's caused by the fact that OpenFF and OpenMM consider the hbond constraints to be part of the FF (since it affects system energy) while most other comp chem packages don't. It's very confusing for everyone involved, but we have to do it this way unless we rewrite large portions of ParmEd. 

[...]

PR #228, opened 2020-02-05
RDKit MMFF94 and CI

dgasmith:
## Description
Adds MMFF94 to RDKit, adds tests for other OpenMM FF's, and re-enables TorchANI CI testing.

WARNING! I had implemented #192 backwards in the method/basis combination, this is also fixed here.

## Changelog description
Adds MMFF94 to RDKit computational abilities.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/tests/test_procedures.py
+++ b/qcengine/tests/test_procedures.py
@@ -150,6 +150,12 @@ def test_geometric_retries(failure_engine, input_data):
             "rdkit", {"method": "UFF"}, [1.87130923886072, 2.959448636243545, 104.5099642579023], marks=using("rdkit")
         ),
         pytest.param(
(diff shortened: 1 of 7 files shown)


qcengine/tests/test_procedures.py
```diff
@@ -149,6 +149,18 @@ def test_geometric_retries(failure_engine, input_data):
...
+        pytest.param(
+            "rdkit",
+            {"method": "mmff94"},
+            [1.8310842343589573, 2.884612338953529, 103.93822919865106],
+            marks=using("rdkit"),
+        ),
+        pytest.param(
+            "rdkit",
+            {"method": "MMFF94s"},
+            [1.8310842343589573, 2.884612338953529, 103.93822919865106],
```
loriab: answers expected to be exactly the same?

dgasmith: Yea, `s` only has different out of plane bending parameters and torsions.

PR #230, opened 2020-02-05
NWCHEM: Adrian dipole/homo/lumo changes

ahurta92:
## Description
fixes logic with parsing homo and lumo energies from nwchem output

## Changelog description
Fixes the logic when parsing for homo and lumo energies

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/nwchem/harvester.py
+++ b/qcengine/programs/nwchem/harvester.py
@@ -634,7 +634,7 @@ def harvest_outfile_pass(outtext):
                 re.MULTILINE,
             )
             if mobj:
-                if mobj.group(2) == "+": # if "positve power eg."2.1225011D+01"
+                if mobj.group(2) == "+":  # if "positve power eg."2.1225011D+01"
                     homo = float(mobj.group(1)) * (10 ** (float(mobj.group(3))))
                     psivar["HOMO"] = np.array([round(homo, 10)])
                 else:
--- a/qcengine/programs/nwchem/harvester.py
+++ b/qcengine/programs/nwchem/harvester.py
@@ -634,7 +634,7 @@ def harvest_outfile_pass(outtext):
                 re.MULTILINE,
             )
             if mobj:
-                if mobj.group(2) == "+":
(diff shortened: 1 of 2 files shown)


qcengine/programs/nwchem/harvester.py
```diff
@@ -636,28 +634,36 @@ def harvest_outfile_pass(outtext):
...
-                    + r"(\d+)",
-                    outtext,
-                    re.MULTILINE,
-                )
+                if mobj.group(2) == "+":  # if "positve power eg."2.1225011D+01"
+                    homo = float(mobj.group(1)) * (10 ** (float(mobj.group(3))))
+                    psivar["HOMO"] = np.array([round(homo, 10)])
+                else:
+                    homo = float(mobj.group(1)) * (10 ** (-1 * float(mobj.group(3))))
+                    psivar["HOMO"] = np.array([round(homo, 10)])
```
loriab: could you capture the sign, 0, and exponent together and `int` them, so that the if/else merges again?
```
>>> a="+01"
>>> 5**(int(a))
5
>>> a="-01"
>>> 5**(int(a))
0.2
```

loriab: or grab the whole thing and replace `a.replace("D", "E")`
or I think NUMBER handles D/E. if you try it and it gives trouble, paste me the error

dgasmith:
Can we add tests for these?

qcengine/programs/nwchem/harvester.py
ahurta92: Okay I made a change that captures the sign,0 and exponent together

ahurta92:
I also added a test for homo and lumo energies

qcengine/programs/nwchem/harvester.py
ahurta92: I captured the +-,0, and exponent and removed the if else blocks

mattwelborn:
What's the hold up on merging this? Did we just forget?

dgasmith:
Looks good, slipped through the cracks.

PR #232, opened 2020-02-16
add procedure for pyberny geometry optimizer

jhrmnn:
## Description

Following a discussion with @dgasmith, this PR adds a procedure for the [pyberny](https://github.com/jhrmnn/pyberny) geometry optimizer.

The PR is functional, the following test case runs sucessfully:

```python
import qcelemental as qcel
import qcengine as qcng

mol = qcel.models.Molecule(
    geometry=[[0, 0, 0], [0, 1.5, 0], [0, 0, 1.5]], symbols=["O", "H", "H"], connectivity=[[0, 1, 1], [0, 2, 1]]
)
opt_input = {
    "keywords": {"program": "mopac"},
    "input_specification": {"driver": "gradient", "model": {"method": "PM7"}},
    "initial_molecule": mol,
}
opt = qcng.compute_procedure(opt_input, "berny")
print(opt)
```

However, the PR is still a work in progress. The following needs to be resolved before merging, and I would welcome input from the QCEngine maintainers:

- [x] there is a lot of boilerplate shared with `GeometricProcedure`. `found()`, `build_input_model()`, `Config`, also it seems that if we have `found`, there should be no need to check for the presence of the package in `compute()` again. Is this necessary?
- [x] how to handle logging? Currently Berny logs verbosly, and this is adjustable by passing appropriate argumetns. How is this handled in QCEngine in general? (Pyberny doesn't use the `logging` module, but I can change that if QCEngine would prefer that.)
- [x] where to store paramters for Berny?

## Changelog description

Added a new geometry optimizer: [pyberny](https://github.com/jhrmnn/pyberny)

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/procedures/berny.py
+++ b/qcengine/procedures/berny.py
@@ -1,0 +1,84 @@
+import sys
+import logging
+import traceback
+from io import StringIO
+from typing import Any, Dict, Union
+
+import numpy as np
+from qcelemental.models import OptimizationInput, OptimizationResult
+from qcelemental.util import which_import
+
+import qcengine
+
+from ..config import TaskConfig
+from .model import ProcedureHarness
+
+
+class BernyProcedure(ProcedureHarness):
+    _defaults = {"name": "Berny", "procedure": "optimization"}
+
+    def found(self, raise_error: bool = False) -> bool:
+        return which_import(
+            "berny", return_bool=True, raise_error=raise_error, raise_msg="Please install via `pip install pyberny`.",
+        )
+
+    def build_input_model(self, data: Union[Dict[str, Any], "OptimizationInput"]) -> "OptimizationInput":
+        return self._build_model(data, OptimizationInput)
+
+    def compute(self, input_data: "OptimizationInput", config: "TaskConfig") -> "OptimizationResult":
+        try:
+            import berny
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError("Could not find Berny in the Python path.")
+
+        # Get berny version from the installed package, use setuptools'
+        # pkg_resources for python < 3.8
+        if sys.version_info >= (3, 8):
+            from importlib.metadata import distribution
+        else:
+            from pkg_resources import get_distribution as distribution
+        berny_version = distribution("pyberny").version
+
+        # Berny uses the stdlib logging module and by default uses per-module
+        # loggers. For QCEngine, we create one logger per BernyProcedure
+        # instance, by using the instance's id(), and send all logging messages
+        # to a string stream
+        log_stream = StringIO()
+        log = logging.getLogger(f"{__name__}.{id(self)}")
+        log.addHandler(logging.StreamHandler(log_stream))
+        log.setLevel("INFO")
+
+        input_data = input_data.dict()
+        geom_qcng = input_data["initial_molecule"]
+        comput = {**input_data["input_specification"], "molecule": geom_qcng}
+        program = input_data["keywords"].pop("program")
+        trajectory = []
+        output_data = input_data.copy()
+        try:
+            # Pyberny uses angstroms for the Cartesian geometry, but atomic
+            # units for everything else, including the gradients (hartree/bohr).
+            geom_berny = berny.Geometry(geom_qcng["symbols"], geom_qcng["geometry"] / berny.angstrom)
+            opt = berny.Berny(geom_berny, logger=log, **input_data["keywords"])
+            for geom_berny in opt:
+                geom_qcng["geometry"] = np.stack(geom_berny.coords * berny.angstrom)
+                ret = qcengine.compute(comput, program)
+                trajectory.append(ret.dict())
+                opt.send((ret.properties.return_energy, ret.return_result))
+        except Exception:
+            output_data["success"] = False
+            output_data["error"] = {"error_type": "unknown", "error_message": f"Berny error:\n{traceback.format_exc()}"}
+        else:
+            output_data["success"] = True
+        output_data.update(
+            {
+                "schema_name": "qcschema_optimization_output",
+                "final_molecule": trajectory[-1]["molecule"],
+                "energies": [r["properties"]["return_energy"] for r in trajectory],
+                "trajectory": trajectory,
+                "provenance": {"creator": "Berny", "routine": "berny.Berny", "version": berny_version},
+                "stdout": log_stream.getvalue(),  # collect logged messages
+            }
+        )
+        if output_data["success"]:
+            output_data = OptimizationResult(**output_data)
+        return output_data
--- a/devtools/conda-envs/psi.yaml
+++ b/devtools/conda-envs/psi.yaml
@@ -21,3 +21,7 @@ dependencies:
   - pytest
   - pytest-cov
   - codecov
+
+  - pip
+  - pip:
+    - pyberny
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -84,6 +84,7 @@ In addition, several procedures are available:
 - Geometry Optimization:
 
   - `geomeTRIC <https://github.com/leeping/geomeTRIC>`_
+  - `Pyberny <https://github.com/jhrmnn/pyberny>`_
 
 Configuration Determination
 ---------------------------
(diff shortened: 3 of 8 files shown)


jhrmnn:
I guess some tests should be added as well, haven't look into that yet.

qcengine/procedures/berny.py
```diff
@@ -0,0 +1,71 @@
...
+# TODO there is a lot of boilerplate shared with GeometricProcedure
+# TODO how to handle logging?
+# TODO where to store paramters for Berny?
+
+
+class BernyProcedure(ProcedureHarness):
+
+    _defaults = {"name": "Berny", "procedure": "optimization"}
+
+    class Config(ProcedureHarness.Config):
```
dgasmith: This can be removed, pydantic no longer requires this.

qcengine/procedures/berny.py
```diff
@@ -0,0 +1,71 @@
...
+        geom_qcng = input_data["initial_molecule"]
+        comput = {**input_data["input_specification"], "molecule": geom_qcng}
+        trajectory = []
+        output_data = input_data.copy()
+        try:
+            geom_berny = berny.Geometry(geom_qcng["symbols"], geom_qcng["geometry"])
+            opt = berny.Berny(geom_berny)
+            for geom_berny in opt:
+                geom_qcng["geometry"] = np.stack(geom_berny.coords)
+                ret = qcengine.compute(comput, **input_data["keywords"])
```
dgasmith: `from ..compute import compute`. I think you will also need a `input_data["keywords"]["program"]` here.

qcengine/procedures/berny.py
```diff
@@ -0,0 +1,71 @@
...
+            import berny
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError("Could not find Berny in the Python path.")
+        input_data = input_data.dict()
+        geom_qcng = input_data["initial_molecule"]
+        comput = {**input_data["input_specification"], "molecule": geom_qcng}
+        trajectory = []
+        output_data = input_data.copy()
+        try:
+            geom_berny = berny.Geometry(geom_qcng["symbols"], geom_qcng["geometry"])
```
dgasmith: Do you use atomic units for everything internally?

jhrmnn: Yes.

qcengine/procedures/berny.py
jhrmnn: `input_data["keywords"]` already contains `{"program": ...}`, so that's why this works. I thought that `input_data["keywords"]` is supposed to be passed to `qcengine.compute()` in this way. Is it?

jhrmnn:
> 2. For logging its best to use the logging module so that we can easily point the `stdout` to a string that we capture in the `stdout/stderr` fields.

I've switched to the `logging` module in `pyberny`. It prints just on the `INFO` level, so I guess by default it won't print at all within QCEngine. Is that what we want? On it's own `pyberny` doesn't set the `logging` module in any way. Should there be something added to QCEngine regarding this?

> 3. Which parameters are you thinking about storing? There is an `extras` field on the return schema where anything can be stored.

Essentially [these](https://github.com/jhrmnn/pyberny/blob/master/berny/berny.py#L87-L88). The `params` has defaults [here](https://github.com/jhrmnn/pyberny/blob/master/berny/berny.py#L26-L49). One should be able to pass these through QCEngine. So the `extras` field is for that?

> 4. There is a `test_procedures.py` file for adding tests. Feel free to add pybernie to the CI by adding is as a pip dependency to one of `devtools/conda-envs/*.yaml`s.

So I don't see Mopac in any of your CI envs. I use Mopac for CI testing of pyberny because it's pretty lightweight. What should I use in QCEngine?

qcengine/procedures/berny.py
jhrmnn: x

jhrmnn: Erm, clearly I should be reading my own documentation. Pyberny works in angstroms, but so does QCEngine, so everything checks out?

loriab: QCEngine woks exclusively in Bohr.

jhrmnn: But `Molecule`'s API uses angstroms, or? At least I thought so based on this paragraph: http://docs.qcarchive.molssi.org/projects/QCElemental/en/stable/model_molecule.html#fragments

loriab: Ah, the _string initialization_ of `Molecule` defaults to angstroms (you can add a `units au` line to the string to make it interpret as Bohr). But the internal storage is stringently au. By the way, there are array or dictionary initializers, too, if something like that would be more convenient.

You can consult http://docs.qcarchive.molssi.org/projects/QCElemental/en/stable/model_molecule.html#fields for the internal fields. (Ignore that some have trailing underscore (e.g., `masses_) -- that's a detail that allows skipping storage of fields with default value.)

jhrmnn: Interesting! So indeed this was incorrect and had to add the conversion factor between angstrom and bohr. What's interesting is that the optimization worked anyway, albeit less efficiently.

As long as the coordinates/gradients are scaled contra/covariantly (as was the case here), the gradient descent is unaffected. The only difference was that Pyberny detected the molecule as three unbonded atoms, that is, as a van der Waals cluster—which I should have notice. But it sill optimized it.

Anyway, thanks for the catch.

qcengine/procedures/berny.py
jhrmnn: So is [this](https://github.com/MolSSI/QCEngine/pull/232/commits/7ef994ede43cac838e119d59eb1e46ce2fec53ba#diff-4384465c212e54921f44238616ea614aR44) how it should be done? Passing only `"program"` explicitly.

jhrmnn:
Ok, so I'm [taking](https://github.com/MolSSI/QCEngine/pull/232/commits/7ef994ede43cac838e119d59eb1e46ce2fec53ba#diff-4384465c212e54921f44238616ea614aR41) the kwargs for `Berny` from `input["keywords"]["berny_kwargs"]`. Is that how it's supposed to be done?

What's the semantic distinction between `"keywords"` and `"extras"`?

dgasmith:
1) All logging information should be stored in a top level `stdout` field.
2) The `keywords` top level field is the pybernie keywords with the execution of "program" which is popped for QCEngine use. This will be redesigned at a later date.
3) I couldn't see what you wish to store in extras as the link appears to be broken, but yes `extra` can be anything.
4) See `tests/test_procedures.py` for examples. We either use Psi4 or RDKit typically. I wish we could use more semi-empirical in the testing, but license issues become a problem. We run MOPAC only on the master branch, security issues prevent us from doing it on any PR.

jhrmnn:
Ok, I think I've resolved all the existing issues. I believe the failing test has nothing to do with Berny, the Berny test succeeds.

The commits should be squashed for a merge.

qcengine/procedures/berny.py
```diff
@@ -0,0 +1,72 @@
...
+    def compute(self, input_data: "OptimizationInput", config: "TaskConfig") -> "OptimizationResult":
+        try:
+            import berny
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError("Could not find Berny in the Python path.")
+        log_stream = StringIO()
+        log = logging.getLogger(f"{__name__}.{id(self)}")  # per-instance logger
+        log.addHandler(logging.StreamHandler(log_stream))
+        log.setLevel("INFO")
+        input_data = input_data.dict()
```
dgasmith: Can you give some whitespace lines here and a few comments similar to other modules?

jhrmnn: Done.

jhrmnn:
> Can you add yourself [here](https://github.com/MolSSI/QCEngine/blob/master/docs/source/index.rst).

Done. This should be good to go. I've already stashed all the commits.

jhrmnn:
I'm happy to have contributed! Nope, no citation at this point. I might think about it if I ever get to implement the linear bends, then the package would implement the [Standard Method](http://link.springer.com/10.1007/s00214-016-1847-3). Then it would make sense to make the package more official.

jhrmnn:
@dgasmith I've created a DOI handle for Pyberny, so you could use that in the QCEngine paper: https://doi.org/10.5281/zenodo.3695038

dgasmith:
Got it thanks! I have it marked down to tweak when we get feedback on the paper.

PR #233, opened 2020-02-19
added optional param for execute: exit_code

Andrew-AbiMansour:
execute method returns False if retcode > exit_code

## Description

Enable util.execute to return True when the return code > 0. This is useful for Unix utilities such as grep.  

## Changelog description

Added optional param 'exit_code' (default 0) to 'execute' function.


dgasmith:
@Andrew-AbiMansour Gentle poke here.

PR #234, opened 2020-02-27
Update to support MDI v1.0.0

taylor-a-barnes:
## Description
The MDI interface implementation in QCEngine currently only supports the pre-release version of the MDI Library.  This PR updates the interface implementation to accommodate changes that were introduced in the 1.0 version of the MDI Library.  The primary modifications include changes to (1) the way NumPy arrays are handled by the MDI Library and (2) updates to the manner in which node information is communicated.

## Changelog description
Add support for MDI Library v1.0.0 release

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/mdi_server.py
+++ b/qcengine/mdi_server.py
@@ -231,7 +231,7 @@ class MDIServer:
         natom = len(self.molecule.geometry)
         if coords is None:
             coords = np.zeros(3 * natom)
-            MDI_Recv(3 * natom, MDI_DOUBLE, self.comm, buf = coords)
+            MDI_Recv(3 * natom, MDI_DOUBLE, self.comm, buf=coords)
         new_geometry = np.reshape(coords, (-1, 3))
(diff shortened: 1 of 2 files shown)


qcengine/mdi_server.py
```diff
@@ -108,24 +109,34 @@ def __init__(
...
+            ">MASSES": self.recv_masses,
             "<TOTCHARGE": self.send_total_charge,
             ">TOTCHARGE": self.recv_total_charge,
             "<ELEC_MULT": self.send_multiplicity,
             ">ELEC_MULT": self.recv_multiplicity,
             "EXIT": self.stop,
         }
 
+        # Register the @DEFAULT node
+        MDI_Register_Node("@DEFAULT")
```
dgasmith: I think we need to add a check for the MDI version at the start of the init with these new features.:

```python
from qcelemental.util import parse_version
if parse_version(mdi.__version__) < parse_version("1.0.0"):
    raise ...
```

taylor-a-barnes: @dgasmith Good catch.  I've added the check.

PR #235, opened 2020-03-03
SAPT-D!

dsirianni:
## Description

Adds additive dispersion correction to SAPT (pub upcoming from CDS group) to engine, and enables return of atomic pairwise dispersion from DFTD3.

## Changelog description

Adds SAPT-D & pairwise DFTD3 analysis.

## Status

- [x] Code base linted
- [x] Ready to go


qcengine/programs/dftd3.py
```diff
@@ -143,6 +143,10 @@ def build_input(
         if input_model.extras["info"]["dashlevel"] == "atmgr":
             command.append("-abc")
 
+        # Append `-anal` for pairwise atomic analysis
+        if input_model.driver != "gradient":
```
dgasmith: @loriab Seems like this should be an optional flag in keywords?

qcengine/programs/dftd3.py
```diff
@@ -187,6 +191,22 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+                # Iterate over block
+                start = stdout.splitlines().index(ln) + 2
+                for l in stdout.splitlines()[start:]:
+                    data = l.replace("-", " -").split()
+                    # print(data)
+                    if len(data) == 0:
+                        break
+                    atom1 = int(data[0]) - 1
+                    atom2 = int(data[1]) - 1
+                    Edisp = Decimal(data[-1])
```
dgasmith: Not sure we want to store this data as Decimal, these things are pretty explicitly floats.

qcengine/programs/dftd3.py
```diff
@@ -279,8 +299,10 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
         }
         output_data["extras"]["local_keywords"] = input_model.extras["info"]
         output_data["extras"]["qcvars"] = calcinfo
-
+        if input_model.driver != "gradient":
+            output_data["extras"]["qcvars"]["PAIRWISE DISPERSION CORRECTION ANALYSIS"] = D3pairs
```
dgasmith: We need to check this term explicitly, perhaps the norm of the values?

loriab: yes, that's the focus of this afternoon's session :-)

qcengine/programs/dftd3.py
loriab: `Decimal` here is so as to not lose the info of how many digits of float we are getting back from the program.

qcengine/programs/empirical_dispersion_resources.py
```diff
@@ -649,7 +649,7 @@
             "pbe": {"params": {"s6": 1.000, "s8": 0.358940, "a1": 0.012092, "a2": 5.938951}},
             "pbe0": {"params": {"s6": 1.000, "s8": 0.528823, "a1": 0.007912, "a2": 6.162326}},
             "wpbe": {"params": {"s6": 1.000, "s8": 0.906564, "a1": 0.563761, "a2": 3.593680}},  # formerly lcwpbe
-            "sapt0": {"params": {"s6": 1.000, "s8": 1.409027, "a1": 0.078803, "a2": 4.553948}},  # experimental
+            "sapt0": {"params": {"s6": 1.000, "s8": 1.03220439, "a1": 0.00704073, "a2": 6.11845609}},  # experimental
```
loriab, line 652: how's the refitting going, @dsirianni? note that 4-6 sig figs is plenty, so perhaps the procedure is optimize, truncate, recompute to make sure we're using the same tidy parameter values here and in paper.

loriab:
superseded by #282

PR #238, opened 2020-04-10
standard suite testing: mp2, ccsd

loriab:
## Description
Revamp testing of QC programs for common method, particularly to check they can be badgered into giving the same answers as other programs. Sets up a testing framework so that can focus on constructing `AtomicInput` and harvesting rather than on collecting reference values, capturing errors, and designing tests.

## Changes
- [x] collect ref values for a few mols/basis sets/refs/algorithms in `standard_suite_ref.py`. used by qcng, qcdb, psi4 (and anyone else who cares to).
- [x] consolidate expectations of what properties should emerge from a level of theory and encode still-in-working-order exceptions for different QC programs/modules. used by qcng, qcdb, psi4
- [x] collect all mp2 & ccsd tests from qcdb and reformat them (more pytest.parameterize) into `test_standard_suite.py`. This new file is a lot more testing than before (old `test_standard_suite_mp2|ccsd.py` was only sampling), is more thorough (checking properties, not just return_result), includes more basis sets and molecules (good way to catch bugs, that), and is prepared to handle & document errors at all stages (won't run, incomplete run, wrong run). file is similar to qcdb test file but not shared.
- [x] fill in parsing of properties as qcvars in cfour, gamess, nwchem for mp2/ccsd, rhf/uhf/rohf, ae/fc. systematically copy qcvars to atomicproperties. some minor changes to the psi4 schema_wrapper were necessary.
- [x] a harvester should be focused on the output file, not in filling in total vs correlation energies or the rarely-reported-under-those-names singles/doubles energies for every MP2 pattern separately. the file `qcvar_identities_resources.py` brings over from qcdb that uses common identities to fill in missing qcvars, no matter what set were harvested. this file also consolidates the qcvar to atomicproperties translations that had been repeated.
- [x] managing all running, asserting, and expected errors has been consolidated in `standard_suite_runner.py`. this file is similar to those in qcdb and psi4 but is not shared.
- [x] without testing the result, dipole components moved to dipole array.
- [x] a previous formatting pass had scrambled the nwchem regexes so that lines weren't 1:1. and the broad formatting exceptions meant the harvester files were looking ragged and hard to read. so the formatting exceptions have been localized and the files black-ed

## Status
**Reviewers: look at first commit, which is all the substantive changes. second commit is purely formatting. further commits are non-formatting but very minor.**

- [x] Code base linted
- [x] Ready to go


loriab:
ok, there's no particular urgency on this, but it's now ready for consideration.

PR #240, opened 2020-04-17
NWChem support for MP2 with open shells

WardLT:
## Description
Adds support for MP2 calculations with open shells. It seems like [NWChem needs to be run with UHF](http://www.nwchem-sw.org/index.php/Special:AWCforum/st/id2020/unknown_SCFTYPE_error.html) for charged systems.

## Changelog description
- Adds a test showing the current qcengine fails for open shell MP2

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_nwchem.py
+++ b/qcengine/programs/tests/test_nwchem.py
@@ -184,3 +184,24 @@ def test_homo_lumo(h20v2):
     # Make sure Dipole Moment and center of charge parsed correctly
     assert compare_values(-0.2636515, float(res["extras"]["qcvars"]["HOMO"][0]), atol=1e-5)
     assert compare_values(0.08207131, float(res["extras"]["qcvars"]["LUMO"][0]), atol=1e-5)
+
+
+@pytest.mark.parametrize("mol,charge", [
+    (nh2, 0),
+    (h20v2, 1)
+])
+@using("nwcherm")
+def test_mp2_charged(mol: qcel.models.Molecule, charge: int):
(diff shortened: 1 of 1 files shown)


loriab:
I don't object to another, but I think I've got that xfailing here https://github.com/MolSSI/QCEngine/pull/238/files#diff-c19d7324f5148307a751cd543726fa45R169 https://github.com/MolSSI/QCEngine/pull/238/files#diff-c19d7324f5148307a751cd543726fa45R28-R31

WardLT:
Thanks, @loriab . I did not check for PRs. 

I'm currently working on fixing this issue. Are you also doing that?

loriab:
So that PR fills out parsing for r/u/rohf ae/fc normal/tce nwchem harvesting for mp2/ccsd. The problem as I understand it is only tce does rohf and tce isn't printing _or_ adding in the singles contribution, so the energy's wrong. Kowalski knows and will add in printing sometime https://github.com/qcdb/qcdb/issues/29 , so I think it's out of our hands, unless you're in a position to work on the tce code.

WardLT:
Interesting. I'll read that qcdb PR to try to understand the TCE+NWChem+ROHF issues (it's a little beyond my QC understanding). 

Will defaulting to use UHF for open shell calculations be an appropriate work around?

loriab:
If the user has handed an open-shell mol and a demand for MP2, then defaulting to UHF can be ok, like for https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/psi4.py#L127-L128 . Different methods have different saner defaults among uhf and rohf, though.

PR #241, opened 2020-04-22
Improve performance of NWChem harvester

WardLT:
## Description
Reduces performance issues in the NWChem harness by using `atoms_map=True`. I had hit an edge-case (script to reproduce below) that caused memory usage to spike during molecular alignment.

## Changelog description
Improve performance of NWChem harness

## Status

- [x] Code base linted
- [x] Ready to go

```python
from qcengine import compute_procedure
from qcelemental.models import Molecule
import json

# Re-create the problematic task task
mol = Molecule.from_data('''17
1 2 CC(C)(C)C_oxidized
C                    -0.018590000703     1.534862999948     0.010398998953
C                     0.002102997861    -0.004137001598     0.001969999753
C                     0.746440000483    -0.513907998052     1.248875997535
C                    -1.441902000132    -0.536735999114     0.013115000987
C                     0.722350001733    -0.500496001647    -1.264456001221
H                     0.997769001867     1.944204999076     0.004355001441
H                    -0.541819000228     1.928414001853    -0.868198998732
H                    -0.527584000967     1.918884001840     0.901485002006
H                     0.255496002037    -0.176026999637     2.168331000349
H                     0.778153998185    -1.608924001759     1.270889002192
H                     1.779648000149    -0.149758002595     1.269439998571
H                    -1.998718999074    -0.188888000854    -0.864080999888
H                    -1.459568999361    -1.632271998808     0.006498999953
H                    -1.980888002537    -0.199795999491     0.905615997443
H                     1.754825001002    -0.135540000506    -1.301181000186
H                     0.753970001687    -1.595239997685    -1.298223000122
H                     0.213582002473    -0.153439001607    -2.170672000953
''', molecular_charge=1.0)
task = {
    "initial_molecule": mol,
    "input_specification": {
        "driver": "gradient", 
        "model": {"method": "b3lyp", "basis": "6-31g(2df,p)"},
    }, 
    "keywords": {"program": "nwchem"}
}

compute_procedure(task, 'geometric', raise_error=True)
```


qcengine/programs/nwchem/harvester.py
```diff
@@ -855,9 +855,9 @@ def harvest(in_mol: Molecule, nwout: str, **outfiles) -> Tuple[PreservingDict, N
     # If present, align the gradients and hessian with the original molecular coordinates
     #  NWChem rotates the coordinates of the input molecule. `out_mol` contains the coordinates for the
     #  rotated molecule, which we can use to determine how to rotate the gradients/hessian
-    amol, data = out_mol.align(in_mol, atoms_map=False, mols_align=True, verbose=0)
-
+    amol, data = out_mol.align(in_mol, atoms_map=True, verbose=0)
```
loriab, line 858: Why remove the `mols_align=True`? This alignment is between the schema mol orientation and nwc's internal standard orientation, so final rmsd should be truly zero.

WardLT: The problem here is that if we run NWChem with auto symmetry detection, it can perturb atomic positions enough to get an RMSD of greater than the error-checking cutoff in QCEl's alignment routine. 

I've got a test case that shows the effect of turning on and off autosym toggling on/off the error with `mols_align=True`

What's your take? My vote is to keep `mols_align=False` so that the code will work with NWChem's autosym. NWChem (at least v7) does not print out the molecular after rotation but before symmetrization, which is what we'd need to get the proper mill for rotating the gradients/hessian.

WardLT:
FYI: I added a test that makes sure that symmetrizing the atomic coordinates does not break the NWChem output parser.

qcengine/programs/nwchem/harvester.py
loriab: aha, incomplete docstring in models/molecule.py . try setting `mols_align` to a relaxed float https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molutil/align.py#L146-L151

WardLT: Done! Thanks for pointing that out. 

PS. I'm updating the qcel docs for `mols_align`

loriab:
good for merge on both PRs?

WardLT:
Yes, I think they are both ready.

PR #243, opened 2020-04-29
Return stdout rather than stderr

WardLT:
## Description
The stdout from NWChem contains better debugging information than the stderr.

## Changelog description
Improved debugging output from NWChem runner

## Status

- [x] Code base linted
- [ ] Ready to go


loriab:
please rebase to fix CI

PR #249, opened 2020-05-29
Implement support for xtb API

awvwgk:
## Description

Add thin wrapper around [`xtb-python`](https://xtb-python.readthedocs.io) project. For the integration into QCEngine this should cover all the changes needed for `xtb` support. The actual harness for `xtb` will be maintained in the `xtb-python` project.

~~Important: `xtb-python` is still work in progress and not yet available on conda-forge!~~

## Changelog description

Implement support for the extended tight binding (`xtb`) program package.

## Status

- [x] Code base linted
- [x] Ready to go


qcengine/programs/xtb.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        Actual interface to the xtb package. The compute function is just a thin
+        wrapper around the native QCSchema interface of xtb-python.
+        """
+
+        self.found(raise_error=True)
+        import xtb
+        from xtb.qcschema.harness import run_qcschema
+
+        return run_qcschema(input_data)
```
loriab, line 55: Note that handling successful xtb runs is comparatively easy to managing error runs (either of execution or job). 

For orientation looking at psi models, the old way (which _doesn't_ depend on qcng in psi itself and only supports executable mode, `psi4 --json`) is handled https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/psi4.py#L133-L185 and https://github.com/psi4/psi4/blob/master/psi4/driver/schema_wrapper.py#L431-L479. 

Whereas the newer way (which _does_ depend on qcng `model_wrapper` and supports both executable mode and python mode, `psi4.run_qcschema`, which is more like what you want) is handled in https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/psi4.py#L189-L234 and https://github.com/psi4/psi4/blob/master/psi4/driver/schema_wrapper.py#L381-L428).

awvwgk: I guess, most of this should get addressed on the `xtb-python` side. There is no command line interface for `xtb-python` yet and I don't really plan to add one in the near future, also making an IO harness for the real `xtb` binary is out-of-scope for this PR.

For `qcng` it would be advantageous to (optionally) wrap the API call in a subprocess in case there is an error, which is not caught in the API and returned properly as error in the `AtomicResult`. This is definitely something I have to look into.

awvwgk:
I also added `xtb` to the documentation. I saw you have a separate section to describe the implemented methods for force fields, while `xtb` provides experimental access to one force field, I doubt it is a suitable place to add the tight binding Hamiltonians there.

Should I create a new section for semiempirical methods (which would also be useful for MOPAC)?

loriab:
sure, a semiempirical docs section sounds good.

awvwgk:
Tests are probably failing due to #252, I guess.

loriab:
> Tests are probably failing due to #252, I guess.

Yes, we'll get #253 in to straighten it out.

awvwgk:
We also have a force-field in `xtb` but it does not follow your conventions for molecular modelling, in the sense that it does not require nor use an external topology input.

It is already functional. but the API access is somewhat experimental and still needs to be smoothed out a bit. Therefore, I did not include it in the testsuite yet, as we might have break it on the `xtb` API side at some point.

awvwgk:
No need to hold this back. With this integration model this is an upstream issue for `xtb-python` or the `xtb` API anyway. Once we got it in a good shape there, I'll offer additional regression tests for qcng.

PR #256, opened 2020-06-27
[WIP] GAMESS Evaluate EFP Energy

zachglick:
## Description

Add support for GAMESS's EFP method, which calculates the interaction energy between multiple molecular fragments using a specified potential for each fragment.

A couple things have to happen before this PR is ready (see https://github.com/MolSSI/QCElemental/pull/124)

- Support for "empty" QCElemental `Molecule` (EFP fragments will be stored in `Molecule.extras`)
- Agreement on how to specify of EFP fragment geometries/orientations, and if/how they should be validated.
- Agreement on how to handle the actual fragment potentials (i.e. the output of GAMESS's makefp command). Is the user expected to provide the potential(s)? If so, in what format. If not, QCEngine needs a library of potentials.

## Changelog description

GAMESS EFP Support

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/gamess/harvester.py
+++ b/qcengine/programs/gamess/harvester.py
@@ -26,7 +26,10 @@ def harvest(p4Mol, gamessout: str, **largs) -> Tuple[PreservingDict, Molecule, l
     if outMol:
         outqcvar["NUCLEAR REPULSION ENERGY"] = outMol.nuclear_repulsion_energy()
         if p4Mol:
-            if abs(outMol.nuclear_repulsion_energy() - p4Mol.nuclear_repulsion_energy()) > 1.0e-3:
+            # temporary hack until qcel lets us do EFP with an 'empty' QM molecule
+            if "efp" in p4Mol.extras:
+                outMol = p4Mol
+            elif abs(outMol.nuclear_repulsion_energy() - p4Mol.nuclear_repulsion_energy()) > 1.0e-3:
                 raise ValueError(
                     """gamess outfile (NRE: %f) inconsistent with Psi4 input (NRE: %f)."""
                     % (outMol.nuclear_repulsion_energy(), p4Mol.nuclear_repulsion_energy())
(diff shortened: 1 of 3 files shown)


qcengine/programs/gamess/runner.py
```diff
@@ -107,7 +107,18 @@ def build_input(
...
         optcmd = format_keywords(opts)
 
-        gamessrec["infiles"]["gamess.inp"] = optcmd + molcmd
+        # Currently passing in the entire GAMESS-formatted EFP command (i.e. the $EFRAG section)
+        # In the future, should be passed in as actual data fragments, coordinates, etc.)
+        if input_model.molecule.extras is not None and "efp" in input_model.molecule.extras:
+            efpcmd = input_model.molecule.extras["efp"]
+            gamessrec["infiles"]["gamess.inp"] = optcmd + efpcmd + molcmd
+        else:
+            gamessrec["infiles"]["gamess.inp"] = optcmd + molcmd
```
loriab, line 116: what's in `molecule.extras["efp"]` again? I don't recall it from https://github.com/MolSSI/QCElemental/pull/124#issuecomment-527462426 (though that needn't be a strict blueprint).

qcengine/programs/gamess/harvester.py
```diff
@@ -445,6 +537,10 @@ def harvest_outfile_pass(outtext):
         qcvar["CURRENT REFERENCE ENERGY"] = qcvar["DFT TOTAL ENERGY"]
         qcvar["CURRENT ENERGY"] = qcvar["DFT TOTAL ENERGY"]
 
+    if "EFP TOTAL ENERGY" in qcvar:
+        qcvar["CURRENT REFERENCE ENERGY"] = qcvar["EFP TOTAL ENERGY"]
```
loriab, line 541: unless something needs it, maybe drop the current ref line. it's questionable.

qcengine/programs/gamess/harvester.py
```diff
@@ -319,6 +322,95 @@ def harvest_outfile_pass(outtext):
...
+            r"^\s+" + r"FINAL EFP ENERGY      =" + r"\s+" + NUMBER + r"\s*",
+            outtext,
+            re.MULTILINE,
+        )
+        if mobj:
+            logger.debug("matched efp")
+            print("matched efp")
+            qcvar["EFP ELECTROSTATIC ENERGY"] = mobj.group(1)
+            qcvar["EFP REPULSION ENERGY"] = mobj.group(2)
+            qcvar["EFP POLARIZATION ENERGY"] = mobj.group(3)
```
loriab, line 354: nice for now, may have to revisit in future. i'm against a 3rd labels set. https://github.com/loriab/pylibefp/blob/master/pylibefp/wrapper.py#L961-L966

qcengine/programs/gamess/harvester.py
zachglick: I had to add that line since `CURRENT REFERENCE ENERGY` is expected here:
https://github.com/MolSSI/QCEngine/blob/4525eb3ea94557c7f9aed400a02bf98a95ce5c92/qcengine/programs/gamess/runner.py#L155

What exactly is questionable about it? Would setting `CURRENT ENERGY` make more sense? Does `CURRENT REFERENCE ENERGY` imply a full QM calculation?

qcengine/programs/gamess/harvester.py
```diff
@@ -319,6 +322,95 @@ def harvest_outfile_pass(outtext):
...
+            + r"^\s+"
+            + r"POLARIZATION ENERGY   ="
+            + r"\s+"
+            + NUMBER
+            + r"\s*"
+            + r"^\s+"
+            + r"--------------------------------------"
+            + r"\s*"
+            r"^\s+" + r"FINAL EFP ENERGY      =" + r"\s+" + NUMBER + r"\s*",
+            outtext,
```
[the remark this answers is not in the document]
zachglick: Thanks for the info, fixed

qcengine/programs/gamess/runner.py
zachglick: I suppose I should rename that variable from `molecule.extras["efp"]` to `molecule.extras["efp_molecule"]` in order to be more consistent with the blueprint. In the blueprint, `molecule.extras["efp_molecule"]` is a schema-like dictionary defining all of the fragment locations. In this PR, `molecule.extras["efp_molecule"]` is instead expected to be a GAMESS-ready string, which is directly inserted into the input file. 

This is a short-tem solution. Eventually, we need an input-parsing function that takes in your `efp_molecule` and converts it to the GAMESS-ready string. I held off on doing this right away because I wasn't sure how final the blueprint was.

I added an example script to the PR description that demonstrates how you'd run MAKEFP and then get an EFP interaction energy. This will give you a better idea of what the `efp_molecule` string looks like in this PR.

qcengine/programs/gamess/harvester.py
zachglick: Noted, that's going to be a headache later on

qcengine/programs/gamess/harvester.py
loriab: If you could get away with `CURRENT ENERGY` only that'd be better, but if `CURRENT REFERENCE ENERGY` needed, fine. It's questionable b/c it's an interaction energy like SAPT rather than a total energy.

PR #261, opened 2020-07-31
improving get_version for Molpro and Q-Chem

loriab:
## Changelog description
- [x] at least won't throw xml errors in the presence of an old molpro when probing presence/version
- [x] modernize Provenance for Q-Chem (have to cheat for qcer tests)
- [x] for mp2 energies, get Q-Chem passing stdsuite
- [x] switch to extracting Q-Chem version from output file, not `qchem -h` since the last isn't available from a source install
- [x] catch up on `isort`
- [x] lower qchem requirements to v5.1 (works at GT). former v5.2 was simply the version at which the harness was developed.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/qchem.py
+++ b/qcengine/programs/qchem.py
@@ -11,9 +11,11 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import numpy as np
 from qcelemental import constants
-from qcelemental.models import AtomicInput, AtomicResult, Molecule
+from qcelemental.models import AtomicInput, AtomicResult, Molecule, Provenance
 from qcelemental.molparse import regex
-from qcelemental.util import parse_version, provenance_stamp, safe_version, which
+from qcelemental.util import parse_version, safe_version, which
+
+from qcengine.config import TaskConfig, get_config
 
 from ..exceptions import InputError, UnknownError
 from ..util import disk_files, execute, popen, temporary_directory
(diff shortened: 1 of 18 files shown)


qcengine/programs/qchem.py
```diff
@@ -59,29 +61,45 @@ def _get_qc_path(self, config: Optional["TaskConfig"] = None):
         return paths
 
     def get_version(self) -> str:
+        # excuse missing Q-Chem for QCEngineRecords tests
+        found = self.found()
+        if not found:
+            return safe_version("0.0.0")
+
         self.found(raise_error=True)
```
dotsdl, line 69: Perhaps remove this one since we have the above now?

loriab: the `found()`s? second is normal behavior and has the `raise_error` behavior. first is the crazy escape path for qcer.

dotsdl: But will the second ever have any effect? If `qchem` isn't found, the first will return the `safe_version("0.0.0")` call and exit the method. If `qchem` is found, it will sail through? I might be misunderstanding the subtleties of `qchem` here...

loriab: yes, you're right. the return v0.0 was newly needed to appease qcer after I made qchem run for version. but the second with raise_error is the correct harness behavior. proper solution would be for the qcer tests to set an envvar that could be used to narrowly trigger the v0.0 special behavior. I'd like to keep the second `found` in memory of what we should return to.

I don't fully understand the subtleties of the qchem harness myself. it's one of those programs that settled into an idiosyncratic groove long ago, and it looks like DGAS had to add extra hoops to to stuff it into a harness.

PR #263, opened 2020-08-10
NWChem: Add ability to turn off autoz

WardLT:
## Description

NWChem's autoz function generates a Z matrix and [is known to have issues with large molecules](https://nwchemgit.github.io/FAQ.html#autoz-fails-to-generate-valid-internal-coordinates-now-what). 
This PR adds an option to allow QCEngine to run NWChem with autoz turned off. 

## Changelog description

Added option for disabling autoz in NWChem

## Status

- [x] Code base linted
- [x] Ready to go


qcengine/programs/nwchem/runner.py
```diff
@@ -154,6 +154,9 @@ def build_input(
         molcmd, moldata = input_model.molecule.to_string(dtype="nwchem", units="Bohr", return_data=True)
         opts.update(moldata["keywords"])
 
+        if opts.pop("noautoz", False):
```
loriab, line 157: what do you think of `geometry__noautoz` for the option name? it'd help with consistency elsewhere.

WardLT: I like that better. Makes sense along with the other NWChem settings. I'll make the change :)

PR #264, opened 2020-08-19
Turbomole energy fix, Hessian, Memory managment

eljost:
##  Description
Receiving the paper Draft led me to revisit the TurbomoleHarness. As my own code [pysisyphus](https://github.com/eljost/pysisyphus) also supports QCEngine I noticed a bug. In gradient/Hessian calculations the current_energy wasn't set in the output properties, only the repsective energy derivative was returned ... Now the the runner uses "build_atomicproperties" and "build_out" from qcvar_identities_resources to be more consistent.

Furthermore I added support for Hessian calculations (analytical via aoforce for HF/DFT and numerical via NumForce via ricc2). Corresponding tests are provided. The Harness now also has basic memory management.

## Changelog description
Fixed missing current_energy in gradient calculation, added basic memory management and support for Hessian calculations.

## Status
- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_turbomole.py
+++ b/qcengine/programs/tests/test_turbomole.py
@@ -19,6 +19,22 @@ def h2o():
     return mol
 
 
+@pytest.fixture
+def h2o_ricc2_def2svp():
+    """NumForce calls only make sense for stationary points. So this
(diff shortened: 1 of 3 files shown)


loriab (approved):
Looks great, and I’m sorry for the delay.

PR #266, opened 2020-08-28
Psi4 error reporting pre-empted by KeyError

dotsdl:
## Description

Currently, useful error reporting with `psi4` under the `Psi4Harness` can be lost due to a missing `"success"` key on `output_data`. This PR makes attempted access of this key return `False` if not present.

Credit to @dgasmith for pointing this out.

## Changelog description

Error messages from `Psi4Harness` no longer swallowed by `KeyError`.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -164,7 +164,7 @@ class Psi4Harness(ProgramHarness):
                         else:
                             output_data["extras"]["qcvars"] = local_qcvars
 
-                    if output_data["success"] is False:
+                    if output_data.get("success", False) is False:
                         if "error_message" not in output_data["error"]:
                             error_message = output_data["error"]
                             error_type = "internal_error"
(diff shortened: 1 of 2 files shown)


qcengine/programs/psi4.py
```diff
@@ -164,7 +164,7 @@ def compute(self, input_model: "AtomicInput", config: "TaskConfig") -> "AtomicRe
                         else:
                             output_data["extras"]["qcvars"] = local_qcvars
 
-                    if output_data["success"] is False:
+                    if output_data.get("success", False) is False:
```
dgasmith, line 167: To be safe you may want to protect the below dict getters in a similar fashion `error_message = output_data.get("error", "Unknown error, error message is not found")` or whatnot.

dotsdl: Ah thanks! Will add this now.

dotsdl:
Looks like we're also being hit by the latest black update here.

loriab:
The CI failure is for the new code, right?

On the `black` changes, I think let's try to merge some PRs, then do a separate one for formatting.

dotsdl:
@loriab ah yes, you're right. It looks like my changes to the error message pathways mangled what is expected in those tests. I'll dive into these to find what's going awry.

dotsdl:
Found the issue that was causing tests to fail; was an inconsistency in the error handling logic. Fixed and moved to its own function so we don't have the same delicate decision tree in two places. 

Should be good to go if all tests pass on Travis!

dotsdl:
@loriab this one is ready to go! Should improve our error reporting from psi4 failures substantially!

dgasmith (approved):
Nice looks pretty bullet proof. It's amazing what you end up with to catch all of the edge case at QCArchive scale.

loriab:
great, thanks, this was the PR I was keen to get in!

PR #271, opened 2020-09-23
Update index.rst

bgpeyton:
## Description

Changing the initial docs example to pass in `model` to `qcng.compute`, instead of `input` (which hasn't been defined).

## Changelog description
- [x] edit index.rst
## Status
- [x] Ready to go


loriab (approved):
good catch

PR #273, opened 2020-10-11
added method field to terachem harness

stvogt:
## Description
The terachem harness was not working do to the absence of a method filed in 
build_input, so I added the missing line.

## Status

- [ ] Code base linted
- [ ] Ready to go


loriab (approved):
thanks for the fix

PR #274, opened 2020-10-11
Updates (Entos -> Qcore) bindings to new python interface

dgasmith:
The Entos package was renamed to Qcore and has since received a python interface.

The only potentially controversial change is the "xTB" method was renamed to "GFN1"/"GFN0" to better reflect the current state of affairs. Since `method=XTB` is a bit like saying `method=DFT`.

@dotsdl This should be good to go with OpenFF datasets.

## Status

- [x] Code base linted
- [x] Ready to go


qcengine/programs/qcore.py
```diff
@@ -0,0 +1,275 @@
...
+            properties["calcinfo_nbeta"] = properties["calcinfo_nalpha"]
+        else:
+            properties["calcinfo_nbeta"] = np.sum(wavefunction["scf_occupations_b"] > 0)
+
+        output_data["properties"] = properties
+
+        output_data["schema_name"] = "qcschema_output"
+        output_data["success"] = True
+
+        return AtomicResult(**output_data)
```
loriab, line 246: anything special needed to merge input and result extras field?

qcengine/programs/qcore.py
```diff
@@ -0,0 +1,275 @@
...
+            energy_command = "dft"
+        else:
+            energy_command = method.lower()
+
+        # Check method is supported
+        energy_commands = [key for key in self._energy_commands]
+        if energy_command not in energy_commands:
+            raise InputError(f"The energy method {method} is not implemented in QCEngine for qcore.")
+
+        return energy_command
```
loriab, line 264: fn unused?

loriab (commented):
neat, thanks for revising!

qcengine/programs/qcore.py
dgasmith: Shouldn't be.

qcengine/programs/tests/test_standard_suite_hf.py
```diff
@@ -83,10 +83,10 @@ def test_sp_hf_rhf(program, basis, keywords, h2o):
         ),
         pytest.param("cfour", "aug-pvdz", {"reference": "uhf"}, marks=using("cfour")),
         pytest.param(
-            "entos",
+            "qcore",
             "aug-cc-pVDZ",
             {"ansatz": "u", "coulomb_method": "direct_4idx", "exchange_method": "direct_4idx"},
-            marks=using("entos"),
+            marks=using("qcore"),
```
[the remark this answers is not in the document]
dgasmith: `test_standard_suite.py` looks like MP2/CCSD(T) only? This PR does not include MP2 and Qcore does not have CCSD(T).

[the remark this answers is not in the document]
loriab: Good point. I thought I could get that patched up in short order but it seems not. No use holding this up.

loriab (approved):
lgtm, merge at will!

PR #277, opened 2020-11-02
Add adcc harness

maxscheurer:
## Description

This PR adds a harness for [adcc](https://adc-connect.org/v0.15.1/index.html), a Python-driven toolkit for excited states within the algebraic-diagrammatic construction scheme for the polarization propagator.

`adcc` requires an SCF driver to run first (e.g. Psi4, pyscf) to provide the SCF reference state, integrals, etc. throughout the rest of the computation.

CC @mfherbst

## ToDo

- [x] robust SCF wrapper
- [x] proper input keyword handling (including defaults etc.)
- [x] error handling
- [x] add tests

## Changelog description

Add a harness for [adcc](https://adc-connect.org/v0.15.1/index.html)

## Status

- [x] Code base linted
- [x] Ready to go


qcengine/programs/adcc.py
```diff
@@ -0,0 +1,155 @@
...
+        return False
+
+    def get_version(self) -> str:
+        self.found(raise_error=True)
+        import adcc
+
+        # candidate_version = self.version_cache[which_prog]
+        candidate_version = adcc.__version__
+
+        return candidate_version
```
loriab, line 79: https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/xtb.py#L48-L58 can be imitated directly to get the `version_cache`

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,155 @@
...
+        xyz = "\n".join(xyz.split("\n")[2:])
+        scfres = adcc.backends.run_hf(
+            backend=None,  # auto-select available backend
+            xyz=xyz,
+            multiplicity=mol.molecular_multiplicity,
+            charge=mol.molecular_charge,
+            basis=model.basis,
+            #   conv_tol=model.scf_conv,
+            #   max_iter=,
+        )
```
mfherbst, line 91: Hmm. I'm not sure about this. The way we designed it this was supposed to be "internal" and for testing only. To be fair the interface is a bit too naive from my point of view to start using it in a third-party package ... of course we can flesh it out a bit and promote it to something proper if that's needed.

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,155 @@
...
+            Passed on to control negative return between False and ModuleNotFoundError raised.
+        Returns
+        -------
+        bool
+            If adcc is found, returns True.
+            If raise_error is False and adcc is missing, returns False.
+            If raise_error is True and adcc is missing, the error message is raised.
+        """
+
+        adcc_found = is_module_available("adcc")
```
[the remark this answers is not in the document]
maxscheurer: Done.

qcengine/programs/adcc.py
maxscheurer: Done.

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,155 @@
...
+                qcvars["MP2 DIPOLE"] = mp.dipole_moment(2)
+            # transition properties
+            qcvars[f"{NAME} TRANSITION DIPOLES (LEN)"] = state.transition_dipole_moment
+            qcvars[f"{NAME} TRANSITION DIPOLES (VEL)"] = state.transition_dipole_moment_velocity
+            qcvars[f"{NAME} OSCILLATOR STRENGTHS (LEN)"] = state.oscillator_strength
+            qcvars[f"{NAME} OSCILLATOR STRENGTHS (VEL)"] = state.oscillator_strength_velocity
+            qcvars[f"{NAME} ROTATIONAL STRENGTHS (VEL)"] = state.rotatory_strength
+            # state properties
+            qcvars[f"{NAME} STATE DIPOLES"] = state.state_dipole_moment
+        return qcvars
```
mfherbst, line 145: That's duplicate code with psi4 if I remember correctly. Should we not just put this into adcc if that's needed from both places?

mfherbst (commented):
Nice! Thanks for integrating this.

qcengine/programs/adcc.py
maxscheurer: Having some kind of wrapper code would be nice I guess. Otherwise, too much of the code here would be checking for program availability etc... I think either the `run_hf` needs to be improved (run from schema) or we limit ourselves to Psi4 for now. What do you think @loriab?

qcengine/programs/adcc.py
maxscheurer: The code is quite similar to what is done in Psi4, right.
In Psi4, it's mixed with some print-out...
Could be a function on `ExcitedStates`, e.g., `to_qcvars` or `to_dict`, and then inside Psi4 the variables would be set on the wave function.

qcengine/programs/adcc.py
loriab: I'm hoping the external HF is a medium-term solution, to be replaced (or at least alternate) by atomicresult.wavefunction and an unknown ERI transport. `import psi4` (or `import pyscf` -- no particular reason we dn't have a wrapper for that) is fine if you don't think `adcc.run_hf` is advanced enough. it'd be good to incorporate the helper program into the `found` detection to avoid aggravation. nwchem has an example as it requires networkx.

qcengine/programs/adcc.py
mfherbst: Yes that, but since it is pretty specific to this ecosystem perhaps a free function in just another file in `adcc`?

qcengine/programs/adcc.py
mfherbst: For now I would feel more comfortable with using psi4 or pyscf, but I agree that having a nice interface to pass such data would be nice in the long run. What we have in adcc works kind of ok, but of course it requires people to actually implement it for every backend. Would be good if the community agreed on a standard.

qcengine/programs/adcc.py
maxscheurer: I think we should finalize the routine here until we're happy with it and then transfer it adcc-side. Otherwise we probably make some mistake and have to create several adcc releases until it works fine 😉

mfherbst: That sounds reasonable.

maxscheurer:
I think the next step would be to discuss how/which input options are transported to the `run_adc` function, and what the preferred naming convention in `input_model.keywords`, @loriab.

Our documentation of `run_adc` can be found [here](https://adc-connect.org/v0.15.1/topics.html#adcc.run_adc).

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Runs adcc
+        """
+        self.found(raise_error=True)
+        import adcc
+        import psi4
+
+        mol = input_model.molecule
+        model = input_model.model
+        kws = input_model.keywords
```
loriab, line 88: > I think the next step would be to discuss how/which input options are transported to the run_adc function, and what the preferred naming convention in input_model.keywords, @loriab.

Pretty much any "Parameters" or "Other Parameters" from your docs should be named _as is_ for `atomicinput.keywords`. Exceptions are items handled elsewhere in the schema like method (`model.method`), data_or_matrix (too involved), output (`protocols.stdout`). The idea is that for a harness, the user should only have to perform data layout translation, not keyword name translation.

maxscheurer: Well, in that case, I suggest to pass `**input_model.keywords` to `run_adc`, because then we don't have to take care of any safe-guarding in the harness, `adcc` will do it itself. `method` is already set by `model.method`, as you mentioned, and `data_or_matrix` will be forced to `wfn`.

mfherbst: What's the convention wrt. to sanity checks and input handling? Some of the keywords are exclusive for example. We should have been pretty careful to die with a ValueError in that case, but that might need some translation.

qcengine/programs/adcc.py
```diff
@@ -74,27 +81,34 @@ def compute(self, input_model: "AtomicInput", config: "TaskConfig") -> "AtomicRe
...
-            backend=None,  # auto-select available backend
-            xyz=xyz,
-            multiplicity=mol.molecular_multiplicity,
-            charge=mol.molecular_charge,
-            basis=model.basis,
-            #   conv_tol=model.scf_conv,
-            #   max_iter=,
+
+        # TODO: smooth way to do this?
+        psi4_molecule = psi4.geometry(mol.to_string(dtype="psi4") + "symmetry c1")
```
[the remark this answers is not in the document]
maxscheurer: Mhm I keep getting:
```
in from_schema
    molschema.get("schema_name", "").startswith("qc_schema")
E   AttributeError: 'NoneType' object has no attribute 'get'```

File: qcengine/programs/adcc.py (hunk 5 above)
Comment (maxscheurer) [reply qcengine/programs/adcc.py:88]: From my experience, `adcc` is pretty strict with the keyword sanity checks. What exactly do you mean by 'translation'?

Comment (mfherbst) [reply qcengine/programs/adcc.py:88]: I could imagine (have not checked) that QCEngine wants error handling be done in a specific way (like use a special exception, set some flag etc.)

File: qcengine/programs/adcc.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (loriab) [reply qcengine/programs/adcc.py:91]: try `qcel.molparse.from_schema({**kmol.dict(), "fix_symmetry": "c1"})` -- I forgot that `update` is in-place.

also please add a table line in docs/source/program_overview.rst

File: qcengine/programs/adcc.py (hunk 5 above)
Comment (loriab) [reply qcengine/programs/adcc.py:88]: right, if possible, it'd be good to sort adcc errors into input, computer resource, mysterious, etc. like https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/psi4.py#L249-L263

File: qcengine/programs/adcc.py [hunk 7]
```diff
@@ -0,0 +1,159 @@
...
+
+        mol = input_model.molecule
+        model = input_model.model
+        kws = input_model.keywords
+
+        # TODO: smooth way to do this?
+        psi4_molecule = psi4.geometry(mol.to_string(dtype="psi4") + "symmetry c1")
+        psi4.core.clean()
+        psi4.core.be_quiet()
+        psi4.set_options(
```
mfherbst, line 94: Just a note, because it came to my mind: We should set the psi4 SCF tolerance in line with the adcc tolerance (i.e. to ensure that https://github.com/adc-connect/adcc/blob/55436b4bbf6c8a716c99add3f15adde341acb9ff/adcc/workflow.py#L338 is not triggered ;)

qcengine/programs/adcc.py
mfherbst: Ok there are a few things we can do. We are pretty generic on `ValueError`s I just saw, meaning that a `ValueError` *can* signal that it's a problem with the input parameters, but it can also be something more internal (and thus a bug). I think the best way to solve this is is to patch adcc and ensure that during the parsing phase some specialised exception is thrown. I'll make a patch in adcc and then we can use that here and redirect to `InputError`. The C++ layer exceptions are mostly translated to `RuntimeError` and errors in the python layer use pretty much the standard python exceptions. I would maybe use UnknownError or RandomError for those?

maxscheurer:
@mfherbst, I gave you write access to my fork, so you can also make changes 😉

qcengine/programs/adcc.py
mfherbst: See https://github.com/adc-connect/adcc/pull/93.

qcengine/programs/adcc.py
mfherbst: done.

qcengine/programs/adcc.py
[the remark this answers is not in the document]
mfherbst: I added the code suggestion from @loriab.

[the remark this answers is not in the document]
maxscheurer: It does not work for me... with your code I get:
```
Traceback (most recent call last):
  File "test_adcc.py", line 15, in <module>
    ret = qcng.compute(inp, "adcc", raise_error=True,
  File "/Users/maxscheurer/Projects/qcengine/qcengine/compute.py", line 91, in compute
    output_data = executor.compute(input_data, config)
  File "/Users/maxscheurer/Projects/qcengine/qcengine/programs/adcc.py", line 102, in compute
    _, wfn = psi4.energy("HF", return_wfn=True, molecule=psi4_molecule)
  File "/Users/maxscheurer/miniconda3/envs/grad/lib/python3.8/site-packages/psi4/driver/driver.py", line 536, in energy
    molecule.update_geometry()
AttributeError: 'dict' object has no attribute 'update_geometry'
```

Does it work on your machine, @mfherbst?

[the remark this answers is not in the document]
mfherbst: I have not tested :smile: ... my python setup is a bit gargled at the moment, so I hoped for CI to catch it. I think what I did now might fix it.

maxscheurer:
Any hints on how to fix the CI, @loriab?

loriab:
> Any hints on how to fix the CI, @loriab?

I think try working from psi-nightly.yaml rather than psi.yaml. The latter is also serving as a "minimum working versions" CI and so has some `=` instead of `>=`. I think that's what's making it select psi4.*dev3xx rather than psi4.*dev10xx.

mfherbst:
Hooray! CI passes. I guess what is left in terms of functionality is to extend the data / property extraction and get that into adcc? Then of course proper tests to wrap it up?

maxscheurer:
What other properties do you have in mind? I think what is exported right now is equivalent to the `psi4` vars.

Regarding tests, I would suggest small functionality tests like in Psi4, what do you think?

loriab:
if these are the kinds of molecules you'd be using for testing anyway, I suggest the following mol/ref/basis combinations: https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/standard_suite_ref.py#L9-L77 . If they don't serve your purposes, disregard.

maxscheurer:
Thanks, @loriab! I was thinking we could also use the molecules/etc that we already have reference data for, but I have no strong opinion on this. If it's better to use the `standard_suite_ref` molecules, I can also generate data for those.

mfherbst:
Since we are not really testing the ADC implementation, I think there is no real reason to reuse the molecules we have data for (which where chosen to provide test cases for specific aspects of our implementation). So I don't really have a strong opinion either. Perhaps we should just use whatever is easiest, i.e. the molecules we have data for? Or is it because of maintainability beneficial to stick to the `standard_suite_ref` molecules @loriab? In that case I would say we do the latter.

mfherbst:
Sorry about the long delay in continuing this PR. We had a few issues with our deployment moving from Travis to GHA. It turned out that the easiest solution was to work around the remaining proprietary bits of adcc and make the code completely open-source, so now we can hopefully finish this up and get it in.

devtools/conda-envs/adcc.yaml
```diff
@@ -0,0 +1,27 @@
+name: test
+channels:
+  - adcc
+  - psi4/label/dev
+  - conda-forge
+dependencies:
+  - adcc=0.15.6
```
loriab, line 7:
```suggestion
  - adcc>=0.15.6
```

devtools/conda-envs/adcc.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - psi4
+  - dftd3 3.2.1
+  - mp2d >=1.1
```
loriab, line 10:
```suggestion
```

you're not using these with psi4, right?

qcengine/programs/tests/test_adcc.py
```diff
@@ -0,0 +1,31 @@
...
+# TODO: test against some ref values...
+@using("adcc")
+def test_run(h2o):
+    inp = qcel.models.AtomicInput(
+        molecule=h2o, driver="properties", model={"method": "adc2", "basis": "sto-3g"}, keywords={"n_singlets": 3}
+    )
+    ret = qcng.compute(inp, "adcc", raise_error=True, local_options={"ncores": 1})
+
+
+# TODO: expand...
```
loriab, line 31: A check for `success` and `return_result` and  `properties.scf_total_energy` would be good.

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,126 @@
...
+            }
+        )
+        _, wfn = psi4.energy("HF", return_wfn=True, molecule=psi4_molecule)
+        adcc.set_n_threads(config.ncores)
+        compute_success = False
+        try:
+            adcc_state = adcc.run_adc(wfn, method=model.method, **input_model.keywords)
+            compute_success = adcc_state.converged
+        except ValueError as e:
+            # TODO Once the new adcc is realeased switch to "adcc.InputError as e"
```
loriab, line 106: ready for this?

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,126 @@
...
+        output_data["success"] = compute_success
+
+        if compute_success:
+            output_data["properties"] = AtomicResultProperties()
+            output_data["return_result"] = adcc_state.excitation_energy
+
+            extract_props = input_model.driver == "properties"
+            output_data["extras"]["qcvars"] = adcc_state.to_qcvars(recurse=True, properties=extract_props)
+
+        provenance = Provenance(creator="adcc", version=self.get_version()).dict()
```
loriab, line 122: I think Provenance has a `routine` field, too, that's required somewhere. Feel free to choose something generic like `routine="adcc"`.

loriab (commented):
Sorry for the great delay. I think it's in nice shape and can be merged after a few update tweaks and test comparison values. Thank you for the harness!

mfherbst:
@maxscheurer I'm on it.

mfherbst:
@maxscheurer Seems like we are not exporting any of the computed properties. Any idea why?

maxscheurer:
I think the computed data should be in `ret["extras"]["qcvars"]`, no?

mfherbst:
Hmm ok, but I'm not sure that is all that reasonable, don't you think?

maxscheurer:
I think since these are "custom" variables (i.e. not following QCSchema), this is the intended spot for the results to live right now. What do you think, @loriab?

mfherbst:
Cool, thanks @maxscheurer !

qcengine/programs/tests/test_adcc.py
```diff
@@ -0,0 +1,39 @@
...
+    ref_hf_energy = -74.45975898670224
+    ref_mp2_energy = -74.67111187456267
+    assert ret["success"] is True
+
+    ret_qcvars = ret["extras"]["qcvars"]
+
+    assert ret_qcvars["EXCITATION KIND"] == "SINGLET"
+    assert compare_values(ref_excitations, ret["return_result"])
+    assert compare_values(ref_hf_energy, ret_qcvars["HF TOTAL ENERGY"])
+    assert compare_values(ref_mp2_energy, ret_qcvars["MP2 TOTAL ENERGY"])
```
loriab, line 39:
```suggestion
    assert compare_values(ref_mp2_energy, ret_qcvars["MP2 TOTAL ENERGY"])
    assert compare_values(ref_hf_energy, ret["properties"]["hf_total_energy"])
    assert compare_values(ref_mp2_energy, ret["properties"]["mp2_total_energy"])
```
would these work, too?

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,125 @@
...
+        except Exception as e:
+            raise UnknownError(str(e))
+
+        input_data = input_model.dict(encoding="json")
+        output_data = input_data.copy()
+        output_data["success"] = compute_success
+
+        if compute_success:
+            output_data["properties"] = AtomicResultProperties()
+            output_data["return_result"] = adcc_state.excitation_energy
```
loriab, line 116: this is returning a list or array right now? I think that's working b/c array is allowed for gradient, but pretty much QCA is expecting driver energy to have a float in return_result, gradient to have gradient array, etc. The driver `properties` can return a dict. Would it be reasonable in the absence of a target state to put `adcc_state.excitation_energy[0]` in this field? I assume the full excitation_energy list is in qcvars somewhere.

loriab:
add yourselves to codemeta, too, if you both like. here's a start:

```
        {
            "@type": "Person",
            "givenName": "Maximilian",
            "familyName": "Scheurer",
            "affiliation": "Interdisciplinary Center for Scientific Computing, Heidelberg University",
            "@id": "https://orcid.org/0000-0003-0592-3464",
            "identifier": "https://github.com/maxscheurer"
        },
        {
            "@type": "Person",
            "givenName": "Michael",
            "additionalName": "F.",
            "familyName": "Herbst",
            "affiliation": "Interdisciplinary Center for Scientific Computing, Heidelberg University",
            "@id": "https://orcid.org/0000-0003-0378-7921",
            "identifier": "https://github.com/mfherbst"
        },
```

qcengine/programs/tests/test_adcc.py
maxscheurer: I think @mfherbst has tried these, but without success. I don't know whether the auto-conversion of results you mentioned should have these covered or not.

qcengine/programs/adcc.py
maxscheurer: Yes, the full `excitation_energy` list is in qcvars. For me it would be okay to just return the lowest excitation energy. Is there any "precedence" for this with other excited state methods?

qcengine/programs/tests/test_adcc.py
mfherbst: Indeed I tried with the properties, but I think the tests already failed in the "excitation kind" assertion, so could be that this actually works.

qcengine/programs/adcc.py
mfherbst: We could also return the MP2 energy. I think I did that in Psi4 for the "energy" version of ADC.

maxscheurer: Done. Returning MP ground state energy now.

qcengine/programs/tests/test_adcc.py
maxscheurer: Fixed it by calling the automatic qcvars->properties conversion routine.

qcengine/testing.py
```diff
@@ -140,6 +140,7 @@ def get_job(self):
 
 # Figure out what is imported
 _programs = {
+    "adcc": is_program_new_enough("adcc", "0.15.1"),
```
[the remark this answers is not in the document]
maxscheurer: Done.

maxscheurer:
I've addressed the comments and added myself to codemeta. @mfherbst please add yourself because I don't know your affiliation 😄

mfherbst:
> I don't know your affiliation

For the next 2 weeks still France, then Aachen ;). I'll add what it is now and update later.

maxscheurer:
I'm happy with the harness now 😄 You can squash-merge this, if you want, @loriab.

codemeta.json
```diff
@@ -289,6 +289,23 @@
...
+            "affiliation": "Interdisciplinary Center for Scientific Computing, Heidelberg University",
+            "@id": "https://orcid.org/0000-0003-0592-3464",
+            "identifier": "https://github.com/maxscheurer"
+        },
+        {
+            "@type": "Person",
+            "givenName": "Michael",
+            "additionalName": "F.",
+            "familyName": "Herbst",
+            "affiliation": "Inria Paris and CERMICS, École des Ponts ParisTech",
```
loriab, line 306:
```suggestion
            "affiliation": "Inria Paris and CERMICS, \u0137cole des Ponts ParisTech",
```

qcengine/programs/tests/test_adcc.py
```diff
@@ -0,0 +1,40 @@
...
+
+    ref_excitations = np.array([0.0693704245883876, 0.09773854881340478, 0.21481589246935925])
+    ref_hf_energy = -74.45975898670224
+    ref_mp2_energy = -74.67111187456267
+    assert ret["success"] is True
+
+    qcvars = ret["extras"]["qcvars"]
+
+    assert qcvars["EXCITATION KIND"] == "SINGLET"
+    assert compare_values(ref_mp2_energy, ret["return_result"])
```
loriab, line 37:
```suggestion
    assert compare_values(ref_excitations[0], ret["return_result"])
    assert compare_values(ref_hf_energy, ret["properties"]["SCF TOTAL ENERGY"])
    assert compare_values(ref_mp2_energy, ret["properties"]["MP2 TOTAL ENERGY"])
```

qcengine/programs/adcc.py
```diff
@@ -0,0 +1,133 @@
...
+        output_data = input_data.copy()
+        output_data["success"] = compute_success
+
+        if compute_success:
+            mp = adcc_state.ground_state
+            # MP(3) energy for CVS-ADC(3) calculations is still a missing feature in adcc
+            # ... we store this variant here to be able to fall back to MP(2) energies.
+            is_cvs_adc3 = adcc_state.method.level >= 3 and mp.has_core_occupied_space
+            mp_energy = mp.energy(adcc_state.method.level if not is_cvs_adc3 else 2)
+            output_data["return_result"] = mp_energy
```
loriab, line 121:
```suggestion
            output_data["return_result"] = adcc_state.excitation_energy[0]
```

I think this would be better, if it's ok with you. There isn't much excited state precedent to follow, and Psi4 should probably change to match, but it's fairly rigorous in QCA that `return_result` corresponds with `model.method`, so in the absence of a target state, lowest excited best. If there was a CIS, I'd push lowest excited, too, rather than HF GS energy.

loriab (commented):
One more change, if you agree. Looks great

codemeta.json
maxscheurer: Done.

qcengine/programs/adcc.py
maxscheurer: Done.

qcengine/programs/tests/test_adcc.py
maxscheurer: Done.

codemeta.json
maxscheurer: I think "\u00C9" is correct, though 😄

mfherbst: Yes please, no funny ks.

loriab: looks like they're the same https://en.wikipedia.org/wiki/List_of_Unicode_characters. I chose from the column the psi4 codemeta used. I don't actually recall what software balked at direct characters anyways at this point. :-)

loriab (approved):
looks great, thank you for the harness

codemeta.json
maxscheurer:
```
>>> x = u"\u0137"
>>> x
'ķ'
```
on my machine 😄

mfherbst: Same here :smile:

PR #280, opened 2020-12-11
Add error correction harness

WardLT:
## Description
Adds a revised error correction harness, as discussed in #210.

The new version of the error correction harness works by catching special exceptions and then updating the "keywords" 
of an AtomicInput to attempt to correct for the error.

To make this happen, I have created as special exception class, a Harness with logic to handle the exception class, and made an example with NWChem

## Changelog description
Implemented the ability for harness to autocorrect common errors

## Status

- [x] Code base linted
- [x] Ready to go


loriab (approved):
looks great to me. somehow, this reads as clearer to me than the previous simpler approach. thanks!

PR #287, opened 2021-02-11
coverage tweaks

loriab:
## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/adcc.py
+++ b/qcengine/programs/adcc.py
@@ -84,7 +84,6 @@ class AdccHarness(ProgramHarness):
         model = input_model.model
         conv_tol = input_model.keywords.get("conv_tol", 1e-6)
 
-
         if input_model.driver not in ["energy", "properties"]:
             raise InputError(f"Driver {input_model.driver} not implemented for ADCC.")
 
--- a/qcengine/programs/adcc.py
+++ b/qcengine/programs/adcc.py
@@ -1,14 +1,14 @@
 """
 Calls adcc
 """
-from typing import Dict, TYPE_CHECKING
+from typing import TYPE_CHECKING, Dict
 
-from qcelemental.util import safe_version, which_import
 from qcelemental.models import AtomicResult, Provenance
-from .qcvar_identities_resources import build_atomicproperties
+from qcelemental.util import safe_version, which_import
 
-from .model import ProgramHarness
 from ..exceptions import InputError, UnknownError
+from .model import ProgramHarness
+from .qcvar_identities_resources import build_atomicproperties
 
 if TYPE_CHECKING:
     from qcelemental.models import AtomicInput
@@ -84,7 +84,7 @@ class AdccHarness(ProgramHarness):
         model = input_model.model
         conv_tol = input_model.keywords.get("conv_tol", 1e-6)
 
-        need_one_of_states = {'n_spin_flip', 'n_triplets', 'n_states', 'n_singlets'}
+        need_one_of_states = {"n_spin_flip", "n_triplets", "n_states", "n_singlets"}
         if not (need_one_of_states & set(input_model.keywords)):
             input_model.keywords["n_states"] = 1
 
--- a/qcengine/programs/adcc.py
+++ b/qcengine/programs/adcc.py
@@ -84,6 +84,16 @@ class AdccHarness(ProgramHarness):
         model = input_model.model
         conv_tol = input_model.keywords.get("conv_tol", 1e-6)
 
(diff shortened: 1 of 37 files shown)


loriab:
@robertodr, for MRChem, I'm adding it to the sanity check tests and came across this:
```
        except Exception as e:
>           raise InputError(f"Failure preparing input to MRChem\n {str(e)}")
E           qcengine.exceptions.InputError: Failure preparing input to MRChem
E            
E           Error occurred when merging:
E           - At user['world_prec']:
E             Keyword 'world_prec' is required but has no value.
```
From https://mrchem.readthedocs.io/en/latest/users/user_ref.html, I gather this is intended behavior. Is a default out of the question so that a plain molecule+model can run?

qcengine/programs/adcc.py
```diff
@@ -84,6 +84,16 @@ def compute(self, input_model: "AtomicInput", config: "TaskConfig") -> "AtomicRe
...
 
+        need_one_of_states = {"n_spin_flip", "n_triplets", "n_states", "n_singlets"}
+        if not (need_one_of_states & set(input_model.keywords)):
+            input_model.keywords["n_states"] = 1
+
+        if input_model.driver not in ["energy", "properties"]:
+            raise InputError(f"Driver {input_model.driver} not implemented for ADCC.")
+
+        if not input_model.model.basis:
+            raise InputError("Method must contain a basis set.")
```
loriab, line 95: @maxscheurer @mfherbst, I was adding ADCC to the basic checks over at test_harness_canonical.py and needed to make a few adjustments so that a molecule+model is sufficient. are these ok? should n_states be larger?

mfherbst: Depends what you want ... I suppose if you go for excited states you rarely only want one. Also just asking for one can be numerically unreliable. I'd go for at least 3.

loriab: thanks, will change it to 3.

I'm guessing the no-keywords case is targeting the user who's at the vague "I want the exited state" stage, probably meaning the lowest. So 3 should catch it.

qcengine/tests/test_harness_canonical.py
```diff
@@ -24,6 +24,9 @@
     ("torchani", {"method": "ANI1x"}),
     ("turbomole", {"method": "pbe", "basis": "6-31G"}),
     ("xtb", {"method": "GFN2-xTB"}),
+    ("adcc", {"method": "adc2", "basis": "6-31G"}),
+    ("gcp", {"method": "hf3c"}),
+    ("mrchem", {"method": "blyp"}),
```
robertodr, line 29: > From https://mrchem.readthedocs.io/en/latest/users/user_ref.html, I gather this is intended behavior. Is a default out of the question so that a plain molecule+model can run?

That parameter is (roughly) equivalent to the basis set for GTO software, so a default wouldn't make sense. What about the following?

```suggestion
    ("mrchem", {"method": "blyp",  "keywords": {"world_prec": 1.0e-3}}),
```

qcengine/programs/adcc.py
mfherbst: Yes, I suppose.

But IMHO the code should crash in this case, however, as at that level you're better off being forced to think a little what you actually want to compute before running the calculation. That's the policy we take in adcc. In particular in this default mode of operation there are no spin guarantees made ... so you'll probably end up finding a triplet state as lowest state, which might anyway not be what you are after.

maxscheurer: I'd suggest to set `n_singlets=3` to make the canonical tests work, but throw at least warning like
```Python
warnings.warn("No user-specified parameter for number of states for adcc. Computing 3 singlet states."
```
or something of the sort.

mfherbst: Be careful: This will cause errors for unrestricted references, so we should have an extra branch for that case (where `n_states=3` is the only option).

loriab:
hmm, between the two cases, maybe molecule+model alone is too strict.

Lnaden (approved):
LGTM. 

What is the purpose behind the `fast-fail: false` in the GHA CI? I don't see a reason to disable it in my review here so I just wanted to know the logic. I'll merge since getting it in was time requested and the CI can be tweaked later.

loriab:
Thanks, @Lnaden. I thought `fast-fail: true` was best suited for CI testing the same code on software variations. The way qcng is testing mostly different bits of code in different CI lanes, it got annoying to iterate on solving one lane before seeing errors from another lane (or even switching btwn both depending on what failed first). I can switch back to `true` if you'd rather.

PR #288, opened 2021-02-15
stdsuite calcinfo, HF ene, HF grad

loriab:
Salvage earlier work

## Description
- [x] harvest basic calcinfo from gamess, nwchem, cfour
- [x] silence `gcp.get_version()`
- [x] harvest and test Hartree--Fock energies and gradients
- [x] harvest MP3 energy
- [x] make the xyz geometries for reference calcs match the reference orientation. was working before b/c psi4 and qcdb reoriented them

## Status

- [x] Code base linted
- [x] Ready to go

PR #290, opened 2021-02-24
Mctc-gcp

loriab:
Ignore until #288 rebased away.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/gcp.py
+++ b/qcengine/programs/gcp.py
@@ -168,14 +168,15 @@ class GCPHarness(ProgramHarness):
         molrec = qcel.molparse.from_schema(input_model.molecule.dict())
 
         executable = self._defaults["name"].lower()
+        calldash = {"gcp": "-", "mctc-gcp": "--"}[executable]
 
-        if method == "FILE":
+        if method == "FILE" and executable == "GCP":
             command = [executable, "gcp_geometry.xyz", "-local"]
         else:
-            command = [executable, "gcp_geometry.xyz", "-level", method]
+            command = [executable, "gcp_geometry.xyz", calldash + "level", method]
 
         if input_model.driver == "gradient":
-            command.append("-grad")
+            command.append(calldash + "grad")
 
         infiles = {
             "gcp_geometry.xyz": qcel.molparse.to_string(molrec, dtype="xyz", units="Angstrom", ghost_format=""),
@@ -298,17 +299,10 @@ class MCTCGCPHarness(GCPHarness):
         version = None
         which_prog = which("mctc-gcp")
         if which_prog not in self.version_cache:
-            # option not (yet) available, instead find in help output
             command = [which_prog, "--version"]
             import subprocess
 
-            proc = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            for ln in proc.stdout.decode("utf-8").splitlines():
-                if re.match("mctc-gcp version", ln):
-                    version = ln.split()[2]
-
-            if version is None:
-                raise UnknownError(f"Could not identify GCP version")
-            self.version_cache[which_prog] = safe_version(version)
+            proc = subprocess.run(command, stdout=subprocess.PIPE)
+            self.version_cache[which_prog] = safe_version(proc.stdout.decode("utf-8").strip().split()[-1])
 
         return self.version_cache[which_prog]
--- a/qcengine/programs/gcp.py
+++ b/qcengine/programs/gcp.py
@@ -299,7 +299,7 @@ class MCTCGCPHarness(GCPHarness):
         which_prog = which("mctc-gcp")
         if which_prog not in self.version_cache:
             # option not (yet) available, instead find in help output
-            command = [which_prog, "-version"]
+            command = [which_prog, "--version"]
             import subprocess
 
             proc = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
--- a/qcengine/programs/gcp.py
+++ b/qcengine/programs/gcp.py
@@ -167,10 +167,12 @@ class GCPHarness(ProgramHarness):
         # Need 'real' field later and that's only guaranteed for molrec
         molrec = qcel.molparse.from_schema(input_model.molecule.dict())
 
+        executable = self._defaults["name"].lower()
+
         if method == "FILE":
-            command = ["gcp", "gcp_geometry.xyz", "-local"]
+            command = [executable, "gcp_geometry.xyz", "-local"]
         else:
-            command = ["gcp", "gcp_geometry.xyz", "-level", method]
+            command = [executable, "gcp_geometry.xyz", "-level", method]
 
         if input_model.driver == "gradient":
             command.append("-grad")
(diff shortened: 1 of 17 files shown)


qcengine/programs/gcp.py
```diff
@@ -168,10 +167,12 @@ def build_input(
         # Need 'real' field later and that's only guaranteed for molrec
         molrec = qcel.molparse.from_schema(input_model.molecule.dict())
 
+        executable = self._defaults["name"].lower()
+
         if method == "FILE":
-            command = ["gcp", "gcp_geometry.xyz", "-local"]
+            command = [executable, "gcp_geometry.xyz", "-local"]
```
awvwgk, line 173: At least `mctc-gcp` does not support a `-local` flag and I don't find this flag in the CLI of the old driver as well.
The option to get the parameter file would be `--level file` for both new and old driver.

qcengine/programs/gcp.py
```diff
@@ -168,10 +167,12 @@ def build_input(
...
         else:
-            command = ["gcp", "gcp_geometry.xyz", "-level", method]
+            command = [executable, "gcp_geometry.xyz", "-level", method]
```
awvwgk, line 175: For the new driver I recommend to use two dashes as shown in the help printout and in the manual page. Note that the new driver still supports one dash and the old driver supports as many dashes as you want by the logic the index function here
https://github.com/grimme-lab/gcp/blob/4be7d5acd7bc87161e0b3d89dcff81c7c761f3f7/old/main.f90#L79

The `-l` flag would be neutral for both drivers.

qcengine/programs/gcp.py
```diff
@@ -268,3 +269,46 @@ def parse_output(self, outfiles: Dict[str, str], input_model: "AtomicInput") ->
...
+            raise_msg="Please install via `conda install gcp -c psi4`.",
+        )
+
+    def get_version(self) -> str:
+        self.found(raise_error=True)
+
+        version = None
+        which_prog = which("mctc-gcp")
+        if which_prog not in self.version_cache:
+            # option not (yet) available, instead find in help output
```
awvwgk, line 301: Misleading comment here, this is actually available.
```suggestion
```

awvwgk (commented):
Some comments from my side on this harness.

qcengine/programs/gcp.py
loriab: I found `-local` in help for classic. My guess is that `method == "FILE"` is prep for `.gcppar` file and not presently used. Does this sound right, @hokru? And if the r2SCAN-d* methods need mctc, is it more likely that using a JSON interface is the next step, rather than `.gcppar`?

```
(gcp) psilocaluser@bash:psinet:/psi/gits/QCEngine: (mctcgcp) gcp -h
  ___________________________________________ 
                                              
 |              **  g C P  **                |
 |  a geometrical counterpoise correction    |
 |     H.Kruse J.G.Brandenburg S.Grimme      |
 |          Version 2.02 Nov 2016            |
 |                                           |
  ___________________________________________ 
 
 Please cite work done with this code as:
 H. Kruse, S. Grimme J. Chem. Phys. 136, 154101 (2012)
 DOI: 10.1063/1.3700154 
 For the periodic version, please also cite:
 J. G. Brandenburg, M. Alessio, B. Civalleri, M. F. Peintinger
 T. Bredow, S.Grimme J. Phys. Chem. A 117, 9282-9292 (2013).
 DOI: 10.1021/jp406658y 
 
 
                                   
  *    ----   HELP    ----    *    
                                    
  gcp <coordinates> -level <method>   
                                    
  example:                                 
  -> gcp water.xyz -level dft/631gd     
  -> gcp POSCAR -pbc -level hf/svp 
                                   
                                   
  Valid input coordinates are      
    - TURBOMOLE files (in bohr)    
    - XMOL files (in angstrom)     
    - VASP POSCAR files (in angstrom)
    - CRYSTAL STRUC.INCOOR files (in bohr)
                                   
  available parametrisation for <method>        
       HF/MINIS       DFT/MINIS
       HF/MINIX       DFT/MINIX
       HF/SV          DFT/SV 
       HF/def2-SV(P)  DFT/def2-SV(P)
       HF/def2-SVP    DFT/def2-SVP
       HF/DZP         DFT/DZP
       HF/def-TZVP    DFT/def-TZVP
       HF/def2-TZVP   DFT/def2-TZVP
       HF/631Gd       DFT/631Gd
       HF/def2-TZVP   DFT/def2-TZVP
       HF/cc-pVDZ     DFT/cc-pVDZ
       HF/aug-cc-pVDZ DFT/aug-cc-pVDZ
         -            DFT/SV(P/h,c)
         -            DFT/LANL
                                   
  note: <method> does not need to be          
        capitelized                           
                                   
  command line options:            
     -h                       this print out   
     -pbc                     use periodic version   
     -level <string>    specify <method>   
       if <method> = file or left empty a parameter   
       file from ~/.gcppar.$HOSTNAME will be read in    
     -l                       same as -level   
     -grad                    request gradient 
     -stress                  request cell gradient 
     -noprint                 surpress output   
     -parfile                 print <gcp.param> 
     -local                   use  .gcppar in work-dir 
     -hess                    request hessian  
     -test                    stop after parameter setup  
     -v                       verbose output: print gradient to stdout  
                              instead of gcp_gradient     
     -vasp                    define input as VASP POSCAR file 
     -crystal                 define input as CRYSTAL STRUC.INCOOR file 
                                   
                                   
  *    ----   ****    ----    *    
 
     normal termination of help
```

qcengine/programs/gcp.py
loriab: fix shall appear soon

qcengine/programs/gcp.py
loriab: looks like FILE is in use.

qcengine/programs/gcp.py
loriab: i've adjusted the version slurping

qcengine/programs/gcp.py
hokru: the `-local` option in the classic version did nothing. Not sure what happend there. `-level file` is the same as specifying no `-level`, ie reading from `~/.gcppar.$HOSTNAME`

loriab:
Ignoring the test failures for now (waiting for a pydantic v0.8.1 release on c-f), I think this is ready for review, @hokru and @awvwgk.

awvwgk (commented):
What is the reason to put the new gCP driver in the `xtb` environment? The more interesting case IMO would be to have both gCP drivers in the same environment, if they can life peacefully together, or together with DFT harness which can actually make use of the gCP correction.

loriab:
> What is the reason to put the new gCP driver in the xtb environment? The more interesting case IMO would be to have both gCP drivers in the same environment, if they can life peacefully together, or together with DFT harness which can actually make use of the gCP correction.

Yes, a dispersion lane for testing would be nice. I've run gcp (intel, iomp5) and mctc-gcp (openblas, gomp) together locally, but psi4 and mctc-gcp can't install together at present. Glad to re-sort in future as variants multiply.

awvwgk (approved):
I see nothing holding this back.

hokru (approved):
looks good.

Lnaden (approved):
Approved by people familiar with code changed.

Issue #174, opened 2019-11-15
Implement DFT-D4 harness


awvwgk:
**Is your feature request related to a problem? Please describe.**
The DFT-D4 dispersion correction has been [recently published](https://doi.org/10.1063/1.5090222) and should be made available for commonly used quantum chemistry packages.

**Describe the solution you'd like**
An integration via the `dftd4` C-API or Python-API in QCEngine.

**Describe alternatives you've considered**
IO based integration would be possible but is not really desirable.

**Additional context**
- the reference implementation of `dftd4` is available here: https://github.com/dftd4/dftd4
- a C++ ported version of `dftd4` is available (used for ORCA 4.2.0)
- I am also volunteering to implement the harness in QCEngine

dgasmith:
This is great, we would be very happy to get this integration in. Let us know if we can do anything to help.

PR #291, opened 2021-02-26
Implement support for DFT-D4

awvwgk:
## Description

- implement harness for `dftd4` Python API (closes https://github.com/MolSSI/QCEngine/issues/174)

See https://github.com/dftd4/dftd4/pull/70 for upstream implementation

## Changelog description

Implement support for DFT-D4

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/dftd4.py
+++ b/qcengine/programs/dftd4.py
@@ -1,0 +1,73 @@
+"""
+Harness for the DFT-D4 dispersion correction.
+This implementation interfaces with the dftd4 Python-API, which provides
+native support for QCSchema.
+
+Therefore, this harness only has to provide a thin wrapper to integrate dftd4.
+"""
+
+from typing import Dict
+
+from qcelemental.models import AtomicInput, AtomicResult
+from qcelemental.util import safe_version, which_import
+
+from ..config import TaskConfig
+from .model import ProgramHarness
+
+
+class DFTD4Harness(ProgramHarness):
+    """Calculation harness for the DFT-D4 dispersion correction."""
+
+    _defaults = {
+        "name": "dftd4",
+        "scratch": False,
+        "thread_safe": True,
+        "thread_parallel": False,
+        "node_parallel": False,
+        "managed_memory": False,
+    }
+    version_cache: Dict[str, str] = {}
+
+    class Config(ProgramHarness.Config):
+        pass
+
+    @staticmethod
+    def found(raise_error: bool = False) -> bool:
+        """Check for the availability of the Python API of dftd4"""
+
+        return which_import(
+            "dftd4",
+            return_bool=True,
+            raise_error=raise_error,
+            raise_msg="Please install via a dftd4 version with enabled Python API",
+        )
+
+    def get_version(self) -> str:
+        """Return the currently used version of dftd4"""
+        self.found(raise_error=True)
+
+        which_prog = which_import("dftd4")
+        if which_prog not in self.version_cache:
+            import dftd4
+
+            self.version_cache[which_prog] = safe_version(dftd4.__version__)
+
+        return self.version_cache[which_prog]
+
+    def compute(self, input_data: AtomicInput, config: TaskConfig) -> AtomicResult:
+        """
+        Actual interface to the dftd4 package. The compute function is just a thin
+        wrapper around the native QCSchema interface of the dftd4 Python-API.
+        """
+
+        self.found(raise_error=True)
+
+        import dftd4
+        from dftd4.qcschema import run_qcschema
+
+        # Run the Harness
+        output = run_qcschema(input_data)
+
+        # Make sure all keys from the initial input spec are sent along
+        output.extras.update(input_data.extras)
+        return output
--- a/qcengine/programs/base.py
+++ b/qcengine/programs/base.py
@@ -8,6 +8,7 @@ from ..exceptions import InputError, ResourceError
 from .adcc import AdccHarness
 from .cfour import CFOURHarness
 from .dftd3 import DFTD3Harness
+from .dftd4 import DFTD4Harness
 from .gamess import GAMESSHarness
 from .gcp import GCPHarness
 from .molpro import MolproHarness
@@ -126,5 +127,6 @@ register_program(OpenMMHarness())
 
 # Analytical Corrections
 register_program(DFTD3Harness())
+register_program(DFTD4Harness())
 register_program(GCPHarness())
 register_program(MP2DHarness())
--- a/qcengine/testing.py
+++ b/qcengine/testing.py
@@ -144,6 +144,7 @@ _programs = {
     "cfour": which("xcfour", return_bool=True),
     "dftd3": which("dftd3", return_bool=True),
     "dftd3_321": is_program_new_enough("dftd3", "3.2.1"),
+    "dftd4": which_import("dftd4", return_bool=True),
     "qcore": is_program_new_enough("qcore", "0.8.9"),
     "gamess": which("rungms", return_bool=True),
     "gcp": which("gcp", return_bool=True),


loriab:
> Note:
The Python API must be explicitly enabled when building dftd4 or can exist separately from a dftd4 installation.
The harness can therefore fail to use dftd4 even if a dftd4 installation is present, in this case it could have a fallback for a FIFO calculator (xyz + command line input, JSON dump output), but this is work for a different PR.

Can `def found` look for both `which_import(dftd4)` and then check separately for the API component or the external API package before declaring the program found? similar to https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/nwchem/runner.py#L55

loriab:
don't worry about nwchem::test_error_correction -- it's ~20% flaky

awvwgk:
> Can `def found` look for both `which_import(dftd4)` and then check separately for the API component or the external API package before declaring the program found?

This should be the way to go, for now I want to just flesh out the Python API of `dftd4`. I had a brief look into the `dftd3` FIFO harness which is incredible more complex than using a Python API, therefore this is nothing I want to tackle now.

I would be more interesting if the proposed interface is optimal for your purpose. I'm using the keywords to specify the damping parameters or the method name that should be looked up in the internal database. For `dftd3` is handled a bit differently, what is there preference here?

loriab:
> I would be more interesting if the proposed interface is optimal for your purpose. I'm using the keywords to specify the damping parameters or the method name that should be looked up in the internal database. For dftd3 is handled a bit differently, what is there preference here?

The Python API alone will be great, thank you. No need for file + cmdline route -- that was just the only way circa 2012.

If it's all the same to you, would you adapt the fields toward https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/empirical_dispersion.py#L193-L209 ? Main changes are: 
*  `d4` goes into `keywords.level_hint` (https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/empirical_dispersion.py#L58)
* any explicit parameters go into `keywords.params_tweaks` as a dictionary https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/empirical_dispersion.py#L62-L65
* `model.method` be, for example, `pbe-d4` if unperturbed or empty string if parameters at all perturbed https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/empirical_dispersion.py#L68-L71

d3/d4 don't fit too well into the AtomicInput layout, unfortunately, but the above would be helpful in integrating into psi if it's not too much trouble.

awvwgk:
Sounds good to me, DFT-D4 will provide this interface in the Python API to run QCSchema input. The only particularity, I don't allow to overwrite individual damping parameters, so either a method name or all parameters are provided.

loriab:
> The only particularity, I don't allow to overwrite individual damping parameters, so either a method name or all parameters are required.

That's no problem for psi, as once user input passes through the resolver, a full set of parameters is always provided.

Thanks for the changes.

awvwgk:
DFT-D4 3.1.0 is rolled out to conda-forge, just another hour until it propagates through the CDN and we can use it here to setup the testing environment.

awvwgk:
What is the current status of this patch? It has been stale for a while now, is there anything I can do to help move this forward?

loriab:
> What is the current status of this patch? It has been stale for a while now, is there anything I can do to help move this forward?

I think it's great and good to merge. I'm sorry for the delay; I was hoping to get https://github.com/awvwgk/QCEngine/pull/63 merged into this beforehand, but then I got a cycle behind you on dftd4 revisions and didn't revisit. Shall we get this merged sooner, and then I'll propose updates once I've rechecked the abovementioned harness PR and https://github.com/psi4/psi4/pull/2142 ?

awvwgk:
@loriab, sorry, I didn't see the PR against my fork, I usually don't watch my own forks because I have bots syncing them with upstream, so 99% of the activity there is usually noise (or myself).

I had a quick look over your changes and added a few adjustments to the parameter reading.

awvwgk:
There are a few tests from test_dftd3 now failing, but I guess those are the added tests for dftd4. I think we should come up with a clearer naming here, otherwise this is just confusing.

Any idea why those fail? Did I break something upstream already?

awvwgk:
I might have to look into the damping parameter generation again, we might be propagating a bad API decision from the C or Python layer up into qcng here (see https://github.com/dftd4/dftd4/issues/98). It might be worth to make both method and param_tweaks input an error at the Python level in dftd4, catch in the qcschema runner in dftd4 and send it back as ComputeError to qcng at some point until dftd4 is actually able to handle this case (load method first, overwrite parameters one by one).

qcengine/programs/empirical_dispersion_resources.py
```diff
@@ -795,9 +795,85 @@
...
         },
     },
+    "d4bj": {
+        "formal": "D4(BJ)",
+        "alias": ["d4"],
+        "description": "    Grimme's -D4 (BJ-damping) Dispersion Correction with ATM",
+        "citation": "    Caldeweyher, E.; Ehlert, S.; Hansen, A.; Neugebauer, H.; Spicher, S.; Bannwarth, C.; Grimmme, S., J. Chem. Phys. 150, 154122 (2019)\n",
+        "bibtex": "Caldeweyher:2019:154122",
+        "doi": "10.1063/1.5090222150",
+        "default": collections.OrderedDict([("s8", 1.0), ("a1", 1.0), ("a2", 1.0), ("s9", 1.0)]),
```
loriab, line 805:
```suggestion
        "default": collections.OrderedDict([("s8", 1.0), ("a1", 1.0), ("a2", 1.0), ("s9", 1.0)]),
        "default": collections.OrderedDict(
            [("s8", 1.0), ("a1", 1.0), ("a2", 1.0), ("s9", 1.0), ("s6", 1.0), ("alp", 16.0)]
        ),
```

qcengine/programs/empirical_dispersion_resources.py
```diff
@@ -795,9 +795,85 @@
...
+    for method in _parameters:
+        try:
+            _entry = _parameters[method]["d4"]
+            params = get_params(_entry, _base, _defaults)
+
+            citation = params.pop("doi", None)
+            definitions[method] = dict(
+                params=params,
+                citation=citation,
+            )
```
loriab, line 867:
```suggestion
            )
            definitions[method]["params"].pop("damping")
            definitions[method]["params"].pop("mbd")
```

The non-bj damping schemes for dftd4 would get a separate block in the big `dashcoeff` anyway. I'm not sure how mbd maps, but I think it'd be a separate block, too. So dropping these items for now if that's ok with you.

loriab (commented):
Thanks for merging and all the patching. I've made a couple suggestions, but with those, the psi4 PR seems to be working, too. Will rebase psi4 and do another check tomorrow.

qcengine/programs/empirical_dispersion_resources.py
loriab: Why not also pop damping in the changes you made?

awvwgk: Deleting `damping` from `params` lead to test failures for me, seemed like it is needed somewhere in the dispersion resources, therefore I kept it.

loriab:
Was renaming the reference values like https://github.com/MolSSI/QCEngine/pull/291/commits/89f467670c5e39acd561de1f90299d430f949c82#diff-c24ed775befa47b0b9de0da35810c7a47d5a6c5aa7c35b3e892a8b12abb09de6L78 solely to get the tests to work? Or do you very much not want the current definition of D4 to appear under the D4(BJ) label? I ask because I think that renaming of the refs is hiding another problem. Right now without `damping` popped, everything is getting caught in https://github.com/MolSSI/QCEngine/pull/291/files#diff-db09fb462ae1f6675bd709a50e8180a58d94038aedc9331b2f71d47e8b0c9d93R90 . Generally, I try to make the `extras["qcvars"]` as specific as possible, so no matter if model.method is `b3lyp-d4` or `b3lyp-d4(bj)`, the qcvar is the latter.

I can fix all this up. But I'd like to (1) not have `damping` in the params list https://github.com/MolSSI/QCEngine/pull/291/files#diff-d3e3a63b874003f6347c470b383ecfc60206b7c05b53f03a03f9e0ca04f081b3R806 since it's already specified at L798 and (2) have the qcvar label be `D4(BJ)` for reliability rather than whatever's in `model.method`. Is either change unsavory to you? Alternately on (2), we can have both `D4(BJ)` and `D4` stored. (But only for D4=D4(BJ) among aliases, as I don't want to declare D=D2 https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/empirical_dispersion_resources.py#L19) :-)

qcengine/programs/empirical_dispersion_resources.py
loriab: Sorry, your reply here didn't pop up on my browser tab until just now, hence the mention again in the consolidation comment just now https://github.com/MolSSI/QCEngine/pull/291#issuecomment-834622132.

awvwgk:
Naming the dashlevel D4(BJ) is not completely accurate actually, we defined D4 as alias for D4(BJ, EEQ)-ATM in the original DFT-D4 publication, which was one of four potential methods we tested in the course of that work (D4(BJ, GFN2-xTB)-ATM, D4(BJ, GFN2-xTB)-MBD and D4(BJ, EEQ)-MBD were the other three). Therefore, I'm not completely happy with D4(BJ) because it misses the actual important information on the dispersion model.

loriab:
> Naming the dashlevel D4(BJ) is not completely accurate actually, we defined D4 as alias for D4(BJ, EEQ)-ATM in the original DFT-D4 publication, which was one of four potential methods we tested in the course of that work (D4(BJ, GFN2-xTB)-ATM, D4(BJ, GFN2-xTB)-MBD and D4(BJ, EEQ)-MBD were the other three). Therefore, I'm not completely happy with D4(BJ) because it misses the actual important information on the dispersion model.

Makes sense. Not recording the 3-body aspect was worrying me a bit, too. See what you think of https://github.com/awvwgk/QCEngine/pull/65, please. It gives me a clean test suite for qcng and for psi4 https://github.com/psi4/psi4/pull/2142.

Also feel free to rebase to incorporate #297 test suite fixes.

loriab:
Thanks for the merge. I've push a commit directly to try to fix the d4 CI.

qcengine/programs/empirical_dispersion_resources.py
```diff
@@ -850,7 +930,7 @@ def from_arrays(name_hint=None, level_hint=None, param_tweaks=None, dashcoeff_su
     dict
         Metadata defining dispersion calculation.
 
-        dashlevel : {'d1', 'd2', 'd3zero', 'd3bj', 'd3mzero', 'd3mbj', 'chg', 'das2009', 'das2010', 'nl'}
+        dashlevel : {'d1', 'd2', 'd3zero', 'd3bj', 'd3mzero', 'd3mbj', 'chg', 'das2009', 'das2010', 'nl', "d4bj"}
```
loriab, line 933:
```suggestion
        dashlevel : {'d1', 'd2', 'd3zero', 'd3bj', 'd3mzero', 'd3mbj', 'chg', 'das2009', 'das2010', 'nl', "d4bjeeqatm"}
```

qcengine/programs/dftd4.py
```diff
@@ -0,0 +1,140 @@
...
+            gradient = output.return_result
+            calcinfo["CURRENT GRADIENT"] = gradient
+            calcinfo["DISPERSION CORRECTION GRADIENT"] = gradient
+            if qcvkey:
+                calcinfo[f"{qcvkey} DISPERSION CORRECTION GRADIENT"] = gradient
+
+        if output.keywords.get("pair_resolved", False):
+            pw2 = output.extras["dftd4"]["additive pairwise energy"]
+            pw3 = output.extras["dftd4"]["non-additive pairwise energy"]
+            assert abs(pw2.sum() + pw3.sum() - energy) < 1.0e-8, f"{pw2.sum()} + {pw3.sum()} != {ene}"
```
loriab, line 132:
```suggestion
            assert abs(pw2.sum() + pw3.sum() - energy) < 1.0e-8, f"{pw2.sum()} + {pw3.sum()} != {energy}"
```

loriab (approved):
lgtm! Thanks for all the extra features and your patience with changes for psi4.

Ready to merge in your view?

loriab:
FYI, upon `import psi4`, psi4 processes the QCEngine empirical_dispersion_parameters file and thus now the parameters file from `dftd4-python` if present. That can crash if [this line](https://github.com/psi4/psi4/pull/2176/files) allowing DOIs isn't present, as it is now in psi4 master. I mention this in case someone has an older psi4 in the presence of dftd4-python and raises the issue with you instead of me, the guilty party.

awvwgk:
Can we make the setup of the parameters for DFT-D4 lazy? Like just populate the fields if somebody actually asks for them when calling DFTD4 or constructing damping parameters? This would greatly reduce the probability that somebody accidentally runs into this issue.

PR #294, opened 2021-03-17
Be able to work without using NWChem's line search

WardLT:
## Description
Fixes a bug where we retrieved only the initial geometry if you turn off the line-search option in NWChem's driver.

## Changelog description
Bug fix: Was retrieving only the first step in a geometry relaxation

## Status

- [x] Code base linted
- [x] Ready to go


loriab (approved):
lgtm! does this address the intermittent failure in `test_error_correction`, too?

WardLT:
I hope so! I was unable to make it happen after a few runs of the test

PR #296, opened 2021-04-12
Prevent util.execute from reading output files

Andrew-AbiMansour:
## Description
Adds  an extra keyword (`outfiles_load`) to [util.execute](https://github.com/MolSSI/QCEngine/blob/master/qcengine/util.py#L375)  that instructs [disk_files](https://github.com/MolSSI/QCEngine/blob/master/qcengine/util.py#L559) how to handle output files. If `outfiles_load=True` (default), the output files are read and their contents returned in `outfiles`. When `outfiles_load=False`, only the file paths are returned instead. This is useful and can be necessary when output files are large so keeping track of the output files while retaining the scratch dir becomes a viable alternative:

```python
inp = {
    "command": ...,
    "outfiles": [huge_data_file, ...],
    "outfiles_load": False, # do not load outfiles in memory
    "scratch_messy": True, # do not delete scratch dir
}

_, proc = execute(**inp)

proc["outfiles"][huge_data_file] -> Path
```

The default behavior is preserved (`outfiles_load=True` in `util.execute`), so by default `outfiles` stores the file contents i.e.

```python
inp = {
    "command": ...,
    "outfiles": [small_data_file, ...],
}

_, proc = execute(**inp)

proc["outfiles"][small_data_file] -> Union[str, bytes]
```

## Changelog description
util.execute supports  tracking output files without loading them in memory

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/util.py
+++ b/qcengine/util.py
@@ -566,7 +566,7 @@ def disk_files(
     cwd: Optional[str] = None,
     as_binary: Optional[List[str]] = None,
     outfiles_load: Optional[bool] = True,
-) -> Dict[str, Union[str, bytes]]:
+) -> Dict[str, Union[str, bytes, Path]]:
     """Write and collect files.
 
     Parameters
@@ -586,7 +586,7 @@ def disk_files(
         outfiles stores the path(s) instead.
     Yields
     ------
-    Dict[str] = str
+    Dict[str, Union[str, bytes, Path]]
         outfiles with RHS filled in.
 
     """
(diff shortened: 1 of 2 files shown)


loriab:
What about making it like `binary` where only the named files are not read into memory?

Perhaps add a note that `scratch_messy=True` required to avoid surprise.

andrew-abimansour:
I like your suggestion. I'll refine this and improve the docstring.

qcengine/util.py
```diff
@@ -377,6 +377,7 @@ def execute(
     infiles: Optional[Dict[str, str]] = None,
     outfiles: Optional[List[str]] = None,
     *,
+    outfiles_track: Optional[List[str]] = [],
```
loriab, line 380: Probably better to have `None` here, like `as_binary`. https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

andrew-abimansour: Good point but shouldn't matter here since `outfiles_track` should never be modified, but if you're concerned someone might I could change its type to an immutable object e.g.  
```python
outfiles_track: Optional[Tuple[str]] = (),
```
If you prefer a list for consistency that is by default `None`, I could add a single line to `disk_files`:
```python
outfiles_track = outfiles_track or []
```
that would make the current code work with no additional changes.

andrew-abimansour: Good point but shouldn't matter here since `outfiles_track` should never be modified, but if you're concerned someone might I could change its type to an immutable object like a `Tuple` OR if you prefer a list for consistency that is by default `None`, I could add a single line to `disk_files`:

```python
outfiles_track = outfiles_track or []
```
that would make the current code work with no additional changes.

PR #297, opened 2021-05-04
probe mdi

loriab:
## Status

- [ ] Code base linted
- [ ] Ready to go


loriab:
This'll fix CI. The psi.yaml change is needed b/c pymdi had a breaking change at 1.2, and psi4 1.3 still has the old interface. The opt-disp.yaml is my fault b/c modern nightly psi conda pkg is behind the mdi fix added to the code.

@awvwgk
@taylor-a-barnes

loriab:
This wasn't psi4's fault, rather the MDI interface needed updating for v1.2 here. @taylor-a-barnes, ideally, there'd be a version constraint now on mdi in testing.py, but I don't see where the mdi python module is storing its version.

qcengine/mdi_server.py
```diff
@@ -83,7 +83,7 @@ def __init__(
         mpi_world = None
         if use_mpi4py:
             mpi_world = MPI.COMM_WORLD
```
taylor-a-barnes, line 85: These lines (83-85) are no longer necessary, and can be removed to resolve an lgtm alert.

taylor-a-barnes (approved):
Thanks; other than the small edit associated with the lgtm alert, this is ready to go.

taylor-a-barnes:
@loriab MDI version information is stored in `MDI_MAJOR_VERSION` / `MDI_MINOR_VERSION` / `MDI_PATCH_VERSION`.  Not very pythonic, but consistent with MDI's MPI-like interface

PR #301, opened 2021-05-19
Properly handle compute errors raised in berny optimization procedure

### LINKED ISSUE #300: [BUG] berny optimization procedure fails to capture compute errors correctly
**Describe the bug**
The current `BernyProcedure.compute()` method does not correctly handle failed gradient calculations correctly. The method assumes that all values returned from `qcengine.compute()` are `AtomicResult` objects when in reality they may also be `FailedOperation` objects. This may occur if unsupported keywords are passed the compute engine specified for the operation.

The bug is caused by this line of code:
```python
"final_molecule": trajectory[-1]["molecule"],
```

The assumption is that the trajectory array will contain only `AtomicResult` objects and that the final one will contain a "molecule" key. In reality, if an error occurred with the gradient calculation the final object in the `trajectory` array is a `FailedOperation` object, so this line of code raises a `KeyError` and the returned `FailedOperation` object is indicative of only the `KeyError` raised in the procedure instead of the underlying error computing the gradient. This makes the failures hard to debug (or impossible if you can't read the source code).

**To Reproduce**
```python
from qcelemental.models import Molecule, OptimizationInput
from qcelemental.models.procedures import QCInputSpecification
import qcengine

water = Molecule.from_data("pubchem:water")

input_spec = QCInputSpecification(
    driver="gradient",
    model={"method": "b3lyp", "basis": "6-31g"},
    # Keywords specific to the compute engine that will execute gradient calculation
    # I.e., bad keywords for psi4; THIS IS WHAT CAUSES THE ERROR!
    keywords={'fake': 'bad'},
)

opt_input = OptimizationInput(
    initial_molecule=water,
    input_specification=input_spec,
    protocols={"trajectory": "all"},
    keywords={"program": "psi4", "maxsteps": 3},
)

qcengine.compute_procedure(opt_input, "berny")
```

The exception raised:
```
FailedOperation(error=ComputeError(error_type='unknown_error', error_message='QCEngine Execution Error:\nTraceback (most recent call last):\n  File "/opt/conda/lib/python3.7/site-packages/qcengine/util.py", line 114, in compute_wrapper\n    yield metadata\n  File "/opt/conda/lib/python3.7/site-packages/qcengine/compute.py", line 147, in compute_procedure\n    output_data = executor.compute(input_data, config)\n  File "/opt/conda/lib/python3.7/site-packages/qcengine/procedures/berny.py", line 78, in compute\n    "final_molecule": trajectory[-1]["molecule"],\nKeyError: \'molecule\'\n'))
```

**Expected behavior**
A `FailedOperation` object should be returned specifying that the keywords for psi4 are incorrect. The procedure code should not incorrectly look for the `molecule` property on the final object in the `trajectory` array; rather, the code should correctly handle the case of a failed `qcengine.compute` call.

coltonbh:
## Description
Closes #300 .

## Changelog description
Handle gradient computation errors correctly inside of `BernyProcedure.compute()` method

## Note on decisions
- Updated the `.compute()` call signature to reflect the possibility of returning a `FailedOperation`. While the `OptimizationResult` object may have `success=False`, it still requires `trajectory` and `energies` attributes, both of which cannot be `None`, and if a gradient computation fails on its first iteration it is not possible to create any `AtomicResult` or `energies` values; thus it is not possible to return an `OptimizationResult` object, even with `success=False` set on the object.

## Open Questions
- Is it necessary to declare types in the function signature using the "" employed previously in this method? Seems more canonical to declare return types without "" values unless they are the class being defined in a @classmethod. I have followed the convention already in the code but would consider updating it to remove the "" if allowed. 

## Status

- [ x] Code base linted
- [ x] Ready to go

### DIFF
--- a/qcengine/tests/test_procedures.py
+++ b/qcengine/tests/test_procedures.py
@@ -3,7 +3,7 @@ Tests the DQM compute dispatch module
 """
 
 import pytest
-from qcelemental.models import OptimizationInput
+from qcelemental.models import OptimizationInput, FailedOperation
 
 import qcengine as qcng
 from qcengine.testing import failure_engine, using
@@ -92,6 +92,23 @@ def test_berny_stdout(input_data):
     ret = qcng.compute_procedure(input_data, "berny", raise_error=True)
     assert ret.success is True
     assert "All criteria matched" in ret.stdout
+
+
+@using("psi4")
+@using("berny")
+def test_berny_failed_gradient_computation(input_data):
+
+    input_data["initial_molecule"] = qcng.get_molecule("water")
+    input_data["input_specification"]["model"] = {"method": "HF", "basis": "sto-3g"}
+    input_data["input_specification"]["keywords"] = {"badpsi4key", "badpsi4value"}
+    input_data["keywords"]["program"] = "psi4"
+
+    input_data = OptimizationInput(**input_data)
+
+    ret = qcng.compute_procedure(input_data, "berny", raise_error=False)
+    assert isinstance(ret, FailedOperation)
+    assert ret.success is False
+    assert ret.error.error_type == qcng.exceptions.InputError.error_type
 
 
 @using("geometric")
--- a/qcengine/tests/test_procedures.py
+++ b/qcengine/tests/test_procedures.py
@@ -100,7 +100,7 @@ def test_berny_failed_gradient_computation(input_data):
 
     input_data["initial_molecule"] = qcng.get_molecule("water")
     input_data["input_specification"]["model"] = {"method": "HF", "basis": "sto-3g"}
-    input_data["input_specification"]["keywords"] = {"badpsi4key", "badpsi4value"}
+    input_data["input_specification"]["keywords"] = {"badpsi4key": "badpsi4value"}
     input_data["keywords"]["program"] = "psi4"
 
     input_data = OptimizationInput(**input_data)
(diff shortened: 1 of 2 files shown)


coltonbh:
Another one for ya @loriab :)

coltonbh:
@loriab wanted to check in on this now that I'm back to developing a bit. Old PR--possible to merge in? This fixes basic error handling that was incorrect inside the `berny` procedure.

Thanks!

qcengine/tests/test_procedures.py
```diff
@@ -94,6 +94,23 @@ def test_berny_stdout(input_data):
...
+def test_berny_failed_gradient_computation(input_data):
+
+    input_data["initial_molecule"] = qcng.get_molecule("water")
+    input_data["input_specification"]["model"] = {"method": "HF", "basis": "sto-3g"}
+    input_data["input_specification"]["keywords"] = {"badpsi4key": "badpsi4value"}
+    input_data["keywords"]["program"] = "psi4"
+
+    input_data = OptimizationInput(**input_data)
+
+    ret = qcng.compute_procedure(input_data, "berny", raise_error=False)
```
loriab, line 108:
```suggestion
    input_model = OptimizationInput(**input_data)

    ret = qcng.compute_procedure(input_model, "berny", raise_error=False)
```

"model" for pydantic obj and "data" for dict is my convention. but my aim is merely a dummy commit to trigger CI.

loriab (approved):
I've hit this too. Thanks for the fix!

PR #307, opened 2021-07-07
Error correction for NWChem computations that fail to converge

WardLT:
## Description
Adds a simple error corrector for NWChem: Increase the number of iterations if it fails to converge

## Changelog description
Automatically increase the number of iterations when NWChem fails to converge

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_nwchem.py
+++ b/qcengine/programs/tests/test_nwchem.py
@@ -282,7 +282,7 @@ def test_autoz_error():
 
 
 @using("nwchem")
-def test_error_correction():
+def test_autoz_error_correction():
     """See if error correction for autoz works"""
 
     # Large molecule that leads to an AutoZ error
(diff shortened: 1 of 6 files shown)


qcengine/programs/tests/test_nwchem.py
```diff
@@ -301,3 +301,37 @@ def test_error_correction():
...
+@using("nwchem")
+def test_conv_threshold(h20v2, method, keyword, init_iters, use_tce):
+    result = qcng.compute(
+        {
+            "molecule": h20v2,
+            "model": {"method": method, "basis": "sto-3g"},
+            "driver": "energy",
+            "keywords": {
+                keyword: init_iters,
+                "qc_module": use_tce,
```
loriab, line 327: I've been using qc_module = tce/cc/mp2grad/directmp2 rather than a bool here. Can that work with your plans?

loriab (commented):
thanks, looks good!

qcengine/programs/tests/test_nwchem.py
WardLT: It would, and I think that gives us better flexibility and clarity that "qc_module" == True means that we're using TCE

loriab: sounds good. I'll merge as-is unless you've got bool->str edits underway for this PR.

loriab (approved):

qcengine/programs/tests/test_nwchem.py
WardLT: I'd merge as-is. Do we have an issue open for the bool->str?

loriab: no open issue. in #304 I've got the harvester collecting the module and storing it in `provenance.module`. that's a catch-all PR that'll be refactored and split before consideration.

PR #308, opened 2021-07-27
slightly loosen xtb testing

loriab:
## Description
will this be ok, @awvwgk? it fixes CI.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_xtb.py
+++ b/qcengine/programs/tests/test_xtb.py
@@ -97,7 +97,7 @@ def test_xtb_task_gfn1xtb_m02():
 @using("xtb")
 def test_xtb_task_gfn1xtb_m03():
 
-    thr = 1.0e-8
+    thr = 1.0e-7
 
     return_result = np.array(
         [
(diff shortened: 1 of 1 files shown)


loriab:
will this be ok, @awvwgk? it fixes CI. (sorry if double-ping -- I don't think GH pings on edits.)

awvwgk:
I think the current threshold is relative rather than absolute (https://docs.pytest.org/en/6.2.x/reference.html#pytest.approx). I switched to `pytest.approx(..., abs=thr)` for testing in my projects since the relative comparison can fail for very small (numerical zero) values otherwise.

awvwgk:
> sorry if double-ping -- I don't think GH pings on edits.

No problem. Looks like GH pings correctly for mentions in edits.

loriab:
> I think the current threshold is relative rather than absolute (https://docs.pytest.org/en/6.2.x/reference.html#pytest.approx). I switched to pytest.approx(..., abs=thr) for testing in my projects since the relative comparison can fail for very small (numerical zero) values otherwise.

Sounds good -- it was checking some elements to 1e-11 before, so this is a sensible relaxation.

awvwgk (approved):

bennybp (approved):

PR #309, opened 2021-08-06
print location in qcengine info

loriab:
## Changelog description
- [x] print program location in qcengine info
- [x] return version for geometric, openmm, rdkit
- [x] fix a few little typos and bug
- [x] check CI after psi4 release

## Status

- [x] Code base linted
- [x] Ready to go


qcengine/programs/gamess/runner.py
```diff
@@ -2,8 +2,10 @@
 
 import copy
 import pprint
+import sys
+import traceback
 from decimal import Decimal
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
```
dotsdl, line 8: These additions don't appear to be used in this module. Is there a reason for adding them?

qcengine/programs/nwchem/runner.py
```diff
@@ -5,6 +5,8 @@
 import logging
 import pprint
 import re
+import sys
+import traceback
```
dotsdl, line 9: These don't appear to be used in this module. Why are they added?

qcengine/programs/cfour/runner.py
```diff
@@ -2,6 +2,8 @@
 
 import copy
 import pprint
+import sys
+import traceback
```
dotsdl, line 6: Not used in this module. Is there a reason for adding them?

dotsdl (approved):
Great work @loriab! Just a few notes to address, but otherwise happy to merge!

qcengine/programs/gamess/runner.py
loriab: This is the first bite-sized PR from a big block of changes. These imports are used later on, so it seemed easier to avoid missing imports to add them now. I don't mind running an analysis at the end of the PR series to prune away unused imports.

devtools/conda-envs/psi.yaml
```diff
@@ -14,8 +14,8 @@ dependencies:
   - pyyaml
   - py-cpuinfo
   - psutil
-  - qcelemental=0.13.0  # test minimum stated version.
-  - pydantic=1.2.0  # test minimun stated version. actually min is 1.0 but no py38 builds
+  - qcelemental=0.21.0  # test minimum constrained version. min stated version is 0.19.
```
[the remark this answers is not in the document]
loriab: "min stated" is what qcng setup.py says. If you were building up your environment from source, it'd work for qcng. "min constrained" is an actual conda env that will install. In this case, I think psi is demanding qcel>=0.21, so this line will keep testing the minimum qcel that can be solved.

PR #313, opened 2021-08-28
fix openmm version

loriab:
## Status

- [x] Code base linted
- [x] Ready to go


dotsdl (approved):
Thanks @loriab! This is due to latest OpenMM 7.6.0. The fix in this PR will work for now, but I've laid out a more general fix in #314 I can pursue later.

PR #315, opened 2021-09-03
return frame for cfour, gms, nwc. harvest and test grad and hess.

loriab:
## Description

- [x] cfour, gms, nwc now returns AtomicInput or program native orientation depending on fix_com & fix_orientation= T or F. Psi4 already did this. Previously these three always returned AtomicInput orientation. Note that when returning program native orientation, the molecule is overwritten, so AtomicResult is not a superset of AtomicInput. Add test that randomly shuffles molecule.
- [x] Gamess read dat file
- [x] harvest gradients and hessians from all four programs. add HF Hessian test
- [x] refactor standard suite test code to accommodate relaxed array tolerances

## Status

- [x] Code base linted
- [ ] Ready to go


qcengine/programs/tests/standard_suite_runner.py
```diff
@@ -91,31 +162,80 @@ def runner_asserter(inp, subject, method, basis, tnm):
         }
     )
 
+    local_options = {}
+    # local_options = {"nnodes": 1, "ncores": 1}  # debug
```
loriab, line 166:
```suggestion
    # local_options = {}  # debug with default (parallel)
    local_options = {"nnodes": 1, "ncores": 1}   # run serial by default
```

bennybp (approved):

PR #317, opened 2021-09-03
a few harness docs

loriab:
## Description
Add some new harness docs.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/docs/requirements.yml
+++ b/docs/requirements.yml
@@ -4,7 +4,7 @@ channels:
     - conda-forge
 dependencies:
     - python=3
-    - sphinx
+    - sphinx <4
     - sphinx_rtd_theme
     - sphinx-automodapi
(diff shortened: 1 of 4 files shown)


WardLT (approved):
Looks good to me. 

Do we have a preferred path for people to document the keywords/options for their harness? If so, let's add mention of that here. If not, shall I create an issue about that?

bennybp (approved):

loriab:
>>> Do we have a preferred path for people to document the keywords/options for their harness? If so, let's add mention of that here. If not, shall I create an issue about that?

Not at the moment since I usually consider that qcdb's domain where one has to register keywords and methods, whereas qcng could be template-like. But qcng harnesses should have the option. Yes, please do make issue.

PR #319, opened 2021-09-10
Fixed OpenMMHarness imports to support version 7.6+

### LINKED ISSUE #314: Update OpenMMHarness to support new import path for openmm >7.6.0
**Describe the bug**

[OpenMM 7.6.0 ](https://github.com/openmm/openmm/releases/tag/7.6.0) is now out, and this brings with it a change to the import paths of OpenMM components from:
*  `import simtk.openmm` to `import openmm`
* `import simtk.unit` to `import openmm.unit`

We should change the imports and version handling for the `OpenMMHarness` to try the new usage first, then fall back to the old usage if it fails in order for older OpenMM releases to still work under the harness.

dotsdl:
Fixes #314

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/openmm.py
+++ b/qcengine/programs/openmm.py
@@ -151,6 +151,7 @@ class OpenMMHarness(ProgramHarness):
         Generate an OpenMM System object from the input molecule method and basis.
         """
         from openmmforcefields.generators import SystemGenerator
+
         try:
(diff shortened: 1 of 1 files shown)


loriab (approved):
those are some import gymnastics!

PR #321, opened 2021-10-01
native_files and error_stamp

loriab:
## Changelog description

- [x] Implement `native_files` protocol for cfour, gamess, nwchem, psi4, dftd3, mp2d, gcp to return the input under `atres.native_files["input"]` and any other useful text files under `atres.native_files[<customary_name>]`. This is associated with MolSSI/QCElemental#275 and won't work here until that's in and qcel 0.24.0 is minted.
- [x] Add `error_stamp` fn that formats native input, stdout, stderr, and traceback as a nice chunk of info to raise during harness exe.
- [x] Add test case `test_canonical_fields.py` to check that `native_files` protocol implemented correctly.

## Status

- [x] Code base linted
- [x] Ready to review and comment upon, but it's not actually going to work until upstream changes
- [x] Ready to go


qcengine/programs/cfour/harvester.py
```diff
@@ -372,7 +372,7 @@ def harvest_outfile_pass(outtext):
         re.MULTILINE | re.DOTALL,
     )
     if mobj:
-        print("matched mp4tro")
+        logger.debug("matched mp4tro")
```
WardLT, line 375: Thanks for replacing all of these prints!

WardLT (approved):
Looks good to me! I like the improved focus on separating out native files and the error_stamp for standardizing exceptions.

PR #323, opened 2021-10-06
bugfix: 'N Atoms' is now set in qcvars for grad/Hessian calcs

eljost:
## Description
The update to 0.20.0 broke the Turbomole harness, as 'N ATOMS' was missing in qcvars, when derivatives of the energy were calculated. This line led to an exception:

https://github.com/MolSSI/QCElemental/blob/36c5e0d3de3655fe02c42c4c1fe52eb4ba933c54/qcelemental/models/results.py#L279

## Changelog description
`qcvars['N ATOMS']` is now set for gradient and/or Hessian runs, so the AtomicProperties can be now be properly created.

## Status
- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/turbomole/runner.py
+++ b/qcengine/programs/turbomole/runner.py
@@ -7,12 +7,13 @@ from decimal import Decimal
 from pathlib import Path
 from typing import Any, Dict, Optional, Tuple
 
-from qcelemental.models import AtomicResult, Provenance
+from qcelemental.models import AtomicResult, Provenance, BasisSet
 from qcelemental.util import safe_version, which
 
-from ...util import execute, temporary_directory
+from ...exceptions import InputError
 from ..model import ProgramHarness
 from ..qcvar_identities_resources import build_atomicproperties, build_out
+from ...util import execute, temporary_directory
 from .define import execute_define, prepare_stdin
 from .harvester import harvest
 from .methods import KEYWORDS, METHODS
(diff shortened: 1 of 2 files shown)


loriab (approved):
Sorry for breaking, and thanks very much for the fix. Shall I mint a v0.20.1 to help downstream?

Why does the `test_harness_canonical` need to be commented?

sheepforce:
> Shall I mint a v0.20.1 to help downstream?

I would very much appreciate a bug fix release; it is easier to package a small release than backporting the changes (https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/qcengine/default.nix)

eljost:
> Sorry for breaking, and thanks very much for the fix. Shall I mint a v0.20.1 to help downstream?
> 
> Why does the `test_harness_canonical` need to be commented?

Because, as far as I can tell, the `qcsk_bs` basis is not a simple, common basis set name as "6-31G", but a QCSchema basis set?! This is not handled by the TurbomoleHarness, so the test fails. That is why I disabled it for now.

loriab:
>> Why does the test_harness_canonical need to be commented?

> Because, as far as I can tell, the qcsk_bs basis is not a simple, common basis set name as 6-31G, but a QCSchema basis set?! This is not handled by the TurbomoleHarness, so the test fails. That is why I disabled it for now.

You're right about that being a qcschema basisset. No qcengine harness can handle such basis sets right now, even though the schema allow them, so that test just tests that a sensible error message is thrown, rather than harnesses reacting in a dozen different ways. Perhaps add something like https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/nwchem/runner.py#L178-L179 to the turbomole harness to allow it to pass.

eljost:
Ah, ok. I'll come up with something tomorrow.

Issue #326, opened 2021-11-02
Use GPU when invoking torchani


kexul:
Dear molssi developers: 
      I found my workstation is not using GPUs when executing the computing using torchani. Are there some environment variables I need to set? 

Here is the code I've used: 
```
import qcengine as qcng
import qcelemental as qcel

mol = qcel.models.Molecule.from_data("""
O  0.0  0.000  -0.129
H  0.0 -1.494  1.027
H  0.0  1.494  1.027
""")
ret = qcng.compute(inp, "torchani")
ret
```

Here is the output of ` ret.provenance`: 
```
Provenance(creator='torchani', version='unknown', routine='torchani.builtin.aev_computer', wall_time=0.008759737014770508, cpu='Intel(R) Xeon(R) Platinum 8255C CPU @ 2.50GHz', username='root', hostname='VM-121-111-centos', qcengine_version='v0.20.1')
```

kexul:
Context: 
I'm using this https://github.com/openforcefield/bespoke-fit package to refit some force filed parameters, when using torchani as the QC backend, it consumes a lot of memory and ran quite slow, so I'm finding ways to move the model to GPU to accelerate the computation.

PR #327, opened 2021-11-02
Allowing torchani to use GPU 

kexul:
## Description
Added config for torchani to use GPU, related to this issue https://github.com/MolSSI/QCEngine/issues/326. 
Now we can use `local_options={'device':'cuda'}` to assign the device: 
```
import qcengine as qcng
import qcelemental as qcel

mol = qcel.models.Molecule.from_data("""
O  0.0  0.000  -0.129
H  0.0 -1.494  1.027
H  0.0  1.494  1.027
""")

inp = qcel.models.AtomicInput(
    molecule=mol,
    driver="energy",
    model={"method": "ani2x"}
    )

ret = qcng.compute(inp, "torchani", local_options={'device':'cuda'})
print(ret.return_result)
```

### DIFF
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -8,7 +8,6 @@ import logging
 import os
 import socket
 from typing import Any, Dict, Optional, Union
-from typing_extensions import Literal
 
 import pydantic
 
@@ -159,8 +158,7 @@ class TaskConfig(pydantic.BaseModel):
     scratch_messy: bool = pydantic.Field(
         False, description="Leave scratch directory and contents on disk after completion."
     )
-    device: Literal['cpu', 'cuda'] = 'cpu' # which device to perform calculation
-    
+        
     class Config:
         extra = "forbid"
 
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -158,7 +158,7 @@ class TaskConfig(pydantic.BaseModel):
     scratch_messy: bool = pydantic.Field(
         False, description="Leave scratch directory and contents on disk after completion."
     )
-        
+
     class Config:
         extra = "forbid"
 
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -8,6 +8,7 @@ import logging
 import os
 import socket
 from typing import Any, Dict, Optional, Union
+from typing_extensions import Literal
 
 import pydantic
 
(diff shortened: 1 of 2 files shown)


loriab:
Thanks for the PR, @kexul . @dotsdl, how does this relate to how OpenMM or OpenFF in general switches cpu/gpu modes?

WardLT:
Instead of having the user manually turn on GPU usage, could we use torch
to detect the presence of a GPU and automatically use it?

kexul:
> Instead of having the user manually turn on GPU usage, could we use torch to detect the presence of a GPU and automatically use it?

That sounds reasonable, I've changed the code according to your suggestion.

qcengine/config.py
```diff
@@ -158,7 +158,7 @@ class TaskConfig(pydantic.BaseModel):
     scratch_messy: bool = pydantic.Field(
         False, description="Leave scratch directory and contents on disk after completion."
     )
-
+
```
WardLT, line 161: Introducing the whitespace here probably caused a problem with our linter.

Could you run `make format` and issue a commit with the changed files?

qcengine/programs/torchani.py
```diff
@@ -172,7 +173,7 @@ def compute(self, input_data: "AtomicInput", config: "TaskConfig") -> "AtomicRes
         ret_data["extras"] = input_data.extras.copy()
         ret_data["extras"].update(
             {
-                "ensemble_energies": energy_array.detach().numpy(),
+                "ensemble_energies": energy_array.cpu().detach().numpy(),
                 "ensemble_energy_avg": energy.item(),
```
WardLT, line 177: To make sure I understand correctly, calling `item` will move the "energy" tensor to the cpu? You don't need a `.cpu()` statement, right?

WardLT (approved):
The changes look great to me. Thanks for the contribution!

Once you fix the linting problems with `make format`, this should be ready to merge.

@loriab , do you want to review this as well? Or, can I click the button?

qcengine/programs/torchani.py
kexul: Yes, calling `item()` would move the tensor to CPU, you are absolutely right.

qcengine/config.py
kexul: Done by https://github.com/MolSSI/QCEngine/pull/327/commits/d7ced823512ce58133739327684cae380592309c  😉

loriab:
Thanks @WardLT , merge at will.

PR #330, opened 2021-11-15
Add in restart capability to NWChem

WardLT:
## Description
Adds the ability for NWChem computations to be restartable.

If NWChem terminates due to walltime it leaves the scratch directory. This commit gives each computation a repeatably-named output directory, so that the QCEngine will restart from that directory if it still exists.

## Changelog description
NWChem can restart if QCEngine is killed

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/nwchem/runner.py
+++ b/qcengine/programs/nwchem/runner.py
@@ -165,6 +165,7 @@ class NWChemHarness(ErrorCorrectionProgramHarness):
 
         # Shortcut: If the user allows restart and the restart directory is present
         allow_restart = opts.pop("allow_restarts", True)
+        scr_name = opts.pop("scratch_name", None)
 
         # Handle memory
         # * [GiB] --> [QW]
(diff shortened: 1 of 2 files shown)


loriab:
Any chance that restart requires a known, small set of files, binary or text, short of the full scratch directory? I fended off (from myself) a plan to record the scratch name in provenance in favor of pulling back useful native files into the AtomicResult. The circumstance you're planning for is qcng dies and there is no AtomicResult?

WardLT:
I think there is only one file - the NWChem DB - that must be preserved. I'll check on the file sizes to see if it is small enough that we should grab it anyway.

However, the target for here is when an AtomicResult is not returned. The specific use case for these changes is when my HPC allocation is terminated and I just want to start from where I left off. No need to shuttle files between systems. Just letting the `compute` function end up in the same directory as last time so NWChem can do its magic.

loriab (commented):
lgtm! What do you think of `extras["nwchem:restart"] = {"allow_restart": bool, "scr_name": str}` or something like that instead of mixing with `keywords`?

loriab (approved):

PR #332, opened 2021-11-20
nwchem restart opt-in

loriab:
## Description

Modification of #330 such that you have to deliberately enable NWChem restarts.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/nwchem/runner.py
+++ b/qcengine/programs/nwchem/runner.py
@@ -205,7 +205,7 @@ class NWChemHarness(ErrorCorrectionProgramHarness):
 
         # Now that we know the computation, check if we want a restart
         #  TODO (wardlt): Store the restart information in the local_options configuration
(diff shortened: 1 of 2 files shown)


WardLT:
What are your thoughts against keeping it automatic? I can think of more reasons to be happy that NWChem picks up from where it left off (the default behavior) than not.

loriab:
> What are your thoughts against keeping it automatic? I can think of more reasons to be happy that NWChem picks up from where it left off (the default behavior) than not.

When I started running pytest repeatedly locally, various nwchem tests started to fail. I suspect it's due to repeated pytest runs. That could be remedied by adding tmpdir, but that's a sudden change in behavior to need that.

WardLT:
I agree with the points you raise above. Let's merge this, and I'll continue working on the documentation for the NWChem harness so people know how to safely use restarts.

WardLT (approved):

PR #333, opened 2021-12-06
Allow NWChem to restart single-point computations

WardLT:
## Description
Allows restarts to work for single-point NWChem computations, adds tests for that functionality.

As written, the NWCHarness does not write the geometry to the input
file for restart calculations. This is great for Driver computations
because it prevents the computation from restarting from the initial
geometry. However, the geometry is required for single point calculations
(esp Hessian, Gradient) because we use the geometry to determine
how NWChem rotates the coordinate system so that we can adjust
any direction-dependant quantities back to the initial coordinate system.

This change is simple: I add a flag that is set by the "driver" computation
to tell the NWCHarness not to write the geometry for restarts.
For clarity, the NWCOptimizerHarness calls the NWChemHardness
to make input files

## Changelog description
Allow NWChem to restart single-point calculations

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_nwchem.py
+++ b/qcengine/programs/tests/test_nwchem.py
@@ -337,7 +337,7 @@ def test_conv_threshold(h20v2, method, keyword, init_iters, use_tce):
     assert result.extras["observed_errors"]["convergence_failed"]["keyword_updates"] == {keyword: init_iters * 4}
 
 
-@using('nwchem')
+@using("nwchem")
 def test_restart(nh2, tmpdir):
     resi = {
         "molecule": nh2,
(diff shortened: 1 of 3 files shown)


qcengine/programs/tests/test_nwchem.py
```diff
@@ -335,3 +335,35 @@ def test_conv_threshold(h20v2, method, keyword, init_iters, use_tce):
...
+    assert "computation failed to converge" in str(result.error)
+
+    # Run again: It should converge
+    result = qcng.compute(
+        resi,
+        "nwchem",
+        local_options=local_options,
+        raise_error=False,
+    )
+    assert result.success
```
loriab, line 369: Just so I'm clear -- the job is working at second submission because the DFT requires between 4 and 8 iterations to converge?

loriab (approved):
Makes sense, especially once "driver = nwc optimizer" recalled. Thanks very much for the fix.

Would you rebase, please, so the right CI checks run?

qcengine/programs/tests/test_nwchem.py
WardLT: I'll add that in as a line comment

loriab:
Yes, I was going to squash.

WardLT:
:+1: Then I'm going to skip fixing my git commits, if that's OK

PR #338, opened 2022-01-04
fix version in tarball

loriab:
## Description
 fixes #337 confirmed

PR #339, opened 2022-01-15
Make QCEngine tolerant of `cpuinfo` failure to populate `brand_raw`, `brand`

dotsdl:
## Description

Currently, we observe frequent failure of `cpuinfo` to give brand information when used within Docker containers on Pacific Research Platform. It is unclear why, but making QCEngine tolerant of this information being missing would help in production usage.

## Changelog description

QCEngine calls now tolerant of missing `brand_raw` or `brand` info from py-cpuinfo.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -53,12 +53,16 @@ def get_global(key: Optional[str] = None) -> Union[str, Dict[str, Any]]:
             _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand_raw"]
         except KeyError:
             # Remove this if py-cpuinfo is pinned to >=6.0.0
-            _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand"]
+            try:
+                _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand"]
+            except KeyError:
+                pass
+
 
     if key is None:
         return _global_values.copy()
     else:
-        return _global_values[key]
+        return _global_values.get(key)
 
 
 class NodeDescriptor(pydantic.BaseModel):
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -53,15 +53,12 @@ def get_global(key: Optional[str] = None) -> Union[str, Dict[str, Any]]:
             _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand_raw"]
         except KeyError:
             # Remove this if py-cpuinfo is pinned to >=6.0.0
-            try:
-                _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand"]
-            except KeyError:
-                pass
+            _global_values["cpu_brand"] = _global_values["cpuinfo"].get("brand", 'unknown')
 
     if key is None:
         return _global_values.copy()
     else:
-        return _global_values.get(key)
+        return _global_values[key]
 
 
 class NodeDescriptor(pydantic.BaseModel):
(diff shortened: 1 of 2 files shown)


qcengine/config.py
```diff
@@ -53,12 +53,15 @@ def get_global(key: Optional[str] = None) -> Union[str, Dict[str, Any]]:
             _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand_raw"]
         except KeyError:
             # Remove this if py-cpuinfo is pinned to >=6.0.0
-            _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand"]
+            try:
```
WardLT, line 56: The nested `try/except` blocks seem a bit unnecessary. Would you mind changing to a `.get('brand', 'unknown')` syntax here so we store `cpu_brand='unknown'` if the resolution fails twice?

qcengine/config.py
```diff
@@ -53,12 +53,15 @@ def get_global(key: Optional[str] = None) -> Union[str, Dict[str, Any]]:
...
+                _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand"]
+            except KeyError:
+                pass
 
     if key is None:
         return _global_values.copy()
     else:
-        return _global_values[key]
+        return _global_values.get(key)
```
WardLT, line 64: I'd keep this line as `[key]` so we get `KeyError`s if a key is unknown still. I'm not sure what returning `None` will do to other parts of QCEngine.

WardLT (changes requested):

dotsdl:
Thanks @WardLT for the review! I've made the changes you requested. Let me know if there's anything else I can do to move this forward!

dotsdl:
Looks like the remaining CI errors may be unrelated to the changes in this PR. @loriab is this consistent with your view?

loriab:
Agreed, if you make this change https://github.com/MolSSI/QCEngine/pull/338/files#diff-d66842b418a1c03f41438563803c987c9defd1811ea0af131e285ba540c8c0d6R208, it'll fix it. May as well get that PR approved and merged, actually. With that change and this, can do a new minor release instead of just the patch. It'll help qcng versions catch up to qcel, anyways.

dotsdl:
Thanks @loriab! I'll review and merge #338, then pull that in to this PR today. Definitely appreciate getting a minor release out; will be a huge help for us in prod use!

qcengine/config.py
```diff
@@ -53,7 +53,7 @@ def get_global(key: Optional[str] = None) -> Union[str, Dict[str, Any]]:
             _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand_raw"]
         except KeyError:
             # Remove this if py-cpuinfo is pinned to >=6.0.0
-            _global_values["cpu_brand"] = _global_values["cpuinfo"]["brand"]
+            _global_values["cpu_brand"] = _global_values["cpuinfo"].get("brand", "unknown")
```
loriab, line 56: not a strong suggestion -- just wondering if `(none)` or `(unknown)` would be more consistent with other "missing" schema values (like `model.basis = "(none)"`)

WardLT: I'm all for simplifying the vocabulary. `(unknown)` would be my vote (the CPU has a brand, we just don't know it).

WardLT (approved):
Last change would be a minor, optional revision.

Thanks!

WardLT:
I just realized that I had the power to make that change, so I did so. 

I'll merge after we merge #338, just to be careful about merging with passing tests and such

dotsdl:
Thank you @WardLT! #338 is now merged, and I've merged `master` here to get those changes. Waiting on the tests now.

PR #350, opened 2022-03-02
Updated docstrings and function annotations for compute and compute_procedure

### LINKED ISSUE #348: SUGGESTION: Cohere `qcng.compute()` call stack naming consistencies and annotations to improve clarity
**Is your feature request related to a problem? Please describe.**
The `qcng.compute()` call stack has a number of naming inconsistencies that could be improved (IMHO) to make the code paths easier to understand. This would help developers like myself to more easily create idiomatic harnesses and implement idiomatic usage of the `qcengine` package in our own software stack.

Specifically:

1. `qcng.compute()` function [has kwarg `local_options`](https://github.com/MolSSI/QCEngine/blob/d9541d641ca5737be0a84c2f29520b72e58a727d/qcengine/compute.py#L34-L40). `local_options` gets merged with `input_data.extras['_qcengine_local_config` and then passed to `get_config()` as seen [here](https://github.com/MolSSI/QCEngine/blob/d9541d641ca5737be0a84c2f29520b72e58a727d/qcengine/compute.py#L76-L83). `get_config()` accepts `local_options` and creates a `TaskConfig` object with defined attributes. The returned `TaskConfig` is passed to `executor.comptue(input_data, config)` as a `TaskConfig` object. `local_options` is a dictionary representation of `TaskConfig` (later referenced in the call stack and harnesses as `config`). I suggest that the dual naming of this configuration object (`local_options` and later referencing it as `config` or `task_config` makes it hard to understand what parameters are passed into `qcng.compute()` and consistent naming would be an improvement :)
2. `qcng.compute()` function annotation suggests it returns `AtomicInput`. In reality it may return `AtomicInput`, `FailedOperation`, or a dictionary of either object as created by `handle_output_metadata`). To pass `mypy` the annotation would need to include all the values returned by `handle_output_metadata` which would include `OptimizationResult` objects as well since this same function handles values returned from `qcng.compute_procedure` too. Ideally, `handle_output_metadata` should be refactored so that two specific functions return only the objects that `qcng.compute()` and `qcng.compute_procedure()` should return.

**Describe the solution you'd like**
1. Rename the top-level `local_options` kwarg for `qcng.compute()` to `task_config`. This clearly identifies the object associated with these values and suggests to the end user what values should live in this dictionary (the attributes of the `TaskConfig` object). Changing the top-level kwarg breaks backwards compatibility so I suggest _adding_ the `task_config` keyword, a user warning/depreciation warning on the `local_options` keyword that can remain in place for a number of versions to alert users to the upcoming change, and then merging `local_options` and `task_config` into a dictionary before passing it to `get_config` to create the `TaskConfig` object. This implementation maintains backwards compatibility (for a time) while creating more clarity about what values are passed in to the top-level `qcng.compute()` function.
2. Change the function annotations on `qcng.compute()` and `qcng.compute_procedure()` to reflect the actual objects one expects these functions to return. This intel is extremely helpful for developers to know the cases they must handle for the various objects these functions may return :) These annotations would not pass `mypy` since the `handle_output_metadata` function may return values for either `compute()` or `compute_procedure()`; however, having them correct in the source code would be beneficial.
3. Ideally someone would refactor `handle_output_metadata` into two/three functions that share common code but respect the true return values for `compute()` and `compute_procedures()`.

**Additional context**
I'd be happy to implement 1 and 2 above. I'd pass on refactoring `handle_output_metadata` as I haven't dug deep into the underlying function calls.

@loriab Here are my thoughts.

coltonbh:
## Description
closes #348 

Still waiting on comment re: updating `local_config` -> `task_config` but wanted to provide these updates before I forgot.

## Changelog description
Updated docstrings and function annotations for compute and compute_procedure. Renamed kwarg `local_config` -> `task_config` to more clearly express that passed options become a `TaskConfig` argument.

## Status

- [x ] Code base linted
- [ x] Ready to go

### DIFF
--- a/qcengine/compute.py
+++ b/qcengine/compute.py
@@ -1,9 +1,10 @@
 """
 Integrates the computes together
 """
+import warnings
 from typing import TYPE_CHECKING, Any, Dict, Optional, Union
 
-from qcelemental.models import AtomicInput, AtomicResult, FailedOperation
+from qcelemental.models import AtomicInput, AtomicResult, FailedOperation, OptimizationResult
 
 from .config import get_config
 from .exceptions import InputError, RandomError
@@ -35,9 +36,10 @@ def compute(
     input_data: Union[Dict[str, Any], "AtomicInput"],
     program: str,
     raise_error: bool = False,
+    task_config: Optional[Dict[str, Any]] = None,
     local_options: Optional[Dict[str, Any]] = None,
     return_dict: bool = False,
-) -> "AtomicResult":
+) -> Union["AtomicResult", "FailedOperation", Dict[str, Any]]:
     """Executes a single CMS program given a QCSchema input.
 
     The full specification can be found at:
@@ -53,15 +55,18 @@ def compute(
         Determines if compute should raise an error or not.
     retries : int, optional
         The number of random tries to retry for.
+    task_config
+        A dictionary of local configuration options corresponding to a TaskConfig object.
     local_options
-        A dictionary of local configuration options
+        Deprecated parameter, renamed to ``task_config``
     return_dict
         Returns a dict instead of qcelemental.models.AtomicResult
 
     Returns
     -------
     result
-        A computed AtomicResult object.
+        AtomicResult, FailedOperation, or Dict representation of either object type
+        A QCSchema representation of the requested output, type depends on return_dict key.
     """
 
     output_data = input_data.copy()  # lgtm [py/multiple-definition]
@@ -73,14 +78,23 @@ def compute(
         # Build the model and validate
         input_data = model_wrapper(input_data, AtomicInput)
 
-        # Build out local options
-        if local_options is None:
-            local_options = {}
+        # Build out task_config
+        if task_config is None:
+            task_config = {}
+
+        if local_options:
+            warnings.warn(
+                "Using the `local_options` keyword argument is deprecated in favor of using `task_config`, "
+                "in version 0.30.0 it will stop working.",
+                category=FutureWarning,
+                stacklevel=2,
+            )
+            task_config = {**local_options, **task_config}
 
         input_engine_options = input_data.extras.pop("_qcengine_local_config", {})
 
-        local_options = {**local_options, **input_engine_options}
-        config = get_config(local_options=local_options)
+        task_config = {**task_config, **input_engine_options}
+        config = get_config(task_config=task_config)
 
         # Set environment parameters and execute
         with environ_context(config=config):
@@ -106,9 +120,10 @@ def compute_procedure(
     input_data: Union[Dict[str, Any], "BaseModel"],
     procedure: str,
     raise_error: bool = False,
+    task_config: Optional[Dict[str, str]] = None,
     local_options: Optional[Dict[str, str]] = None,
     return_dict: bool = False,
-) -> "BaseModel":
+) -> Union["OptimizationResult", "FailedOperation", Dict[str, Any]]:
     """Runs a procedure (a collection of the quantum chemistry executions)
 
     Parameters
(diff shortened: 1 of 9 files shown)


qcengine/util.py
```diff
@@ -146,13 +146,13 @@ def handle_output_metadata(
     metadata: Dict[str, Any],
     raise_error: bool = False,
     return_dict: bool = True,
-) -> Union[Dict[str, Any], "BaseModel"]:
+) -> Union[Dict[str, Any], "AtomicResult", "OptimizationResult", "FailedOperation"]:
```
loriab, line 149: I wonder if it might be clearer here to leave the `-> ... "BaseModel"]:` since BaseModel is mentioned in `output_data`. The likely AtRes, OptRes, FailedOp could be a note in the docstring.

qcengine/compute.py
```diff
@@ -54,14 +54,15 @@ def compute(
...
+        A dictionary of local configuration options corresponding to a TaskConfig object.
     return_dict
         Returns a dict instead of qcelemental.models.AtomicResult
 
     Returns
     -------
     result
-        A computed AtomicResult object.
+        AtomicResult, FailedOperation, or Dict representation of either object type
+        A QC Schema representation of the requested output, type depends on return_dict key.
```
loriab, line 65:
```suggestion
        A QCSchema representation of the requested output, type depends on return_dict key.
```

loriab (commented):
lgtm. It sounds like a release is wanted this week, so do you want to get this PR w/non-breaking changes in for that?

qcengine/util.py
coltonbh: Will try and finish up tonight!

coltonbh: Good point. I think it may be better the other way; i.e., have `output_data` more specifically define the classes that this function will actually return. I agree the discrepancy is confusing!

coltonbh:
I realized the `local_config` nomenclature goes one layer deeper into the [`get_config` function](https://github.com/MolSSI/QCEngine/blob/1a7f3cdc4019e2d7708cc688dc3cd166a8b96718/qcengine/config.py#L270-L323). I updated these functions calls. I'm treating `get_config` as a more internal function and just updating it's signature and internal package calls of the function directly. Are we ok with this choice?

qcengine/util.py
coltonbh: Let me know if you'd like me to go the other way on this. It appears the function is only used in the `compute()` and `compute_procedure()` functions so it should only be returning the values I added the the `ouput_data` annotation.

coltonbh:
@loriab if this looks good to you I'll squash to create a single commit with a cohesive commit message. Any other changes needed let me know :)

qcengine/compute.py
```diff
@@ -53,15 +55,18 @@ def compute(
         Determines if compute should raise an error or not.
     retries : int, optional
         The number of random tries to retry for.
+    task_config
+        A dictionary of local configuration options corresponding to a TaskConfig object.
     local_options
-        A dictionary of local configuration options
+        Depreciated kwarg for task_config
```
loriab, line 61:
```suggestion
        Deprecated parameter, renamed to ``task_config``
```

qcengine/compute.py
```diff
@@ -119,8 +134,10 @@ def compute_procedure(
         The name of the procedure to run
     raise_error : bool, option
         Determines if compute should raise an error or not.
-    local_options : dict, optional
-        A dictionary of local configuration options
+    task_config
+        A dictionary of local configuration options corresponding to a TaskConfig object.
+    local_options
+        Depreciated kwarg for task_config
```
loriab, line 140:
```suggestion
        Deprecated parameter, renamed to ``task_config``
```

loriab (approved):
lgtm! thanks for tracing these types through the code.

PR #353, opened 2022-03-10
Added TeraChemFrontEndHarness. Refactored TeraChemPBSHarness share co…

coltonbh:
…de effectively with TeraChemFrontEndHarness.

## Description
TeraChem's server mode has been extended to support file input/output alongside its protocol buffer interface. This allows more expressive put/get of large files to/from TeraChem via its API. The `TeraChemFrontEndHarness` captures this new functionality and will allow TeraChem to meet additional QCSchema specs (like returning stdout/stderror files, wavefunction data, etc.).

Since the `TeraChemFrontEndHarness` is an extension of the `TeraChemPBSHarness` the latter was refactored to more effectively reuse the shared code.

## Changelog description
Added `TeraChemFrontEndHarness`.

## Status

- [x ] Code base linted
- [ x] Ready to go

### DIFF
--- a/qcengine/programs/base.py
+++ b/qcengine/programs/base.py
@@ -22,6 +22,7 @@ from .qchem import QChemHarness
 from .qcore import EntosHarness, QcoreHarness
 from .rdkit import RDKitHarness
 from .terachem import TeraChemHarness
+from .terachem_frontend import TeraChemFrontEndHarness
 from .terachem_pbs import TeraChemPBSHarness
 from .torchani import TorchANIHarness
 from .turbomole import TurbomoleHarness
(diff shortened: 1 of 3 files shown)


coltonbh:
@loriab an additional harness for `QCEngine`. Not trying to create an absurd proliferation of TeraChem harnesses, but our interfaces to the program are evolving and I want to keep `QCEngine` up to date for use in our cloud compute infrastructure. I don't anticipate adding other harnesses ;P

loriab (approved):
lgtm! thanks for consolidating so much btwn the harnesses. not needed for this PR, but it'd be nice to get tc integration into https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_canonical_config.py where relevant (it may only be for managed_memory).

PR #361, opened 2022-06-10
fix CI and stdsuite: add HF Hessian refs, density-fitted

loriab:
## Description
- [x] As it says on the tin. These aren't tested here (psi4 is the only harness set up for density-fitting), but I have used them in psi4 stdsuite machinery, so format, etc. good.
- [x] Fix CI that had deteriorated after 2 months and a psi release. Had to pin to older d4 temporarily.

## Status

- [x] Code base linted
- [x] Ready to go


devtools/conda-envs/psi.yaml
```diff
@@ -4,6 +4,7 @@ channels:
   - conda-forge
 dependencies:
   - psi4=1.5
+  - libint2=*=hc9558a2_9
```
dotsdl, line 7: Is this tying to a *very specific* build of `libint2` on the `psi4` channel? If so, can we do a version instead?

qcengine/tests/test_procedures.py
```diff
@@ -48,7 +48,7 @@ def test_geometric_psi4(input_data, optimizer):
     # Check keywords passing
     for single in ret.trajectory:
         assert "scf_properties" in single.keywords
-        assert "WIBERG_LOWDIN_INDICES" in single.extras["qcvars"]
+        assert "WIBERG_LOWDIN_INDICES" in single.extras["qcvars"] or "WIBERG LOWDIN INDICES" in single.extras["qcvars"]
```
dotsdl, line 51: Oh no, is either form possible right now?

dotsdl (approved):
Looks good @loriab, but I have some clarification questions. Non-blocking for approval, however.

devtools/conda-envs/psi.yaml
loriab: Yes it is, and sadly, no, it's build conditions, not source versions that matter. Different conditions work with psi through time (https://github.com/psi4/psi4/issues/2442). Psi started pinning build to L2 this spring, so v1.5 is really the most affected.

qcengine/tests/test_procedures.py
loriab: The underscored version has been deprecated since v1.4, and Ben has been pushing to not store two copies. The opt-disp env is hitting the space version, so I figured both would cover all bases. Do you foresee a problem?

dotsdl: No problem, that's enough for me. May want to add in a comment over this line with this info, and perhaps a `TODO` to eventually remove.

PR #363, opened 2022-06-23
dir structure in execute and properties in Cfour

loriab:
## Description
One test fix and some miscellaneous little improvements. The change to `execute` is the only slightly daring one, I think.

## Changelog description
- [x] For Cfour harness, skip explicit `DERIV_LEVEL` setting when `driver = "properties"`. This is needed for DBOC, for example, where any user deriv level (at least 0 or 1) halts calc.
- [x] ~Fix `AtomicResult.native_files` empty test after qcel 0.25.~ Separated to #366 
- [x] Add Wiberg variable note David Dotson suggested in review.
- [x] In `qcengine.execute`, allow directory structure in files to be written to scratch/work directory. This is needed for some Psi4 tests where we copy in a plugin directory.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/cfour/runner.py
+++ b/qcengine/programs/cfour/runner.py
@@ -98,7 +98,7 @@ class CFOURHarness(ProgramHarness):
         opts.update(moldata["keywords"])
 
         # Handle calc type and quantum chemical method
-        mdcopts = muster_modelchem(input_model.model.method, input_model.driver.derivative_int())
+        mdcopts = muster_modelchem(input_model.model.method, input_model.driver)
         opts.update(mdcopts)
 
         # Handle basis set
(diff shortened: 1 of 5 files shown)


qcengine/programs/cfour/runner.py
```diff
@@ -98,7 +98,7 @@ def build_input(
         opts.update(moldata["keywords"])
 
         # Handle calc type and quantum chemical method
-        mdcopts = muster_modelchem(input_model.model.method, input_model.driver.derivative_int())
+        mdcopts = muster_modelchem(input_model.model.method, input_model.driver)
```
WardLT, line 101: Nice. Switching from the integer to the text string is a good change to make on the way

WardLT (approved):
Nice. The changes for the directory file look both useful and save and nice little improvements (e.g., switching from int to string for a if/then statement)

PR #364, opened 2022-06-23
gradients with ghost atoms for cfour and nwchem

loriab:
## Changelog description
- [x] Expand the Xenon hack so that Cfour can run and harvest gradient jobs when molecule has ghost atoms.
- [x] Allow `geometry__autosym` keyword for NWChem. NWChem has a loose symmetrizer that can change the working geometry slightly more than QCEngine expects. Now that can be dialed back.
- [x] Test gradients with ghosts for Cfour (newly works), NWChem (newly matches), GAMESS (already skipped even for energies), and Psi4 (already worked).

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/nwchem/runner.py
+++ b/qcengine/programs/nwchem/runner.py
@@ -179,7 +179,9 @@ class NWChemHarness(ErrorCorrectionProgramHarness):
 
         if opts.pop("geometry__noautoz", False):
             molcmd = re.sub(r"geometry ([^\n]*)", r"geometry \1 noautoz", molcmd)
-        if val := opts.pop("geometry__autosym", False):
+        # someday when minimum >=py38 `if val := opts.pop("geometry__autosym", False):`
+        if opts.get("geometry__autosym", False):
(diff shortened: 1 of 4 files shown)


qcengine/programs/nwchem/runner.py
```diff
@@ -179,6 +179,10 @@ def build_input(
 
         if opts.pop("geometry__noautoz", False):
             molcmd = re.sub(r"geometry ([^\n]*)", r"geometry \1 noautoz", molcmd)
+        # someday when minimum >=py38 `if val := opts.pop("geometry__autosym", False):`
```
WardLT, line 182: Thanks for the line comment! `:=`a syntax I keep forgetting about (probably because many of my projects are "as many Python versions as possible)

WardLT (approved):
LGTM!

qcengine/programs/nwchem/runner.py
loriab: This was going to be my first use for `:=` before I remembered qcng isn't >=py38.

PR #367, opened 2022-06-23
Feature berny handle failed operation v3

### LINKED ISSUE #300: [BUG] berny optimization procedure fails to capture compute errors correctly

coltonbh:
## Description
v3 for you! @loriab you can push to this fork.

Closes #300 
Closes #301 
Closes #365 

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/procedures/berny.py
+++ b/qcengine/procedures/berny.py
@@ -1,11 +1,12 @@
 import logging
+from qcengine.exceptions import UnknownError
 import sys
 import traceback
 from io import StringIO
 from typing import Any, Dict, Union
 
 import numpy as np
-from qcelemental.models import OptimizationInput, OptimizationResult
+from qcelemental.models import OptimizationInput, OptimizationResult, FailedOperation
 from qcelemental.util import which_import
 
 import qcengine
@@ -28,7 +29,9 @@ class BernyProcedure(ProcedureHarness):
     def build_input_model(self, data: Union[Dict[str, Any], "OptimizationInput"]) -> "OptimizationInput":
         return self._build_model(data, OptimizationInput)
 
-    def compute(self, input_data: "OptimizationInput", config: "TaskConfig") -> "OptimizationResult":
+    def compute(
+        self, input_data: "OptimizationInput", config: "TaskConfig"
+    ) -> Union["OptimizationResult", "FailedOperation"]:
         try:
             import berny
         except ModuleNotFoundError:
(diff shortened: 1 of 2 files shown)


loriab (approved):
lgtm, thanks for the fix!

PR #372, opened 2022-07-07
mopac ci lane. pin to qcel 0.25

### LINKED ISSUE #373: mopac test refs
Hi @godotalgorithm, I hear you're the MolSSI consult on Mopac. When you get a chance, please take a look at the changes to test refs in #372 . I wouldn't be surprised if those tests hadn't been run regularly since 2019. I'm comfortable enough with the changes to include in release, but I wanted to keep you in the loop. Thanks.

loriab:
## Changelog description

- [x] add mopac testing to CI since @awvwgk kindly packaged it
  * forgiving the electonic energy key seems ok since it was dropped here https://github.com/openmopac/mopac/commit/d6bdb50bcf74cef84e469097601a4e4944926bee#diff-06ee723679a49e30f3287e5ace72eb84e2e7b0703fb00be2c2de151cfb04e26cL422
  * not too sure about other test loosening
- [x] upcoming changes make it safer to pin tighter to qcel

EDIT: closes #373 

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/opt-disp.yaml
+++ b/devtools/conda-envs/opt-disp.yaml
@@ -8,6 +8,7 @@ dependencies:
   - blas=*=mkl  # not needed but an example of disuading solver from openblas and old psi
   #- intel-openmp!=2019.5
   - rdkit
+  - mopac
 
     # Mixed Tests
   - dftd3 3.2.1
--- a/devtools/conda-envs/qcore.yaml
+++ b/devtools/conda-envs/qcore.yaml
@@ -4,6 +4,7 @@ channels:
   - conda-forge
 dependencies:
   - py-qcore
+  - mopac
 
     # Core
   - python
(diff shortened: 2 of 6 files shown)


loriab:
@godotalgorithm, any apprehensions about the test changes?

godotalgorithm:
This looks ok to me - you are now getting MOPAC from conda-forge (which I will be actively maintaining), some reference numbers have been changed, and the "electronic energy" was removed.

The "electronic energy" has been confusing people (it is NOT a total electronic energy for the molecule), so it has been removed from the default output (it is still accessible with a keyword) and I'd prefer not to emphasize it. The forces produced by MOPAC are derivatives of the "heat of formation", so that is the primary energetic quantity of relevance.

The reference numbers (they seem like geometric data for water?) have shifted a bit, but that isn't surprising of an optimized geometry as MOPAC's default geometric convergence isn't that tight and a lot of small changes have happened behind the scenes in the last few years to perturb these results.

awvwgk:
Maybe also update the suggestion on how to install Mopac at

https://github.com/MolSSI/QCEngine/blob/43487205d114de4760ea64fd0fa46652d31eda0d/qcengine/programs/mopac.py#L45

PR #375, opened 2022-07-30
stdsuite testing remp2, oo, dfcc

loriab:
## Description
This pairs with https://github.com/psi4/psi4/pull/2633. It doesn't affect QCEngine's general purpose, so I'll plan to mint a v0.24.1 with this when the Psi4 use is finalized.

Adds ref values for df ccsd, ccsd(t), a-ccsd(t), for orbital-optimized mp2, mp2.5, mp3, ccd, ccd(t), a-ccd(t), and for remp2 and oremp2. Existing reference values computed with Psi4's dfocc and occ modules have been updated now that it can converge tighter.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/tests/standard_suite_contracts.py
+++ b/qcengine/programs/tests/standard_suite_contracts.py
@@ -691,21 +691,41 @@ def contractual_lccd(
         expected = True
         if (
             (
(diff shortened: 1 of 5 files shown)


WardLT (approved):
LGTM. Don't fully follow it, but it looks reasonable

PR #385, opened 2022-11-22
adapt s-dftd3 harness for psi4

loriab:
## Description
- [x] Changes the dispersion coefficient resources for D3 variants so that there's a `2b` set (e.g., `d3bj2b`) that is pure 2-body and doesn't accept `s9` (effectively fixed at 0.0) and a `atm` set (e.g., `d3zeroatm`) that does accept `s9` (by default 1.0 but user-variable).
- [x] Previous aliases are mapped to `2b`, so classic dftd3 doesn't change. s-dftd3 tests remain unchanged but if keyword `apply_qcengine_aliases` is True (psi4 does this), the level hints are interpreted into standard s-dftd3 input.
- [x] @awvwgk, this is my attempt at transitioning from dftd3 -> s-dftd3 while (1) not breaking Psi4's d3 behavior, (2) not locking Psi4's d3 into s9=0.0, (3) not requiring full parameter spec just to tweak s9, and (4) not overwriting your reasonable and released defaults and behavior for the s-dftd3 interface. It's not as clean as I would like, but I think it's workable. Any comments, concerns, or questions?
- [x] fixes classic dftd3 harness pairwise analysis array to be scaled like s-dftd3. see changelog for details.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_canonical_config.py
+++ b/qcengine/programs/tests/test_canonical_config.py
@@ -102,7 +102,7 @@ def test_local_options_memory_gib(program, model, keywords, memory_trickery, req
     #  <<  Run
 
     inp = AtomicInput(molecule=molecule, driver="energy", model=model, keywords=use_keywords)
-    ret = qcng.compute(inp, program, raise_error=True, local_options=config.dict())
+    ret = qcng.compute(inp, program, raise_error=True, task_config=config.dict())
     pprint.pprint(ret.dict(), width=200)
     assert ret.success is True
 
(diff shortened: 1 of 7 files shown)


qcengine/programs/tests/test_canonical_config.py
```diff
@@ -243,7 +243,7 @@ def test_local_options_ncores(program, model, keywords, ncores):
     #  <<  Run
 
     inp = AtomicInput(molecule=molecule, driver="energy", model=model, keywords=keywords)
-    ret = qcng.compute(inp, program, raise_error=True, local_options=config.dict())
+    ret = qcng.compute(inp, program, raise_error=True, task_config=config.dict())
```
WardLT, line 246: Good call to change these deprecated arguments while at it.

WardLT (approved):
LGTM from the code side. Nice work on making code quality improvements (e.g., adding typing, improving tests) while making the changes.

PR #388, opened 2022-12-02
Save parallel set up for MRChem in provenance

robertodr:
## Description
The harness saves the parallel set up for the MRChem calculation in the `provenance` dictionary. 

## Changelog description

- The MRChem harness saves more detailed info about the parallel set up in the `provenance` dictionary of the output.

## Status

- [X] Code base linted
- [X] Ready to go

### DIFF
--- a/qcengine/programs/mrchem.py
+++ b/qcengine/programs/mrchem.py
@@ -101,6 +101,7 @@ class MRChemHarness(ProgramHarness):
 
         job_input = self.build_input(input_model, config)
         input_data = copy.deepcopy(job_input["mrchem_json"])
+
         output_data = {
             "keywords": input_data,
             "schema_name": "qcschema_output",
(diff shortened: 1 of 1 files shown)


qcengine/programs/mrchem.py
```diff
@@ -132,6 +133,10 @@ def compute(self, input_model: "AtomicInput", config: "TaskConfig") -> "AtomicRe
                 mrchem_output = mrchem_json["output"]
                 output_data["success"] = mrchem_output["success"]
                 output_data["provenance"] = mrchem_output["provenance"]
+                # parallel set up
+                output_data["provenance"].update(job_input["parallel"])
+                # remove nthreads, above info is more detailed anyway
+                output_data["provenance"].pop("nthreads")
```
loriab, line 139: It's troubled me since I first read it that `nthreads` gets removed. Even though `total_cores` added below is the more detailed counterpart, people may expect the field. Do you think it should (a) be left to un-detailed value, (b) have `total_cores` copied in, or (c) definitely remove the field?

loriab: Let's add for now, pending trouble.

loriab (approved):
My only concern is dropping nthreads. Thanks for the addition.

qcengine/programs/mrchem.py
robertodr: Hei Lori, sorry for not replying earlier. The reason to drop it was that the information in `job_input["parallel"]` reflects more accurately how the MPI ranks and OpenMP threads were distributed for the computation. MRChem allocates these resources a bit peculiarly...

Issue #292, opened 2021-03-10
qcengine info psi4 harness raises exception if system uses pyenv


coltonbh:
**Describe the bug**
Running `qcengine info` raises the following error if a system has installed python versions using `pyenv`.
```
❯ qcengine info
>>> Version information
QCEngine:    v0.18.0+1.g02aa5ef
QCElemental: v0.18.0

>>> Program information
Traceback (most recent call last):
  File "/Users/cbh/dev/mtzgroup/QCEngine/env/bin/qcengine", line 33, in <module>
    sys.exit(load_entry_point('qcengine', 'console_scripts', 'qcengine')())
  File "/Users/cbh/dev/mtzgroup/QCEngine/qcengine/cli.py", line 162, in main
    info_cli(args)
  File "/Users/cbh/dev/mtzgroup/QCEngine/qcengine/cli.py", line 125, in info_cli
    info_programs()
  File "/Users/cbh/dev/mtzgroup/QCEngine/qcengine/cli.py", line 84, in info_programs
    avail_progs = list_available_programs()
  File "/Users/cbh/dev/mtzgroup/QCEngine/qcengine/programs/base.py", line 97, in list_available_programs
    if p.found():
  File "/Users/cbh/dev/mtzgroup/QCEngine/qcengine/programs/psi4.py", line 64, in found
    sys.path.append(exc["stdout"].split()[-1])
IndexError: list index out of range
```

**To Reproduce**
1. Install pyenv
2. Install a miniconda/anaconda version
3. Create a conda env and install psi4 inside it
2. Run `qcengine info` when this environment is _not_ activated

**Expected behavior**
I would expect `qcengine` to simply omit `psi4` from the available programs. 

**Additional context**
The issue is raised due to the `exc` variable having the following value:
```
{'proc': <Popen: returncode: 127 args: ['/Users/cbh/.pyenv/shims/psi4', '--module']>, 'stdout': '', 'stderr': "pyenv: psi4: command not found\n\nThe `psi4' command exists in these Python versions:\n  miniconda3-4.7.12/envs/p4env\n\nNote: See 'pyenv help global' for tips on allowing both\n      python2 and python3 to be found.\n"}
```
which does not include a `stdout` value that can be split.

loriab:
[...]

PR #389, opened 2022-12-05
Prepend psiimport path instead of overwriting `$PYTHONPATH`

Flamefire:
## Description

When `psi4` binary is found but the Python package is not we should not overwrite the `$PYTHONPATH` before importing `psi4` as that may cause other required packages to be not found anymore. Hence prepend it.

See https://github.com/MolSSI/QCEngine/issues/292#issuecomment-1337320606

## Changelog description

Fix `IndexError` in PSI4 Harness when `psi4` Python package is not found but the binary is.

### DIFF
--- a/qcengine/programs/psi4.py
+++ b/qcengine/programs/psi4.py
@@ -66,7 +66,9 @@ class Psi4Harness(ProgramHarness):
         if psiapi and not psithon:
             psiimport = str(Path(which_import("psi4")).parent.parent)
             env = os.environ.copy()
-            env["PYTHONPATH"] = psiimport
+            pythonpath = env.get("PYTHONPATH")
+            # Prepend psiimport to $PYTHONPATH (carefull if it's empty/unset)
+            env["PYTHONPATH"] = os.pathsep.join([psiimport, pythonpath] if pythonpath else [psiimport])
             with popen(["python", "-c", "import psi4; print(psi4.executable[:-5])"], popen_kwargs={"env": env}) as exc:
                 exc["proc"].wait(timeout=30)
             os.environ["PATH"] += os.pathsep + exc["stdout"].split()[-1]


loriab:
Thanks, I think I agree. In answer to your 

> So why not simply import psi4 and directly access psi4.executable?

it's because `which_import` finds the location w/o actually importing. so the pythonpath adjustment is to get enough dependencies locatable so that `import psi4` can work and `psi4.executable` can be accessed.

Flamefire:
> it's because `which_import` finds the location w/o actually importing. so the pythonpath adjustment is to get enough dependencies locatable so that `import psi4` can work and `psi4.executable` can be accessed.

And that's what I don't understand: You find the location w/o importing to then adjust the PYTHONPATH with the found location to import it in a subprocess. But the location can only be found if PYTHONPATH already contains it so the import would work. So why changing it? Or did I miss how `which_import` works?

loriab:
>> it's because which_import finds the location w/o actually importing. so the pythonpath adjustment is to get enough dependencies locatable so that import psi4 can work and psi4.executable can be accessed.

> And that's what I don't understand: You find the location w/o importing to then adjust the PYTHONPATH with the found location to import it in a subprocess. But the location can only be found if PYTHONPATH already contains it so the import would work. So why changing it? Or did I miss how which_import works?

The overall aim for this `def found()` is to make sure a _paired_ `bin/psi4` and `psi4/__init__` are found and ready to load. There's some configuring that the program knows above, so it's best to use `psi4 --module` and `psi4.executable`. https://github.com/psi4/psi4/blob/master/psi4/__init__.py#L38-L42 The overall pattern for the QCEngine harnesses is that `found()` looks for the installation (but doesn't run or import) and `version()` does a simple run. https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/nwchem/runner.py#L73-L87

You're correct that `which_import` is finding the signature of the psi4 package but not actually trying to import it (https://github.com/MolSSI/QCElemental/blob/next/qcelemental/util/importing.py#L36). So when that returns `True`, it may still be that not all dependencies required for psi4 are importable. My first thought (and how I answered above) was that the `env` modification was to add a likely dependency location to `sys.path` so that `import psi4` worked and `psi4.executable` could be accessed. But I entirely agree with your analysis that that is redundant. My second thought was that the `PYTHONPATH` truncation for `env` was to _restrict_ paths so that, say a `psi4.py` couldn't interfere with the targeted psi4 installation. But I'm not thinking of what scenario would hit that. I've removed the env complication in #391 and will try that locally for a few days. 

Would that satisfy your intent the same as this PR? Hopefully I'll have a bright idea (or hit a bug) in the meantime about what I was doing.

Flamefire:
> Would that satisfy your intent the same as this PR? Hopefully I'll have a bright idea (or hit a bug) in the meantime about what I was doing.

But why the subprocess? What about the more common:

```
try:
  import psi4
except ImportError:
  return None
return psi4.executable[:-5]
```

If you still want the subprocess please include at least error handling, see the linked comment:

> So I'd strongly suggest to add any kind of error checking first before accessing `exc["stdout"]`

I.e. at least check the exit code and possibly that `exc["stdout"]` is not empty which both suggests that the import failed.

loriab:
@Flamefire, in case you're still monitoring this issue, I think I've got it fixed at #418 . If you see any deficiencies, please let me know.

Flamefire:
> @Flamefire, in case you're still monitoring this issue, I think I've got it fixed at #418 . If you see any deficiencies, please let me know.

Looks good I think. If you want to be safe I'd also check the exit code as suggested in my prior comment, i.e. `exc["proc"].wait(timeout=30) == 0`

loriab:
> Looks good I think. If you want to be safe I'd also check the exit code as suggested in my prior comment, i.e. exc["proc"].wait(timeout=30) == 0

Thanks, done, and I caught another bug along the way.

PR #394, opened 2023-01-05
fix qcore and adcc CI

loriab:
## Description

* according to https://software.entos.ai/qcore/tutorial/ , entos needs a tbb constraint. that fixes the CI error that was showing up at license check time
* @maxscheurer, I think what was happening with the h5 error in adcc CI (e.g., https://github.com/MolSSI/QCEngine/actions/runs/3847917831/jobs/6555003472) was that psi needs hdf5 1.10, and with c-f higher in channel priority, it was having to grab old h5py 3.3 since c-f switched to 1.12 earlier. Now it's getting 1.10-based hdf5 packages from defaults and adcc itself from c-f instead of adcc channel. Presently adcc 0.15.14 and psi4 1.7.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/adcc.yaml
+++ b/devtools/conda-envs/adcc.yaml
@@ -1,10 +1,10 @@
 name: test
 channels:
-  - adcc
+  #- adcc
   - psi4/label/dev
   - conda-forge
 dependencies:
-  - adcc>=0.15.7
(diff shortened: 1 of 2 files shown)


maxscheurer:
adcc channel is discontinued anyways, so the switch is worthwhile 👍🏻

PR #396, opened 2023-01-24
canonical smiles string with rdkit plus unit testing

dr-marsmm:
## Description

QCEngine can now generate canonical smiles string (with stereochemistry) using RDKit

## Changelog description

canonical smiles string with rdkit plus unit testing
## Status

- [X] Code base linted
- [X] Ready to go


qcengine/stock_mols.py
```diff
@@ -9,22 +9,19 @@
 _test_mols = {
     "hydrogen": {
         "symbols": ["H", "H"],
-        "geometry": [0, 0, -0.65, 0.0, 0.0, 0.65],
+        "geometry": [
+            [0, 0, -0.65],
+            [0.0, 0.0, 0.65]
```
loriab, line 14: `geometry` field in qcsk is flat list. why the conversion to nested?

dr-marsmm: Readability.  Probably wasn't necessary for H2, but I was on autopilot.  I know the list gets flattened in schema, but stock_mols aren't schema and nested both makes more sense (len(geom) = len(symbols)) and is more readable.  I'll be adding a few more things in there and it was just bugging me.
Is problem?

PR #400, opened 2023-04-04
Allow specifying task config via environment and CLI

bennybp:
## Description

This allows for setting the task config (number of nodes, memory, etc) via the CLI and by environment variables.

(I am looking to run more via the CLI than the python interface, for various reasons.)

## Changelog description

Task configuration can now be set via CLI and environment variables

## Status

- [X] Code base linted
- [ ] Ready to go


loriab (approved):
looks and tests good to me, thanks! evading the `parse_environment` is a little troubling, but the tests involving `$envvar` still work, so ok.

PR #405, opened 2023-06-12
Some Turbomole fixes

cvsik:
## Description

This PR incorporates some fixes/additions for the Turbomole runner:
- Pass SCF convergence and SCF max. iterations to `define`
- Correctly enable OMP
- Pass `environment` to `execute`

@loriab I'm actually @maxscheurer but from my corporate GH account 😄 

## Questions
- [x] Is the new test okay using `xfail`?
- [ ] How should one allow a user-specified scratch directory? Is `TaskConfig` -> `execute` the correct route?
- [ ] I changed the (unusual?) `jobs_per_node` default from `2` to `1`, is that okay, or was there a reason for it being set to `2` specifically?
- [ ] `make format` reformats almost all files...

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -84,7 +84,7 @@ class NodeDescriptor(pydantic.BaseModel):
     
     The default value, ``None``, will allow QCEngine to autodetect the number of cores.""",
     )
-    jobs_per_node: int = 1
+    jobs_per_node: int = 2
     retries: int = 0
 
     # Cluster options
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -84,7 +84,7 @@ class NodeDescriptor(pydantic.BaseModel):
     
     The default value, ``None``, will allow QCEngine to autodetect the number of cores.""",
     )
-    jobs_per_node: int = 2
+    jobs_per_node: int = 1
     retries: int = 0
 
     # Cluster options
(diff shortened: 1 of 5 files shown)


loriab:
> Is the new test okay using xfail?

sure, that's fine

> How should one allow a user-specified scratch directory? Is TaskConfig -> execute the correct route?

Yes, that's it. User-side here: https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_canonical_config.py#L160-L161 . Code side here (first and last lines): https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/cfour/runner.py#L85-L154 . The former is a something of a standard test you could add turbomole to sometime.

> I changed the (unusual?) jobs_per_node default from 2 to 1, is that okay, or was there a reason for it being set to 2 specifically?

I'm leery of that change. Was it causing trouble as 2? I don't know why it's not 1 (admittedly more usual) -- maybe 2 avoided only serial testing?

> make format reformats almost all files...

I've squashed that now :-)


Anyways, all lgtm. After 414 is in and tests can pass, I can adapt this PR. Main question is jobs_per_node, I guess. Thanks!

cvsik:
> I'm leery of that change. Was it causing trouble as 2? I don't know why it's not 1 (admittedly more usual) -- maybe 2 avoided only serial testing?

IIRC if you don't specify anything, then this divides the number of cores for the job by 2, which I think is a bit unexpected. 😄

loriab:
The `jobs_per_node` traces back to here https://github.com/dgasmith/QCEngine/commit/41e7da039102195a5f17e5526a485f5cdefc6bff . I still don't see why it's 2, unless just to have something multiprocess to test against. 

Changing to 1 causes a gamess test to fail, but not for a good reason. Still investigating ...

loriab:
Ok, see #415. I think the `jobs_per_node` should be reset to 1, but not in isolation. Do you agree? If so, you or I can remove it, then we'll get this one in, and maybe openmm will cooperate in CI.

loriab (approved):
lgtm turbomole-wise, thanks!

I've removed the `jobs_per_node` part and will open a separate PR for that.

PR #409, opened 2023-07-27
Psi4 random fail restarts 

### LINKED ISSUE #407: Psi4 `_handle_output` stops retries
**Describe the bug**

Task retries are triggered only if the error raised from the [compute call](https://github.com/MolSSI/QCEngine/blob/1887e7119e4a4b2fd836cae410227efcfefc4588/qcengine/compute.py#L103-L107) is a  QCEngine `RandomError`. The Psi4 harness has a [method](https://github.com/MolSSI/QCEngine/blob/1887e7119e4a4b2fd836cae410227efcfefc4588/qcengine/programs/psi4.py#L102) to try and classify some failure modes, with general random failures with no error message given a new generic error message which does not contain the required keywords to trigger a retry. This is problematic in procedures like torsiondrives or optimizations as the entire calculation is stopped when in practice they normally work fine after retrying the single point calculation which has failed.

I think simply changing the [error message](https://github.com/MolSSI/QCEngine/blob/1887e7119e4a4b2fd836cae410227efcfefc4588/qcengine/programs/psi4.py#L111) to include the correct keyword ([segmentation fault](https://github.com/MolSSI/QCEngine/blob/1887e7119e4a4b2fd836cae410227efcfefc4588/qcengine/programs/psi4.py#L261)) would fix this. 

**To Reproduce**

Seems to happen when running a pool of workers executing psi4 single points on the same computer, see attached qcschema temp file which just ends suddenly when computing the gradient. 

[qcschema.txt](https://github.com/MolSSI/QCEngine/files/12147502/qcschema.txt)

**Expected behavior**

A psi4 task which fails with no error message should be restarted.

**Additional context**

jthorton:
## Description

This PR fixes #407 by adding a valid error message which raises a `RandomError` needed to restart the calculations.

## Changelog description

Psi4 tasks are now correctly restarted when there is no valid error message.

## Status

- [ ] Code base linted
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/openmm.yaml
+++ b/devtools/conda-envs/openmm.yaml
@@ -16,7 +16,7 @@ dependencies:
   - py-cpuinfo
   - psutil
   - qcelemental >=0.11.1
-  - pydantic >=0.30.1
+  - pydantic >=1.8.2
 
     # Testing
   - pytest
--- a/devtools/conda-envs/qcore.yaml
+++ b/devtools/conda-envs/qcore.yaml
@@ -11,7 +11,7 @@ dependencies:
   - py-cpuinfo
   - psutil
   - qcelemental >=0.24
-  - pydantic >=0.30.1
+  - pydantic >=1.8.2
   - tbb<2021
 
     # Testing
--- a/devtools/conda-envs/xtb.yaml
+++ b/devtools/conda-envs/xtb.yaml
@@ -11,7 +11,7 @@ dependencies:
   - py-cpuinfo
   - psutil
   - qcelemental >=0.11.1
-  - pydantic >=0.30.1
+  - pydantic >=1.8.2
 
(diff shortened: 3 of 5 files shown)


jthorton:
Ah just noticed there is a `programs/tests` folder should I move the new psi4 test to `programs/tests/test_programspy` ?

loriab:
> Ah just noticed there is a programs/tests folder should I move the new psi4 test to programs/tests/test_programspy ?

No need, there's no particular distinction in what goes where for tests.

(I'll patch up the branch after 414 gets merged. That'll solve the black/lint fail, too.)

loriab (approved):
Thanks, I learned about pytest monkeypatch. I'll be removing all the pydantic constraints, but the psi4 part lgtm.

PR #411, opened 2023-07-31
Enable Pydantic v2 using v1 API

Lnaden:
## Description
As it says. This is a stopgap until we can properly implement Pydantic v2.

there was also a small change in the way `cpuinfo` does its `brand` key, its now `brand_raw` but I changed it to try both. 
Part of the larger QCA upgrade

## Changelog description
Enable Pydantic v2 support

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/config.py
+++ b/qcengine/config.py
@@ -9,8 +9,10 @@ import os
 import socket
 from typing import Any, Dict, Optional, Union
 
-import pydantic
-
+try:
+    import pydantic.v1 as pydantic
+except ImportError:
+    import pydantic
 from .extras import get_information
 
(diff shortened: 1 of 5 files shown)


loriab:
Changes look good, thanks for rolling them out to qcng! How do you come to have conflicts with a new PR?

Lnaden:
Huh... I don't know. I'll look into this. Probably something to do with some bad merge somewhere.

Lnaden:
So this PR was branched from like right after 0.11 on my local copy. Whoops. This is correctly done in #412.

PR #412, opened 2023-07-31
Use Pydantic v1 API even with v2

Lnaden:
## Description
The correct version of #411. I had originally had branched from a point way in the past.

Supersedes #411

## Status

- [ ] Code base linted
- [ ] Ready to go


loriab:
I bet https://github.com/MolSSI/QCElemental/pull/305 has to do with the QChem QCEngineRecords trouble. https://github.com/MolSSI/QCElemental/pull/315 may be on track to solution or just related.

PR #414, opened 2023-07-31
Pydanticv2again

loriab:
iterating on #412 -- I was going to PR to that, but this PR grew, so better to collect comments here.

## Description

- [x] Apply Levi's pydantic v1/v2 tolerant import approach. qcng is still using v1 API  closes #412
- [x] Update GHA versions for checkout and codecov and switch GHA from conda to mamba for speedier solves and more comprehensive can't-solves
- [x] Update QCEngineRecords to a commit that also has Levi's pydantic v1/v2 tolerant approach. Add an excuse for QChem molecules for None != {} resulting from https://github.com/MolSSI/QCElemental/pull/305
- [x] Pin black to year 2022, since that's before the great blank line removing kerfluffle, which apparently can't be compromised on until 2024.
- [x] I saw a case where it was sometimes "Driver gradient not implemented" and sometimes "Driver DriverEnum.gradient not implemented", so pass either since it's just a check.
- [x] Fixed a couple Sphinx complaints, so it'll build again. Note that Sphinx is restrained to pydantic=1 for autodoc-pydantic package's sake.
- [x] Tweaked some environment spec files. Note that any with psi4 could use an overhaul b/c moved to c-f. But qcel, qcng, and psi4 are so co-dependent, we need to get pydantic straightened out first.
- [x] Edit versioneer according to plan in https://github.com/MolSSI/QCElemental/issues/313 so that it hopefully works with Python 3.12

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,56 +17,56 @@ jobs:
       fail-fast: false
       matrix:
         cfg:
-          #- conda-env: psi
-          #  python-version: 3.7
-          #  label: Psi4-release
-          #  runs-on: ubuntu-latest
+          - conda-env: psi
+            python-version: 3.7
+            label: Psi4-release
+            runs-on: ubuntu-latest
 
-          #- conda-env: psi-nightly
-          #  python-version: "3.10"
-          #  label: Psi4-nightly
-          #  runs-on: ubuntu-latest
+          - conda-env: psi-nightly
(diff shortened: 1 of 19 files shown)


Lnaden (approved):
This all looks awesome. Thanks for all the hard testing on this, I know its been a pain.

PR #418, opened 2023-08-07
overhaul psi4 detection parsing

### LINKED ISSUE #292: qcengine info psi4 harness raises exception if system uses pyenv

loriab:
## Description
The existing psi4 detection has worked great when psi4 is in order and conda environments prevail, but #389 and #292 show that if something is set up a little wrong with psi4, then everything devolves into an IndexError with the existing psi4 detection.

This PR improves matters by
* in the complete-the-psithon-psiapi-pair blocks:
  * only doing string-safe operations on stdout
  * forming clearer error responses
  * catching stderr instead of tossing it and parsing stdout anyways
  * getting rid of `env` passing. As @Flamefire pointed out, the psi4 module is already in sys.path or which_import wouldn't have worked. unlike #389, code is now passing no env so current env is used (and can provide numpy, etc.). I've tested that there's no need to provide env=os.environ to get current env.
  * checking returncode
  * it was also not showing error messages when psithon found but not psiapi b/c final check was of psithon only. now checking at the weak point
* in the version fn:
  * only doing string-safe operations on stdout
  * catching stderr instead of tossing it and parsing stdout anyways
  * checking returncode

Hopefully this will be more robust. closes #389 closes #391 closes #292

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/procedures/optking.py
+++ b/qcengine/procedures/optking.py
@@ -20,7 +20,7 @@ class OptKingProcedure(ProcedureHarness):
             "optking",
             return_bool=True,
             raise_error=raise_error,
-            raise_msg="Please install via `conda install optking -c psi4/label/dev`.",
+            raise_msg="Please install via `conda install optking -c conda-forge`.",
         )
 
     def build_input_model(self, data: Union[Dict[str, Any], "OptimizationInput"]) -> "OptimizationInput":
--- a/qcengine/programs/adcc.py
+++ b/qcengine/programs/adcc.py
@@ -50,13 +50,13 @@ class AdccHarness(ProgramHarness):
             "adcc",
             return_bool=True,
             raise_error=raise_error,
-            raise_msg="Please install via `conda install adcc -c adcc`.",
+            raise_msg="Please install via `conda install adcc -c conda-forge`.",
         )
         found_psi4 = which_import(
             "psi4",
             return_bool=True,
             raise_error=raise_error,
-            raise_msg="Please install psi4 for adcc harness via `conda install psi4 -c psi4`.",
+            raise_msg="Please install psi4 for adcc harness via `conda install psi4 -c conda-forge/label/libint_dev -c conda-forge`.",
         )
         return found_adcc and found_psi4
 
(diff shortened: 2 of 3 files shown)


loriab:
Yes, if there was any good reason behind that env reset, I haven't been able to reconstruct it. It'll be nice to get good traceback back, rather than remembering that IndexError in psi4 found() means the version/tag is broken.

Issue #399, opened 2023-03-01
Importing qcengine breaks OpenMPI programs


awvwgk:
**Describe the bug**

Importing qcng modifies the environment and can break running all OpenMPI programs.

**To Reproduce**

````python
import subprocess
subprocess.run("printenv > previous.txt", shell=True)
import qcengine as qcng
subprocess.run("printenv > later.txt", shell=True)
````

````diff
--- previous.txt        2023-03-01 11:21:03.382801986 +0000
+++ later.txt   2023-03-01 11:21:12.898774636 +0000
@@ -2,0 +3 @@
+OMPI_MCA_pmix=^s1,s2,cray,isolated
@@ -5,0 +7 @@
+OMPI_MCA_ess=singleton
@@ -7,0 +10 @@
+PMIX_RANK=0
@@ -14,0 +18,4 @@
+PMIX_NAMESPACE=1256783873
 ...
````

**Expected behavior**

Modify environment transparently in the qcng provided `util.execute` wrapper, with possibility to turn off if unwanted.

**Additional context**

````
python                    3.10.9          he550d4f_0_cpython    conda-forge
python_abi                3.10                    3_cp310    conda-forge
qcengine                  0.26.0             pyhd8ed1ab_0    conda-forge
qcelemental               0.25.1             pyhd8ed1ab_1    conda-forge
````

loriab:
Yes, I see it. Easy to avoid, but a completely robust solution is harder since MDI is in a class by itself.

PR #401, opened 2023-04-06
Fix MDI forces structure

taylor-a-barnes:
## Description
This is a simple fix to ensure that the MDI interface properly formats the forces prior to communication.

## Changelog description
Fix bug in the shape of the MDI forces structure

## Status

- [x] Code base linted
- [x] Ready to go

PR #420, opened 2023-08-08
defer MPI in MDI

loriab:
## Description

This addresses #399. If one has `pymdi` and `mpi4py` packages in the environment, then merely `import qcengine` adds a bunch of potentially interfering mpi envvars. The culprit in particular is `from mpi4py import MPI`.

This PR solves #399 by deferring import of mpi4py until the `MDIServer` is created. But (1) I don't know if that interferes with MDI, hence the @taylor-a-barnes ping (though he's out of the office for a while). Also (2) since MDI isn't a harness like all other programs or procedures, its code is more exposed and can't really be hidden behind `execute()`. Thus, the problem is still present if MDI has ever been invoked, and qcengine can still interfere with other, potentially non-qcng-controlled processes. Probably there should at least be a toggle to turn off MDI MPI even if mpi4py is present.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/mdi_server.py
+++ b/qcengine/mdi_server.py
@@ -31,12 +31,7 @@ try:
 except ImportError:
     use_mdi = False
 
-try:
-    from mpi4py import MPI
-
-    use_mpi4py = True
-except ImportError:
-    use_mpi4py = False
+use_mpi4py = which_import("mpi4py", return_bool=True)
 
 
 class MDIServer:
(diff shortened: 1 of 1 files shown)


loriab:
aha, the mpi4py is already gone in #401, so that fixes #399 nicely.

loriab:
rendered unnecessary by #401

Issue #377, opened 2022-09-30
7 tests fail


yurivict:
**Describe the bug**
[qcengine-test-failures.txt](https://github.com/MolSSI/QCEngine/files/9681115/qcengine-test-failures.txt)

Python 3.9
FreeBSD 13.1 STABLE

loriab:
The first (and most alarming looking) failure, the `scf ene: computed value (-1.7055776) does not match (-1.6561385) ` has to do with nwchem versions. The root that nwc finds varies btwn versions, so it fails for me locally, too, but passes with the c-f version in CI. Not a great test case, I admit, but nothing to be concerned about.

The geometric or rdkit fails look like error msg formatting of some kind. Those would need investigation.

Thanks for report.

yurivict:
For the version 0.25.0 one test fails: [qcengine-test-failures.txt](https://github.com/MolSSI/QCEngine/files/10006887/qcengine-test-failures.txt)

yurivict:
0.27.0 has 31 test failures: [qcengine-0.27.0-tests.log](https://github.com/MolSSI/QCEngine/files/12249001/qcengine-0.27.0-tests.log)

loriab:
Thanks for the update. Most or all are `gcp` errors? Is there any chance it's detecting the wrong gcp -- there's a system one (https://manpages.ubuntu.com/manpages/trusty/man1/gcp.1.html). Of the proper "geometric counter poise" project, there's two implementations -- the original with executable `gcp` and the newer preferred one with executable `mctc-gcp`. Does that provide any clues as to what's going wrong?

yurivict:
Yes, the gcp executable was unrelated.

Now with the ```mctc-gcp``` there are 2 failures. The first failure was a runaway process that I killed after many hours. The second failure is a regular test failure.


[log](https://people.freebsd.org/~yuri/qcengine-0.27.0-test.log)

loriab:
Is the "regular test failure" the nwchem one with `scf ene: computed value (-1.7055776) does not match (-1.6561385) to atol=3e-06 by difference (-0.0494391).`? That one I know about -- the SCF root that nwchem finds differs by nwchem version.

yurivict:
Now I am rerunning the tests and
* ```test_hf_energy_module[hf  uhf ae: psi4-dz-ene0]``` hangs for a very long time so I have to kill it
[...]

PR #423, opened 2023-08-08
cleanup: known test fault, codeql

loriab:
## Description

* there's a highly symmetric system that different versions of nwchem return a different root for. I've kept it at the one our CI uses, but others (#377) have hit it. so let's allow it.
* fix a couple codeql complaints to see what that looks like.

## Status

- [x] Code base linted
- [ ] Ready to go

### DIFF
--- a/qcengine/programs/tests/test_ghost.py
+++ b/qcengine/programs/tests/test_ghost.py
@@ -44,7 +44,7 @@ def test_simple_ghost(driver, program, basis, keywords, hene):
 
     if program == "gamess":
         with pytest.raises(qcng.exceptions.InputError) as e:
-            res = qcng.compute(resi, program, raise_error=True, return_dict=True)
+            qcng.compute(resi, program, raise_error=True, return_dict=True)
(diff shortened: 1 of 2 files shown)


WardLT (approved):
LGTM. 

Is the ".pop" the thing CodeQL complained about? Any idea what the problem with the one-line version is?

loriab:
> Is the ".pop" the thing CodeQL complained about? Any idea what the problem with the one-line version is?

Yeah, CodeQL didn't like the unused `res` in `res = qcng.compute(resi, program, raise_error=True, return_dict=True)`. And it didn't like that there were side-effects (pop) in a statement that could be optimized out (assert) in `assert ret.extras.pop("psiapi_evaluated", False)`. Those two I at least understand, but many of CodeQL's complaints are like this https://github.com/MolSSI/QCEngine/security/code-scanning/156 where I've clearly made a copy before altering input parameters, so if anyone understands the trouble or a solution, I'd be glad to hear of it.

PR #424, opened 2023-08-11
SVWN reference Hessians

loriab:
## Description
After tightening screening (https://github.com/psi4/psi4/pull/2834) and adding analytic UHF Hessians (https://github.com/psi4/psi4/pull/2992), Psi4 has generated better references for SVWN Hessians, so replacing. (Originals were from psi4 also.)

## Status

- [x] Code base linted
- [x] Ready to go


WardLT (approved):
Is it worth noting which version of Psi4 was used to compute these references?

loriab:
Good idea, done!

PR #441, opened 2024-01-05
Add a harness for MACE-OFF23 and local MACE models

jthorton:
## Description

This PR adds a new harness for ML models based on the MACE architecture, supporting local models and three general organic force field published models (MACE-OFF23) which can be found [here](https://github.com/ACEsuit/mace-off). More information about the performance can be found [here](https://doi.org/10.48550/arXiv.2312.15211).

## Changelog description

Add new harness for MACE ML models supporting local and the general MACE-OFF23 models.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -98,6 +98,12 @@ jobs:
             runs-on: windows-latest
             pytest: "-k 'not (hes2 or qchem)'"
 
+          - conda-env: mace
+            python-version: 3.10
+            label: MACE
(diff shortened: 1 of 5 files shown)


loriab (approved):
thanks, lgtm. Is there a test that can be added that checks a numerical result? I think the CI failures are unrelated and minor, but I can't fix them until next week.

jthorton:
Hi, @loriab just checking if anything else needs to be done to get this finished up :)

PR #442, opened 2024-02-14
QuChem gaussian added

QuChem:
## Description
This is the implementation of Gaussian software into QCEngine


bennybp:
I think `programs.py` should be `base.py` inside the programs folder. You should modify that file with your changes

qcengine/programs/gaussian.py
```diff
@@ -0,0 +1,269 @@
...
+        cclib_vars = data.getattributes(True)
+        
+        last_occupied_energy = data.moenergies[0][data.homos[0]]
+        #output_data['HOMO ENERGY'] = last_occupied_energy
+        
+        scf_energy = data.scfenergies[0] / constants.conversion_factor("hartree", "eV") # Change from the eV unit to the Hartree unit
+        #output_data['SCF ENERGY'] = scf_energy
+        
+        if input_model.driver == 'energy':
+            output_data['return_result'] = scf_energy
```
loriab, line 217: does `data.scfenergies` from cclib store SCF energies or all method energies? If the former, right now you're allowing _any_ `AtIn.model.method` in the input file but setting the `AtRes.return_result` to the SCF value, not the method value.

qcengine/programs/gaussian.py
```diff
@@ -0,0 +1,269 @@
...
+            gaussian_kw.append("force")
+        elif input_model.driver == "hessian":
+            gaussian_kw.append("freq")
+        else:
+            raise InputError(f"Driver {input_model.driver} not implemented for Gaussian.")
+
+        #if input_model.molecule.fix_com or input_model.molecule.fix_orientation:
+        #    keywords["SYM_IGNORE"] = "TRUE"
+        if 'SCF_CONVERGENCE' in keywords:
+           gaussian_kw.append('SCF=' + keywords["SCF_CONVERGENCE"])
```
loriab, line 132: Where is the name "SCF_CONVERGENCE" coming from? QCNG tends to want the same keyword names in `AtIn.keywords` as the experience user would use in the program natively. Is `Conver=N` on https://gaussian.com/scf/ ("Options" tab) what you're targeting?

qcengine/programs/gaussian.py
```diff
@@ -0,0 +1,269 @@
...
+        "managed_memory": True,
+    }
+    version_cache: Dict[str, str] = {}
+
+    class Config(ProgramHarness.Config):
+        pass
+
+    def found(self, raise_error: bool = False) -> bool:
+        return which(
+            "g09",
```
loriab, line 43: Some things here seem Gaussian09 specific. Probably this should try to accommodate g09 and g16.

qcengine/programs/gaussian.py
```diff
@@ -0,0 +1,269 @@
...
+            return_bool=True,
+            raise_error=raise_error,
+            raise_msg="Please install Gaussian. Check it's in your PATH with `which g09`."
+        )
```
loriab, line 47: you probably also want to check if `cclib` is available when using this harness. see the nwchem harness (that needs networkx) for a model of how to set up looking for both.

loriab (commented):
Thanks for working on this! Here's some initial queries.

PR #443, opened 2024-03-14
Add AIMNET2 harness

jthorton:
## Description
This PR adds an interface to the new [AIMNET2 models](https://github.com/isayevlab/AIMNet2) via a small utility package [pyaimnet2](https://github.com/jthorton/AIMNet2) which allows access to the models via python. 

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/aimnet2.py
+++ b/qcengine/programs/aimnet2.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Dict
 from qcengine.programs.model import ProgramHarness
 from qcelemental.util import safe_version, which_import
 from qcelemental.models import AtomicResult, Provenance
@@ -23,7 +23,7 @@ class AIMNET2Harness(ProgramHarness):
         "managed_memory": False,
     }
 
-    version_cache: dict[str, str] = {}
+    version_cache: Dict[str, str] = {}
 
     @staticmethod
     def found(raise_error: bool = False) -> bool:
@@ -92,8 +92,12 @@ class AIMNET2Harness(ProgramHarness):
             "success": False,
             "properties": {
                 "return_energy": out["energy"].item() * ureg.conversion_factor("eV", "hartree"),
-                "return_gradient": (-1.0 * out["forces"][0].detach().numpy() * ureg.conversion_factor("eV / angstrom", "hartree / bohr")).ravel().tolist(),
-                "calcinfo_natom": len(input_data.molecule.atomic_numbers)
+                "return_gradient": (
+                    -1.0 * out["forces"][0].detach().numpy() * ureg.conversion_factor("eV / angstrom", "hartree / bohr")
+                )
+                .ravel()
+                .tolist(),
+                "calcinfo_natom": len(input_data.molecule.atomic_numbers),
             },
             "extras": input_data.extras.copy(),
         }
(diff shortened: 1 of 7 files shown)


loriab:
Perhaps go ahead and rebase this to fix the conflicts. CI should pass now.

As an AI model, is this training-to-emulate the 3c, etc. methods? I'm checking that I understand why the dftd3/d4/gcp programs _aren't_ dependencies.

qcengine/programs/aimnet2.py
```diff
@@ -0,0 +1,127 @@
...
+
+        ret_data = {
+            "success": False,
+            "properties": {
+                "return_energy": out["energy"].item() * ureg.conversion_factor("eV", "hartree"),
+                "return_gradient": (
+                    -1.0 * out["forces"][0].detach().numpy() * ureg.conversion_factor("eV / angstrom", "hartree / bohr")
+                )
+                .ravel()
+                .tolist(),
```
loriab, line 99: I'm pretty sure you can leave the value as a shaped numpy array. the AtomicResult() creation will handle the rest.

qcengine/programs/aimnet2.py
```diff
@@ -0,0 +1,127 @@
...
+            },
+            "extras": input_data.extras.copy(),
+        }
+        # update with calculated extras
+        ret_data["extras"].update(
+            {
+                "charges": out["charges"].detach()[0].cpu().numpy(),
+                "charges_std": out["charges_std"].detach()[0].cpu().numpy(),
+                "energy_std": out["energy_std"].item(),
+                "forces_std": out["forces_std"].detach()[0].cpu().numpy(),
```
loriab, line 110: these are pretty generic field names. do you want them behind a namespace in extras?

qcengine/programs/aimnet2.py
```diff
@@ -0,0 +1,127 @@
...
+            }
+        )
+        if input_data.driver == "energy":
+            ret_data["return_result"] = ret_data["properties"]["return_energy"]
+        elif input_data.driver == "gradient":
+            # assume the output is eV/bohr
```
loriab, line 116: properties["return_gradient"] above mentions eV/Angstrom. Is this ending up as Hartree/bohr like the above conversion suggests?

qcengine/programs/tests/test_programs.py
```diff
@@ -413,3 +413,50 @@ def test_mace_gradient():
...
+    result = qcng.compute(atomic_input, "aimnet2")
+    assert result.success
+    # make sure the gradient is now the return result
+    assert np.allclose(
+        result.return_result,
+        np.array(
+            [
+                [-0.0, 2.6080331227973375e-09, -0.04097248986363411],
+                [-0.0, -0.029529934749007225, 0.020486244931817055],
+                [-0.0, 0.029529931023716927, 0.020486244931817055],
```
loriab, line 456: How is AIMNET2 handling orientation for arrays? When Molecule fix_com/fix_orientation=T, qcng harnesses are supposed to return the gradient in the same frame as the Molecule.geometry. https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_alignment.py tests this for a few programs, but an AI model probably can't produce HF/Dunning basis. 

fix_*=False (the default) should return the program's standard orientation (whatever that means for the program).

qcengine/programs/aimnet2.py
jthorton: Good spot this is an old comment when the conversion was done in two places rather than using the value already converted in `properties`.

qcengine/programs/tests/test_programs.py
jthorton: The model should always return the result in the same orientation as the input similar to the torchani interface.

qcengine/programs/aimnet2.py
jthorton: Good point I have tried to make them more specific to the aimnet2 model.

jthorton:
@loriab this should be good to go now, let me know if you spot anything else!

loriab (approved):
lgtm, thanks!

CI is fine in a fresh PR, and the failing lane shouldn't be hitting this content, so merging.

jthorton:
Thanks @loriab !

PR #453, opened 2024-09-18
Csse pyd2 506b

loriab:
currently atop #452 

## Status

- [ ] Code base linted
- [ ] Ready to go

PR #454, opened 2024-09-24
Csse pyd2 test use multiple schema versions

loriab:
## Description

~currently atop #453~

## Changelog description

- [x] adapted all test cases running `qcengine.compute` or `qcengine.compute_procedure` to have their commands sandwiched by `checkver_and_convert`, which runs through the 5 `schema_versions` cases: `as_v1` (longstanding), `as_v2`, `to_v1` (user presents a v2 Input and wants a v1 Result back), `to_v2` (reverse of previous), and `None` (default). Tests presently use `schema_versions` to create inputs of the correct version, then the checkver converts them to v1 (since that's what the harnesses use), compute runs and returns them as v1, then checkver converts them to the stated return version, and existing test checks are run. This verifies the qcel conversion functions are working well and that all the tests aren't hardcoded to versions and can run with v2.
- [x] added a less exotic example of an uncommon calc type (driver=properties) since v1 and v2 in qcelemental were handling types differently
- [x] `checkver_and_convert`'s role will be reduced in future when conversion is handled w/i `qcengine.compute`, but it will retain its usefulness as a check of models inheriting from appropriate pydantic and qcschema base classes.
- test suite is 5x longer, but that's temporary. in the end, we'll run `as_v1` and `as_v2` and spot-check the others.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -157,7 +157,7 @@ jobs:
         echo pup
         docker inspect mtzgroup/terachem
         echo cat
-        docker exec mtzgroup/terachem:latest
+        docker exec mtzgroup/terachem:latest which terachem
         echo mouse
         docker exec terachem cat /usr/local/bin/_entrypoint.sh
         echo bun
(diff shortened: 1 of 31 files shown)


loriab:

davidbrownell (approved):

PR #457, opened 2024-10-10
fix auto-enumeration

mikemhenry:
## Description

Markdown style auto-enumeration was used instead of rst

## Changelog description

Fix "Adding a New Program Harness" step numbering

## Status

- [x] Code base linted
- [x] Ready to go


loriab (approved):
Thanks for the fixes!

The last failing CI lane has to do with numpy pinning in the psi4 c-f package. I'll deal with that elsewhere; nothing to do with this PR.

PR #458, opened 2024-11-05
Csse layout 550 error

loriab:
## Description

- [x] see changelog

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/qcengine/programs/model.py
+++ b/qcengine/programs/model.py
@@ -76,19 +76,27 @@ class ProgramHarness(BaseModel, abc.ABC):
         # Note: Someday when the multiple QCSchema versions QCEngine supports are all within the
         #  Pydantic v2 API base class, this can use discriminated unions instead of logic.
 
-        if isinstance(data, qcelemental.models.v1.AtomicInput):
-            mdl = model_wrapper(data, qcelemental.models.v1.AtomicInput)
-        elif isinstance(data, qcelemental.models.v2.AtomicInput):
-            mdl = model_wrapper(data, qcelemental.models.v2.AtomicInput)
+        v1_model = getattr(qcelemental.models.v1, "AtomicInput")
+        v2_model = getattr(qcelemental.models.v2, "AtomicInput")
+
+        if isinstance(data, v1_model):
+            mdl = model_wrapper(data, v1_model)
+        elif isinstance(data, v2_model):
+            mdl = model_wrapper(data, v2_model)
         elif isinstance(data, dict):
             # remember these are user-provided dictionaries, so they'll have the mandatory fields,
             #   like driver, not the helpful discriminator fields like schema_version.
 
-            # for now, the two dictionaries look the same, so cast to the one we want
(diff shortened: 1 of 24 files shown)


loriab:

qcengine/programs/model.py
```diff
@@ -76,19 +76,27 @@ def build_input_model(
         # Note: Someday when the multiple QCSchema versions QCEngine supports are all within the
         #  Pydantic v2 API base class, this can use discriminated unions instead of logic.
 
-        if isinstance(data, qcelemental.models.v1.AtomicInput):
-            mdl = model_wrapper(data, qcelemental.models.v1.AtomicInput)
-        elif isinstance(data, qcelemental.models.v2.AtomicInput):
-            mdl = model_wrapper(data, qcelemental.models.v2.AtomicInput)
+        v1_model = getattr(qcelemental.models.v1, "AtomicInput")
```
davidbrownell, line 79: Is there a reason to access `AtomicInput` this way rather than though an import? If so, it would be helpful to add a comment to explain why that is necessary.

qcengine/programs/psi4.py
```diff
@@ -7,7 +7,8 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar, Dict
 
-from qcelemental.models.v2 import AtomicResult, BasisSet
+from qcelemental.models.v1 import AtomicResult as AtomicResult
```
davidbrownell, line 10: It looks like `as AtomicResult` can be removed.

qcengine/programs/psi4.py
```diff
@@ -7,7 +7,8 @@
...
+from qcelemental.models.v2 import BasisSet
```
davidbrownell, line 11: Is `BasisSet` being used?

davidbrownell (approved):

qcengine/programs/model.py
loriab: Not a particularly good reason -- I switched in here in programs/model.py (where it's always `AtomicInput`) so the analog was clearer with procedures/model.py https://github.com/MolSSI/QCEngine/pull/458/files#diff-2e40dbc50862d2bd8716a45e9dafab44cb3be4499b972dc1d58b8144ef1ab046R63 where the flavor or Input isn't fixed.

qcengine/programs/psi4.py
loriab: Yes, albeit briefly https://github.com/MolSSI/QCEngine/pull/458/files#diff-d66842b418a1c03f41438563803c987c9defd1811ea0af131e285ba540c8c0d6R154

qcengine/programs/psi4.py
loriab: Thanks, that's the second time I've messed that line up. First time was something like `... import AtomicInput as AtomicResult`, which was confusing. I've fixed this and added a comment for your previous comment in the next PR.

PR #495, opened 2026-03-02
SAPT(DFT)-D3/-D4 Empircal Dispersion Parameters

Awallace3:
## Description

I have added optimized parameters for using two-body -d3 and -d4 dispersion with sapt(pbe0) and sapt(b3lyp). There are both intermolecular and supermolecular parameters with preference towards an intermolecular usage when computing interaction energies. Perhaps there should be a not of this somewhere to inform users of this difference. 

## Changelog description

My goal is to include these parameters here in QCEngine ahead of a PR for psi4 that will make use of these parameters for SAPT(DFT)-D4(I) and SAPT(DFT)-D4(S). This would be consistent with how SAPT0-D3 and SAPT0-D4 in psi4 operates, leaving the parameters all in QCEngine.

## Status

- [X] Code base linted
- [X] Ready to go


qcengine/programs/empirical_dispersion_resources.py
```diff
@@ -950,21 +954,48 @@
...
-            #     },
-            #     # "citation": '    Wallace, A. M.; Sherrill, C. D.',
-            # },  # AMW 7/25, SAPT(PBE0)(S)
-            # "sapt(pbe0)(i)": {"params": { "s6": 1.0, "s8": 0.89529649, "a1": -0.82043591, "a2": 0.03264695, "s9": 0.00000000e00, "alp": 16.0, "ga": 3.0, "gc": 2.0, "wf": 6.0, }, "citation": '    Wallace, A. M.; Sherrill, C. D.', }, # AMW 7/25
+            "sapt(pbe0)(s)": {
+                "params": {
+                    "s6": 1.0,
+                    "s8": 1.2042,
+                    "a1": 0.9090,
+                    "a2": 0.0,  # change to 0.0?
```
loriab:
```suggestion
                    "a2": 0.0,
```
much better to have fewer decimal places

qcengine/programs/empirical_dispersion_resources.py
```diff
@@ -950,21 +954,48 @@
...
+                    "a1": 0.7773,
+                    "a2": 0.0,
+                    "s9": 0.00000000e00,
+                    "alp": 16.0,
+                    "ga": 3.0,
+                    "gc": 2.0,
+                    "wf": 6.0,
+                },
+                "citation": "    Wallace, A. M.; Sherrill, C. D. (TBD)",
+            },  # AMW 2/26, LoS-II Total aDZ MAE 0.29 kcal/mol
```
loriab, line 1031: no `sapt(b3lyp)(s)` for d4?

loriab (approved):

qcengine/programs/empirical_dispersion_resources.py
Awallace3: > no `sapt(b3lyp)(s)` for d4?

I have added these parameters for `sapt(b3lyp)-d4(s)` to make it complete. Also, I have added the `sapt0-d4m(i)` parameters as well. The psi4 sapt0-d4m default will be updated in [Psi4 PR#3372](https://github.com/psi4/psi4/pull/3372) to use these parameters along with the intermolecular sum instead of the supermolecular approach due to unphysical behavior at long-range.

Awallace3:
I think I didn't pull your changes before checking out my saptdft-d4 branch, so that is where that merge came from. I think everything is okay from seeing your commit in the history but thought I would mention it in case I am missing something.

PR #501, opened 2026-04-10
Implementation of a Gaussian Harness

taylor-a-barnes:
## Description

This PR contains an initial effort to implement support for Gaussian in QCEngine.

The code was largely generated with Claude Code, using requirements files in the `rqm` directory that this PR introduces.  The various requirements hashes (e.g., `rq-ef62979b`, `rq-c266b9bb`, `rq-f1c116db`, etc.) are identifiers for individual elements of the requirements specification, and form an explicit map between the code and the requirements.  These hashes and the requirements files can be removed from the PR, if desired.  If nothing else, perhaps this serves as an opportunity to assess the quality of AI-generated harnesses.

In addition to the tests in `test_gaussian.py`, you can reference the following as an example.  It should work with either `g16` or `g09`, but I've only tested with `g16` thus far.

```python
from qcelemental.models.v2 import AtomicInput, AtomicSpecification, Molecule
import qcengine as qcng

mol = Molecule(
    symbols=["O", "H", "H"],
    geometry=[
        0.000000,  0.000000,  0.226296,
        0.000000,  1.430901, -0.905183,
        0.000000, -1.430901, -0.905183,
    ],
)

inp = AtomicInput(
    molecule=mol,
    specification=AtomicSpecification(
        program="gaussian",
        driver="energy",
        model={"method": "HF", "basis": "STO-3G"},
    ),
)

result = qcng.compute(inp, "gaussian")

print(f"Success:       {result.success}")
print(f"Total energy:  {result.return_result:.10f} Eh")
print(f"Version:       {result.provenance.version}")
```

## Changelog description

Implementation of a Gaussian harness.

## Status

- [ ] Code base linted
- [ ] Ready to go

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,7 @@ markers = [
     "dftd4: tests using DFTD4 software; skip if unavailable",
     "dftd4_350: tests using DFTD4 >=3.5.0 software; skip if unavailable",
     "gamess: tests using GAMESS-US software; skip if unavailable",
+    "gaussian: tests using Gaussian software; skip if unavailable",
     "mctc-gcp: tests using modern mctc-gcp software; skip if unavailable",
     "classic-gcp: tests using GCP executable software; skip if unavailable",
     "geometric: tests using geomeTRIC software; skip if unavailable",
--- a/qcengine/programs/gaussian/harvester.py
+++ b/qcengine/programs/gaussian/harvester.py
@@ -1,0 +1,172 @@
+"""Parse Gaussian log output using cclib with regex fallback."""
+
+import io
+import re
+from typing import Optional, Tuple
+
+import numpy as np
+import qcelemental as qcel
+from qcelemental.models import Molecule
+
+from ..util import PreservingDict
+
+# rq-bb0a5d96 rq-c20746b1
+_NORMAL_TERMINATION_RE = re.compile(r"Normal termination of Gaussian")
+_VERSION_RE = re.compile(r"Gaussian\s+(\d+),\s+Revision\s+([\w.]+),")
+
+# eV → Hartree conversion factor
+# rq-bb0a5d96
+_EV_TO_HARTREE = qcel.constants.conversion_factor("eV", "hartree")
+
+
+# rq-d9434480
+def harvest(
+    in_mol: Molecule,
+    method: str,
+    log_content: str,
+) -> Tuple[PreservingDict, Optional[np.ndarray], Optional[np.ndarray], Molecule]:
+    """Parse a Gaussian log file and return QC data for building an AtomicResult.
+
+    Parameters
+    ----------
+    in_mol:
+        The input molecule used for geometry alignment and NRE cross-check.
+    method:
+        Lowercase QCSchema method string (e.g. "hf", "mp2", "b3lyp").
+    log_content:
+        Full text of gaussian.log.
+
+    Returns
+    -------
+    (qcvars, grad, hess, out_mol)
+        qcvars:  PreservingDict of QC variables (energies, charges, etc.).
+        grad:    Gradient array shape (3*natoms,) in Hartree/Bohr, or None.
+        hess:    Hessian matrix shape (3*natoms, 3*natoms) in Hartree/Bohr², or None.
+        out_mol: Output molecule aligned to in_mol.
+
+    Raises
+    ------
+    ValueError
+        If no coordinate information can be extracted from the log.
+    ValueError
+        If the parsed geometry's NRE deviates from in_mol's by more than 1e-3 Hartree.
+    """
+    # rq-bb0a5d96 — parse with cclib; exceptions propagate to caller (rq-bb775969)
+    import cclib
+
+    ccdata = cclib.io.ccopen(io.StringIO(log_content)).parse()
+
+    qcvars = PreservingDict()
+    method_lower = method.lower()
+
+    # --- Geometry and output molecule ---
+    # rq-8d6fc024 rq-a76c3d7b rq-b7c7b8cb
+    if not hasattr(ccdata, "atomcoords") or len(ccdata.atomcoords) == 0:
+        raise ValueError("No coordinate information could be extracted from the Gaussian log.")
+
+    # atomcoords in Ångström → convert to Bohr for QCElemental
+    angstrom_to_bohr = qcel.constants.conversion_factor("angstrom", "bohr")
+    coords_bohr = ccdata.atomcoords[-1] * angstrom_to_bohr  # shape (natoms, 3)
+
+    symbols = [qcel.periodictable.to_symbol(z) for z in ccdata.atomnos]
+
+    calc_mol = Molecule(
+        symbols=symbols,
+        geometry=coords_bohr.flatten(),
+        fix_com=True,
+        fix_orientation=True,
+    )
+
+    natoms = len(ccdata.atomnos)
+    qcvars["N ATOMS"] = str(natoms)
+
+    # rq-0753cd03
+    qcvars["NUCLEAR REPULSION ENERGY"] = str(round(calc_mol.nuclear_repulsion_energy(), 10))
+
+    # rq-90604835 — cross-check NRE against input molecule
+    if abs(calc_mol.nuclear_repulsion_energy() - in_mol.nuclear_repulsion_energy()) > 1.0e-3:
+        raise ValueError(
+            f"Gaussian output geometry (NRE: {calc_mol.nuclear_repulsion_energy():.6f}) "
+            f"is inconsistent with the input molecule (NRE: {in_mol.nuclear_repulsion_energy():.6f})."
+        )
+
+    # Align output molecule to input molecule (same pattern as GAMESS harvester)
+    # rq-8d6fc024
+    return_mol, _ = in_mol.align(calc_mol, atoms_map=False, mols_align=True, run_mirror=True, verbose=0)
+
+    # --- Energies ---
+    # rq-37f40e92 — HF/SCF energy is always extracted from scfenergies
+    scf_hartree = float(ccdata.scfenergies[-1]) * _EV_TO_HARTREE
+    qcvars["HF TOTAL ENERGY"] = str(scf_hartree)
+
+    # rq-929a262a rq-c5165f1f — method-specific variables
+    if method_lower in ("hf", "scf") or _is_dft(method_lower):
+        qcvars["CURRENT ENERGY"] = str(scf_hartree)
+
+    elif method_lower == "mp2":
+        # rq-578bb785
+        mp2_hartree = float(ccdata.mpenergies[-1][-1]) * _EV_TO_HARTREE
+        qcvars["MP2 TOTAL ENERGY"] = str(mp2_hartree)
+        qcvars["MP2 CORRELATION ENERGY"] = str(mp2_hartree - scf_hartree)
+        qcvars["CURRENT ENERGY"] = str(mp2_hartree)
+
+    elif method_lower == "ccsd":
+        # rq-1b7ae62e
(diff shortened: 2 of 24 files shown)


rqm/gaussian/harvester.md
```diff
@@ -0,0 +1,281 @@
...
+- Any `cclib` exception is allowed to propagate (caller wraps in `UnknownError`).
+
+### QC Variables Populated <!-- rq-929a262a -->
+
+The following variables are placed in `qcvars` based on the method:
+
+**All calculations**:
+- `"NUCLEAR REPULSION ENERGY"` — from parsed geometry. <!-- rq-0753cd03 -->
+- `"HF TOTAL ENERGY"` — always set from `ccdata.scfenergies[-1]` (in Hartree). <!-- rq-37f40e92 -->
+- `"CURRENT ENERGY"` — set to the highest-level energy available (see below). <!-- rq-c5165f1f -->
```
loriab, line 214: I can see where you'd get the pattern to populate both extras["qcvars"] and properties from the other harnesses. We'll want to focus on properties since that's proper QCSchema. But then that's a smaller list ... . Mostly FYI for now that properties is the focus.

loriab (commented):
Thanks very much! And I'm sorry for the much-delayed review. I think the main item is more end-to-end tests and integration with current program harness tests. In particular the below would be nice. Glad to discuss!

* https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_ghost.py#L42
* https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_ghost.py#L169
* https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_ghost.py#L264
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_alignment.py#L95
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_canonical_fields.py#L16
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_canonical_config.py#L67
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_canonical_config.py#L16
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_canonical_config.py#L129
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_canonical_config.py#L215
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_canonical_config.py#L215 (and other refs)
* https://github.com/MolSSI/QCEngine/blob/cdda3660afee241968907c3020ee368ef8230b99/qcengine/programs/tests/test_standard_suite_ccsd(t).py#L45 (and other refs)

qcengine/programs/gaussian/harvester.py
```diff
@@ -0,0 +1,300 @@
...
+    # precision loss from cclib's eV round-trip (cclib and qcelemental use
+    # slightly different Hartree↔eV conversion factors, causing ~1e-6 error).
+    # Fall back to cclib values if regex fails.
+
+    # rq-37f40e92 — HF/SCF energy
+    scf_match = _SCF_DONE_RE.search(log_content)
+    if scf_match:
+        scf_hartree = float(scf_match.group(1))
+    else:
+        scf_hartree = float(ccdata.scfenergies[-1]) * _EV_TO_HARTREE
```
berquist: Note that this is without an (empirical) dispersion correction, that's in `dispersionenergies`, which I think is SCF + disp.  If you do something that has NLC, self-consistent or otherwise, I think it will be in `scfenergies`.

qcengine/programs/gaussian/harvester.py
```diff
@@ -0,0 +1,300 @@
...
+        grad = mill.align_gradient(calc_grad).flatten()
+
+    # --- Hessian ---
+    # rq-f2e32162 rq-344f3432
+    hess: Optional[np.ndarray] = None
+    if hasattr(ccdata, "hessian") and ccdata.hessian is not None:
+        # cclib hessian would be in standard orientation → transform with mill
+        hess = mill.align_hessian(np.array(ccdata.hessian))
+    else:
+        # cclib does not parse force constants from Gaussian Freq output;
```
berquist: According to the table in https://cclib.github.io/data.html it should have `vibfconsts`, or at least one version of it.

rqm/gaussian/harvester.md
```diff
@@ -0,0 +1,393 @@
...
+The `ccdata` object exposes parsed attributes used as follows:
+
+| cclib attribute | Units (cclib) | Conversion | QCEngine usage |
+|-----------------|---------------|------------|----------------|
+| `ccdata.scfenergies[-1]` | eV | (fallback only) | SCF/HF/DFT total energy |
+| `ccdata.mpenergies[-1][-1]` | eV | (fallback only) | MP2 total energy |
+| `ccdata.ccenergies[-1]` | eV | (fallback only) | CCSD or CCSD(T) total energy |
+| `ccdata.grads[-1]` | Hartree/Bohr | negate (forces → gradient) | Gradient array |
+| `ccdata.atomcoords[-1]` | Ångström | × Bohr/Å | Geometry for output molecule (standard orientation) |
+| `ccdata.atomnos` | atomic numbers | via qcel periodic table | Element symbols |
```
berquist: I forget where it is in the harvester code, but I saw `natom` or `natoms` derived from the length of `atomnos`, but there is a simple `natom` attribute that's an integer.

rqm/gaussian/harvester.md
```diff
@@ -0,0 +1,393 @@
...
+"Direct Energy Parsing" below). cclib energy attributes are used only as a fallback
+if the regex fails to match.
+
+> **Considered alternative:** Using `cclib.parser.utils.convertor(ccdata.scfenergies[-1],
+> 'eV', 'hartree')` produces a lossless round-trip (cclib's own forward and reverse
+> factors are exact inverses). This was rejected in favour of direct parsing because:
+> (a) it couples the harvester to cclib's internal conversion remaining self-consistent
+> across future versions, (b) direct log parsing captures the full precision printed by
+> Gaussian (up to 12 significant digits), and (c) cclib provides no public API contract
+> guaranteeing lossless eV↔Hartree round-trips.
```
berquist: a. It has always been self-consistent.
c. Related to (a), it is true that there is no public API contract, but in practice it should have always been lossless within floating point guarantees.
b. This is a bigger problem: for new code we try to parse out the value that has the greatest precision, but for historical code (such as the Gaussian parser) a lot of these decisions have been lost to time and we can't make guarantees about it without going and reevaluating the entire parser.  A good start for cclib devs would be to see if we are reading the same fields that the QCEngine regex solution is reading.

berquist: Ignoring the possible solution of doing the round trip using `cclib.parser.utils.convertor` back to hartree, all of this eV madness is going away in cclib v2.

rqm/gaussian/harvester.md
```diff
@@ -0,0 +1,393 @@
...
+
+**2. Hessian not extracted by cclib:**
+
+cclib does not parse the "Force constants in Cartesian coordinates" block from
+Gaussian frequency (`Freq`) calculations. The `ccdata.hessian` attribute is always
+absent for Gaussian logs.
```
berquist: I am surprised to find that this is true.

pyproject.toml
berquist, line 48: If you add `cclib` to one of the dependency lists here, make sure to pin it with `<2`.

qcengine/programs/gaussian/harvester.py
taylor-a-barnes: Thanks for pointing this out.  Commit e0eb8c9 adds support for dispersion correction into the Gaussian harness, and I believe handles this correctly.  There are quite a few edge cases in the parsing, so it's possible I've missed something.

qcengine/programs/gaussian/harvester.py
taylor-a-barnes: You're correct; the comment is inaccurate.  The real issue is that, as far as I can tell, there isn't any way to get the full 3N x 3N hessian matrix from cclib parsing Gaussian.  1b8c160 rewords the comment.

rqm/gaussian/harvester.md
taylor-a-barnes: Yes, I've tried looking through the cclib code (gaussianparser.py, etc.), and I don't see any place that extracts the hessian.  Feel free to correct me if I'm missing something.

rqm/gaussian/harvester.md
taylor-a-barnes: Thanks; 190ff9a switches to using `natom`.

rqm/gaussian/harvester.md
taylor-a-barnes: Upon review, I can confirm that cclib does not lose precision relative to the printed Gaussian output.  The misstatement was a hold-over from my confusion while attempting to track down the cause of the small errors from the cclib / qcelemental conversion disagreement.  I've converted to using cclib for the energies in d884648.

rqm/gaussian/harvester.md
taylor-a-barnes: Thanks for the review!  I believe the harness is now incorporated into all the tests you listed, and added several end-to-end tests.  Please let me know if there is anything else I should modify.

qcengine/programs/gaussian/harvester.py
```diff
@@ -0,0 +1,360 @@
...
+
+    # --- Empirical dispersion correction ---
+    # rq-1d5da737 — Gaussian writes the dispersion contribution on a separate
+    # "Dispersion energy=" line; cclib keeps it out of `scfenergies` and
+    # stores it (in eV) in `dispersionenergies`. Add it to the SCF total when
+    # present.
+    disp_hartree: Optional[float] = None
+    if (
+        hasattr(ccdata, "dispersionenergies")
+        and ccdata.dispersionenergies is not None
```
berquist, line 279: This probably is generated because the attribute may not exist.  I can definitively say that for version < 2, if an attribute exists it will never be `Optional`.  `hasattr` is always required, though.

qcengine/programs/gaussian/harvester.py
```diff
@@ -0,0 +1,360 @@
...
+from typing import Optional, Tuple
+
+import numpy as np
+import qcelemental as qcel
+from qcelemental.models import Molecule
+
+from ..util import PreservingDict
+
+# rq-0743e952
+_NORMAL_TERMINATION_RE = re.compile(r"Normal termination of Gaussian")
```
berquist, line 14: You can use `ccdata.metadata["success"]`, which is `bool` that may not be present, so you'll need to use `.get`.  It is not defaulted to `None` or a boolean value.

qcengine/programs/gaussian/harvester.py
```diff
@@ -0,0 +1,360 @@
...
+_VERSION_RE = re.compile(r"Gaussian\s+(\d+),\s+Revision\s+([\w.]+),")
```
berquist, line 15: If you're brave, you can use `ccdata.metadata["parsed_version"]` which is probably always present but technically needs `get`, and it will return a [`Version`](https://packaging.pypa.io/en/latest/version.html).

berquist: Sorry, it's `"package_version"` and is a `str` that is guaranteed parsable into a `Version`.

PR #511, opened 2026-07-06
Add put xml line to molpro input builder.

cgbriggs99:
## Description

After a certain version, MolPro reduced the amount of output in their XML dump. The way to get the old style back is to add a line containing `{put,xml}` at the point where you want it to put its information. It is currently set to put it at the end. This will hopefully capture the last energy calculation and optional gradients.

## Changelog description

Updated MolPro XML output to take the new default dump level into consideration.

## Status

- [x] Code base linted
- [x] Ready to go


cgbriggs99:
Ready for review.

PR #513, opened 2026-07-29
Add a Jaguar harness

TyBalduf:
## Description
Add a Jaguar harness that can implements the standard `energy`, `gradient`, and `hessian` drivers. The harness is registered in both the standard suite and the alignment tests. The harness relies on the Schrodinger Suite tools, namely `JaguarInput` (for generating an input file from a qcschema) and `JaguarOutput` (for parsing properties from the output, to generate an  `AtomicResult`).

Incidentally, I also sped up the testing utility `is_program_new_enough`, which previously searched the PATH for all registered programs, even though it just needs to check if the specific program is installed/compatible.

## Changelog description
Add a harness for running Schrodinger's Jaguar QM program

## Status

- [x] Code base linted
- [x] Ready to go
- [x] Jaguar related tests pass for me locally with my Schrodinger suite installation


qcengine/programs/tests/test_standard_suite.py
```diff
@@ -199,16 +205,19 @@ def test_hf_energy_module(inp, dertype, basis, subjects, clsd_open_pmols, reques
...
         pytest.param({"call": "nwchem", "reference": "uhf",  "fcae": "ae", "keywords": {"scf__uhf": True, "scf__thresh": 1.e-6},                                             }, id="hf  uhf ae: nwchem",     marks=using("nwchem")),
         pytest.param({"call": "psi4",   "reference": "uhf",  "fcae": "ae", "keywords": {"reference": "uhf", "scf_type": "pk"},                                               }, id="hf  uhf ae: psi4",       marks=using("psi4")),
 
         pytest.param({"call": "cfour",  "reference": "rohf", "fcae": "ae", "keywords": {"reference": "rohf", "scf_conv": 12},                                                }, id="hf rohf ae: cfour",      marks=using("cfour")),
         pytest.param({"call": "gamess", "reference": "rohf", "fcae": "ae", "keywords": {"contrl__scftyp": "rohf"},                                                           }, id="hf rohf ae: gamess",     marks=using("gamess")),
+        pytest.param({"call": "jaguar", "reference": "rohf", "fcae": "ae", "keywords": {"dconv": 1.e-8, "econv": 1.e-9, "iuhf": 0, "nops": 1},                                                               }, id="hf rohf ae: jaguar",     marks=using("jaguar")),
```
loriab, line 220: Are this (and the third change up) for rohf getting the same keywords as uhf?

loriab:
Dont' worry about those CI errors. Pyberny did a new release that newly rejects unfamiliar keywords. I've fixed it in #512

qcengine/programs/jaguar.py
```diff
@@ -0,0 +1,544 @@
...
+        the module that supplies the product version.
+
+        Returns
+        -------
+        str
+            A PEP 440-compatible Jaguar version such as ``13.4.39``.
+        """
+        self.found(raise_error=True)
+
+        module_path = which_import("schrodinger.infra.mm")
```
loriab, line 83: If this `which_import` gives you trouble (or poor error messages), try https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/openmm.py#L131-L133 for nested import.

qcengine/programs/tests/test_standard_suite.py
```diff
@@ -81,6 +81,7 @@ def _trans_key(qc, bas, key):
             return {
                 "cfour": ("pvdz", {}),
                 "gamess": ("ccd", {"contrl__ispher": 1}),
+                "jaguar": (bas, {}),
```
loriab, line 84: This is looking really nice, thank you! I'm still reading through, but I thought I'd post a couple comments and some possible next test files to integrate with. 
- https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_standard_suite_hf.py
- https://github.com/MolSSI/QCEngine/blob/master/qcengine/tests/test_harness_canonical.py
- https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_canonical_fields.py
- https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/test_canonical_config.py (some might not be applicable)

qcengine/programs/tests/test_standard_suite.py
TyBalduf: Our `iuhf` keyword has the interpretation:
* `iuhf=0`: Do a restricted spin calculation 
* `iuhf=1`: Do an unrestricted spin calculation
* `iuhf=2`: (the default) do a restricted spin calculation for singlets, unrestricted for any other multiplicity.

So the `iuhf=0` here just enforces running ROHF, rather than it defaulting to UHF (assuming multiplicity>1). If the test examples are just singlets, it will just do RHF.

loriab (approved):
very nice, thank you!

I'm going to update the branch, as I think those CI errors should be fixed by the recent release.

PR #520, opened 2026-08-17
add Orb universal machine learned force field harness

jacob-votava:
## Description

I've added a harness for [orb-models](https://github.com/orbital-materials/orb-models). It is based on the previously implemented AIMNET2 harness. It supports `energy` and `gradient` drivers, same as AIMNET2. Three main things are different from the AIMNET2 implementation.

- **Force output keys differ between orb's two model types.** AIMNET2 only has one mode
  (forces by autograd on the energy). For Orb direct models return a
  `forces` key while conservative models return `grad_forces` since Orb has both direct and conservative models. The harness reads the key name off the model rather than hardcoding it, so both families of models go through the same code path.
- **The omol models need both charge and multiplicity for the input** AIMNET2 passes charge but never needs
  multiplicity whereas Orb's omol models need both charge and spin multiplicity. Orb requires both on `atoms.info` and will throw an error if either is missing.
- **The device is selectable for orb models** AIMNET2 hardcodes cpu but in Orb a `device` keyword chooses cpu/cuda and defaults to cuda when available. The tests pin `"cpu"` so reference values don't depend on the runner.

### Verification

- Tests are the same as AIMNET2 with hardcoded water energies/gradients so a dropped unit conversion or flipped sign fails CI, which was checked by introducing both bugs.
- As a final check I ran geometry optimizations on some caffeine conformers. I generated 4 ETKDG
  starting conformers and ran a geometry optimization with 3 models (orb direct, orb conservative and AIMNET2) and every run converged.

Convergence plot, inset is one of the rendered minima, done using OVITO the marker is the last step above each run's converged energy which is off scale.

*Convergence plot, inset is a render of one of the minima done using OVITO, the marker is the last step above each run's converged energy which is off scale.

## Changelog description

ORB - Added harness for orb-models universal ML interatomic potentials.

## Status

- [x] Code base linted
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -159,6 +159,13 @@ jobs:
             runs-on: ubuntu-latest
             pytest: ""
 
+          # 3.12 only: orb pins dm-tree==0.1.8, which has no wheel beyond cp312
+          - conda-env: orb
+            python-version: "3.12"
+            label: ORB
+            runs-on: ubuntu-latest
+            pytest: ""
+
           #- conda-env: opt-disp-cf
           #  python-version: 3.12
           #  label: QCSk-next
--- a/devtools/conda-envs/orb.yaml
+++ b/devtools/conda-envs/orb.yaml
@@ -1,0 +1,27 @@
+name: test
+channels:
+  - conda-forge
+dependencies:
+    # keep the scientific stack on the pip side: conda's llvm-openmp and the libomp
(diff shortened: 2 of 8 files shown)


jacob-votava:
If there's any interest I wanted to use UMA for related work as well, so if this is reviewed/merged I'd like to contribute a FAIRChem harness in the future.

